PR #17, opened 2020-03-31
Small fixes and refactor mainly on the Medical.js

linnnruoo:
- The bug was caused by the incorrect use of axios, so I did a quick fix and I'll probably extract the api layer away from the component layer soon
- My vscode auto-formats code upon saving and is prettier-enabled so most of the changes come from the IDE reformatting the code 🤪 I can add the prettier config in the codebase!
- added `useUnifiedTopology` in the mongoose connection config because it keeps complaining, I referred from https://github.com/Automattic/mongoose/issues/8156
- I also referred to https://github.com/manualzuru/FightPandemics.com/pull/15 for the fix on getting the geolocation

### DIFF
--- a/client/src/pages/Medical.js
+++ b/client/src/pages/Medical.js
@@ -1,52 +1,62 @@
-import axios from 'axios';
-import React, { useState, useEffect } from 'react';
-import { Nav } from 'react-bootstrap';
+import axios from "axios";
+import React, { useState, useEffect } from "react";
+import { Nav } from "react-bootstrap";
 
-import { getLocalStorageJson } from '../utils/local-storage';
+import { getLocalStorageJson } from "../utils/local-storage";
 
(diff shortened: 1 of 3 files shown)


client/src/pages/Medical.js
```diff
@@ -1,52 +1,62 @@
...
-    );
+  const [state, setState] = useState(INITIAL_STATE);
+  const fetchData = () => {
+    const geolocation = getGeoLocation();
+    axios
+      .post("/api/geo/country", geolocation)
+      .then(res => setState({ ...state, country: res.data }))
+      .catch(err => console.log(err));
+  };
+  useEffect(fetchData, []);
```
robinv85: I think this dependency of the hook was correct, because if `geolocation` changes the hook needs to execute again to refetch the country

PR #28, opened 2020-04-01
Added a few base components and refactored certain components 

linnnruoo:
- refactored bits and pieces
- extracted logo, checkboxitem and h2 into individual components
- changed the primary color to #425af2 according to the palette given in figma 
- added poppins as this was given in figma too
- added prettier yaml file according to the current implementation as my default config overrode it
- added dependencies `styled-components` and `typeface-poppins`


linnnruoo:
Linked to those two issues that are relevant

PR #29, opened 2020-04-01
Symptops checker

Gtstride:


robinv85:
this has been rebased to master with fixed conflicts already, but the symptoms checker should probably also use the react-wizard component on one single "page"

PR #34, opened 2020-04-01
Docker-compose file

adace123:
Would appreciate testing of this.
Not sure how to serve the client build directory since that's outside the build context of the backend. Would we want to just change the build context to the root directory?


robinv85:
@adace123 cool ok I'll take a look when my current task is finished

adace123:
Trying to figure out why the axios request to the geo server fails with ECONNREFUSED in Docker but the same request with curl in my terminal doesn't.

docker-compose.yml
```diff
@@ -0,0 +1,37 @@
+version: "3"
+services:
+  mongo:
+    image: mongo
+    container_name: mongodb
+    ports:
+      - "27017:27017"
+    command: mongod
+    volumes:
+      - "~/data:/data/db"
```
neemiasjnr, line 10: I have some data folder for other projects, it's better to put it inside the project (also adding it to .gitignore)
```suggestion
      - "./data:/data/db"
```

backend/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM node:slim
```
neemiasjnr, line 1: I would suggest to use node 12 alpine, it's 1/5 the size of the slim image
```suggestion
FROM node:12-alpine
```

docker-compose.yml
robinv85: Yes, good suggestion, I am currently testing it locally

backend/Dockerfile
robinv85: Yep probably better since I don't need we need any of the additional packages

backend/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+
+WORKDIR /app
+
+COPY package.json .
+
+COPY package-lock.json .
```
neemiasjnr, line 7:
```suggestion
COPY package.json package-lock.json ./
```

robinv85, line 7: the project was bootstrapped with `yarn` instead of npm and currently has `yarn.lock` files so probably need to replace it with npm then and use `package-lock` instead

backend/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+
+RUN npm install && npm install -g nodemon
+
+ARG NODE_ENV
+ENV NODE_ENV=${NODE_ENV}
```
neemiasjnr, line 12: The ENV line is not necessary

backend/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM node:slim
+
+WORKDIR /app
+
+COPY package.json .
+
+COPY package-lock.json .
+
+RUN npm install && npm install -g nodemon
```
neemiasjnr, line 9: Running with nodemon is not recommended for production because it consumes a LOT of memory, and the orchestration tool (or even Heroku) should do the job to restart the process in case it fail

backend/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+
+ARG NODE_ENV
+ENV NODE_ENV=${NODE_ENV}
+
+COPY . .
+
+CMD ["npm", "run", "server"]
```
neemiasjnr, line 16:
```suggestion
CMD ["npm", "start"]
```

backend/routes/geo.js
```diff
@@ -5,10 +5,10 @@ const countries = require("../data/countries");
 const { validateGeolocation } = require("../validation/geolocation");
 
 const router = express.Router();
-const GEO_SERVICE_URL = "http://127.0.0.1:5000/geo-service";
+const GEO_SERVICE_URL = "http://0.0.0.0:5000/geo-service";
```
neemiasjnr, line 8: Can we move GEO_SERVICE_URL to an environment variable?

backend/server.js
```diff
@@ -35,13 +37,15 @@ app.use("/api/geo", geo);
 //  Serve static asset assets if in production
 if (process.env.NODE_ENV === "production") {
   // Set Static Folder
-  app.use(express.static("client/build"));
+  app.use(express.static("../client/build"));
 
   app.get("*", (req, res) => {
-    res.sendFile(path.resolve(__dirname, "client", "build", "index.html"));
+    res.sendFile(path.resolve(__dirname, "../client", "build", "index.html"));
```
neemiasjnr, line 43: I would suggest to separate the client and backend containers. The client container can run the build files with nginx in front of it. I can help on this if you want

client/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM node:slim
```
neemiasjnr, line 1:
```suggestion
FROM node:12-alpine
```

backend/Dockerfile
neemiasjnr: Another comment, can we remove yarn.lock and start using npm everywhere with package-lock.json? I don't see any reason to use yarn anymore

backend/Dockerfile
robinv85: nodemon is good for development, could also use `forever` or `pm2` ... but I think we will probably move to AWS EKS or EC2 instead of Heroku

backend/routes/geo.js
robinv85: yes, definitely good suggestion

backend/server.js
robinv85: yes this was also the intention of using docker compose, better with Nginx or some other static server (Caddy maybe?)

robinv85:
@adace123 I tnink the connection error you mentioned is because there is no more proxy configured so there is a CORS error `Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at http://0.0.0.0:5050/api/geo/country. (Reason: CORS request did not succeed)` 
This was set with create-react-app proxy so with Docker compose we need to update that to work with the containers.

PR #38, opened 2020-04-02
Refactoring + docker improvements

neemiasjnr:
This PR will decouple the front end and the backend container. To test both now:
locahost:3000 -> client
localhost:8000/version -> backend


backend/.env.example
```diff
@@ -0,0 +1,4 @@
+PORT=8000
```
robinv85, line 1: ok if we change this might need to adjust the auth url's for dev also, but is currently going through client directly

backend/config.js
```diff
@@ -0,0 +1,37 @@
...
+};
+
+const config = {
+  name,
+  server: {
+    port: process.env.PORT || 8000,
+  },
+  mongo: {
+    get host() {
+      return process.env.MONGO_URI;
```
robinv85: why use a getter here, in some places and not in others... has to be dynamic?

backend/lib/endpoints/users.js
```diff
@@ -43,18 +43,13 @@ router.post("/auth", (req, res) => {
...
-      { expiresIn: "7d" },
-      (err, token) => {
-        console.log(err);
-        res.json({
-          success: true,
-          token: `Bearer ${token}`,
-        });
-      }
-    );
+    jwt.sign(payload, config.jwt.key, { expiresIn: "7d" }, (err, token) => {
```
robinv85, line 46: can also put JWT_EXPIRATION_TIME in env

backend/.env.example
neemiasjnr: It's working now, can you please try the links I put on the PR description?

robinv85:
ok fixed, good to go !

PR #40, opened 2020-04-02
Added AntD mobile UI

shivanshs9:
- Homepage for mobile done
- Added webpack resolve alias
- Added theme scss
- Added ImageButton and FeedbackButton


client/package.json
```diff
@@ -36,9 +37,9 @@
     "typeface-roboto": "^0.0.75"
   },
   "scripts": {
-    "start": "react-scripts start",
-    "build": "react-scripts build",
-    "test": "react-scripts test",
+    "start": "react-app-rewired start",
```
robinv85, line 40: Why did you need these script overrides?

client/package.json
```diff
@@ -57,7 +58,8 @@
     ]
   },
   "devDependencies": {
-    "node-sass": "^4.13.1"
+    "node-sass": "^4.13.1",
```
robinv85, line 61: So we should remove node-sass and scss and Grommet and Bootstrap and only keep Ant and Styled Components, but we can do that in the next iteration and merge this first

client/src/index.scss
```diff
@@ -1,14 +1,10 @@
...
-@import "~bootstrap/scss/bootstrap.scss";
+// Import Theme Style
+@import "./styles/theme.scss";
 
 body {
   margin: 0;
-  font-family: "Poppins", "Roboto", "Oxygen", "Ubuntu", "Cantarell", "Fira Sans",
-    "Droid Sans", "Helvetica Neue", sans-serif;
+  background: $body-bg;
+  font-family: $font-family;
```
robinv85, line 7: yes can probably remove these things and move them to Styled Components theme

client/package.json
shivanshs9: To extend webpack config.. I didn't want to replace the config provided by create-react-app.
And I needed to configure webpack for aliases

client/src/index.scss
shivanshs9: You mean using ThemeProvider?

robinv85: yes I guess replace what we had before to use https://styled-components.com/docs/advanced#theming

PR #44, opened 2020-04-03
Fix auth

anniegiang:
Backend auth works fine now.

- Added body-parser to fix the issue where the `req.body` in `POST /users/auth` was empty and not being read.
- Renamed `name` to `firstName` for better semantics. On that note, should we require a user to input their `lastName` on registration?

### DIFF
--- a/backend/lib/index.js
+++ b/backend/lib/index.js
@@ -3,6 +3,7 @@ const cors = require("cors");
 const passport = require("passport");
 const pino = require("pino")();
 const pinoExpress = require("pino-express");
+var bodyParser = require("body-parser");
 
 const passportMiddleware = require("./middlewares/passport");
 const errorMiddleware = require("./middlewares/error");
@@ -17,6 +18,8 @@ function createApp() {
   app.use(cors());
   app.use(pinoExpress(pino));
   app.use(express.json());
+  app.use(bodyParser.json());
+  app.use(bodyParser.urlencoded());
 
   // Passportn Middleware
   app.use(passport.initialize());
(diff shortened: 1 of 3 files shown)


backend/lib/index.js
```diff
@@ -17,6 +18,8 @@ function createApp() {
   app.use(cors());
   app.use(pinoExpress(pino));
   app.use(express.json());
+  app.use(bodyParser.json());
+  app.use(bodyParser.urlencoded());
```
neemiasjnr, line 22: Can you please try to remove the body parser and see if still works?

anniegiang: Without body-parser `req.body` is `{}` when it should be something like `{ email: test@gmail.com, firstName: Test }` in the use case for `/api/users/auth`. Do you have any suggestions to get this to work without body-parser?

anniegiang: Right now without body-parser the response when I login/register is the following since `req.body` is `{}` 

`{
    "name": "First name field is required",
    "email": "Email is invalid"
}`

MohamedEl1: also try use `express.json(), express.urlencoded()`

neemiasjnr: It seems that it's because this endpoint uses a GET method, and to pass a body payload we need to use POST method

robinv85: Yes only POST or PUT usually have JSON body otherwise can use query params with GET
you can see in docs this is recommended to only add it on the specific routes where it's needed: https://expressjs.com/en/resources/middleware/body-parser.html

anniegiang: `/api/users/auth` is a POST method

PR #47, opened 2020-04-03
Integration with Auth0 + schema validation

neemiasjnr:
This PR will introduce quite good tools:
- Jest: now we can start adding some tests and run them before commiting using `npm test`
- Eslint: now we can see if our code is properly formatted by running `npm run lint`
- Husky: will block a commit if someone didn't format the javascript file correctly, so our codebase will be cleaner applying the same style everywhere
- Joi: now we can do schema validation for the API, it's much cleaner than adding a lot of ifs to check the properties. Take a look on `./backend/lib/endpoints/schema`
- Auth0: basic auth and create user functions working! the signup via backend is 60% complete


backend/config.js
```diff
@@ -21,14 +14,34 @@ const config = {
     },
   },
   geoService: {
-    host: process.env.GEO_SERVICE_URL,
+    host: `http://${process.env.GEO_SERVICE_URL}`,
```
robinv85, line 17: This environment variable probably shouldn't be called `URL` if you prefix it with `http://` and actually use `hostname` as the environemtn variable
So either use hostname or use a real URL, maybe better consistent with the other settings

PR #52, opened 2020-04-03
User can like and unlike a post once

anniegiang:
- Created `/api/posts/:postId/like` in `post.js` that allows users to Like and Unlike a post once.


backend/lib/endpoints/posts.js
```diff
@@ -73,13 +73,41 @@ router.post(
...
+ * @access Protected
+ */
+router.post(
+  "/:postId/like",
+  passport.authenticate("jwt", { session: false }),
+  (req, res) => {
+    Post.findOne({ _id: req.params.postId }, (err, post) => {
+      if (err) res.send(err);
+      const userId = req.user.id;
+      if (!post.likes.includes(userId)) {
```
robinv85, line 99: yeah so this is probably good for first version
we could also add a table "likes" with { likeId, userId, postId }

robinv85: also small potential improvement but you could use only `indexOf`check probably and if / else without `includes` :)

anniegiang: @robinv85 Thanks for the suggestions! I'll look into them when I get a chance. 👍

robinv85: So if you fetch a list of posts and you want to show the user if they have liked it or not, it can get very big and slow I think. Would be best to index both on like id, post id and user id in that case.

backend/lib/endpoints/posts.js
```diff
@@ -73,13 +73,41 @@ router.post(
     req.body.authorId = req.user.id;
     req.body.postId = postId;
     new Comment(req.body).save().then((comment) => {
```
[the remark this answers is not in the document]
robinv85: yes good point, I wasn't really doing strict reviews but it's better to use async/await as much as possible.
Normally also Express is using the callback pattern, I like koa more because it's made for async/await by default... I am looking at some other routes right now where it's a bit confusing if the middleware should reject or use a callback

backend/lib/endpoints/posts.js
```diff
@@ -157,12 +157,10 @@ router.delete(
...
     try {
-      const removedPost = await Post.findOneAndRemove({
-        _id: req.params.postId,
-      });
-      res.status(200).json({ sucessMessage: "Successfully deleted post" });
+      const removedPost = await Post.findByIdAndRemove(req.params.postId);
+      res.status(200).json(removedPost);
     } catch (error) {
-      res.status(404).send(error);
+      res.send(error); // not sure why error is {} here
```
robinv85, line 163: yeah if something goes wrong you probably want to send all the error info to the client
but that's why we want to change the web framework and take a more systematic approach to error handling, logging, API responses etc.

anniegiang:
Hi everyone, I've refactored the endpoints to use try-catch and async-await. I also created a Like model. I do want to point out the error handling isn't perfect, and I have no idea why that is ... But I'm aware we might migrate to a different stack so either way I can look into it at some point!

robinv85:
Yeah I think this is good to merge now and finish the error handling and such when we add Fastify

PR #57, opened 2020-04-04
Mobile Filter box 

anniegiang:
Not entirely done but there is some structure! I will continue to add some features such as:
 - Location input search
- Change accordion icons
- Add the two buttons at the bottom
- Fix filter tag positioning

-------------**ORIGINAL**-------------
image
Screen Shot 2020-04-04 at 12 40 25 PM

-------------**THIS PR**--------------
Screen Shot 2020-04-04 at 12 40 12 PM
Screen Shot 2020-04-04 at 12 39 54 PM


client/src/assets/data/filterOptions.js
```diff
@@ -0,0 +1,27 @@
+module.exports = {
```
robinv85, line 1: in the frontend you can probably better use `export default ` or `export const`

client/src/components/FilterBox/FilterBox.js
```diff
@@ -1,26 +1,71 @@
...
+        visible={modal}
+        onClose={closeModal}
+        animationType="slide-up"
+        afterClose={() => {
+          console.log("afterClose");
+        }}
+      >
+        <FilterAccordion className="my-accordion" onChange={onChange}>
+          {Object.values(filterOptions).map((filter) => (
+            <FilterAccordionPanel header={filter.label}>
```
robinv85, line 61: probably need to add keys in loops?

PR #59, opened 2020-04-04
added allow to exit the wizard symptoms checker + find help route fix

MohamedEl1:
added allow to exit the wizard symptoms checker + Find Help in the medical page now routes to the right feed page


client/src/pages/SymptomsCheck.js
```diff
@@ -1,4 +1,9 @@
 import React, { useState } from "react";
+import { Button, Radio } from 'antd';
+import { Link, withRouter } from "react-router-dom";
+import "../styles/style.css";
```
robinv85, line 4: Probably better to use StyledComponents :) 
you can just copy that css file into  something like 
```
`const MyComponent= styled.p`
  // put css here
`
```

and then use that component in a JSX tag as html element / React compoment...

client/src/pages/SymptomsCheck.js
```diff
@@ -53,6 +59,12 @@ const Step1 = (props) => {
 
   return (
     <WizardStep>
+      <p className="mb-99">
+        <Link to="medical">
+          {/* By clicking on “X”, users can exit the symptoms questions to medical page */}
+          X
```
robinv85, line 65: probably better to use svg icon (we should use material design, still has to be added I think it's a work in progress)

MohamedEl1: @robinv85 Thanks for the review and suggestion, i'll do that 👍

PR #60, opened 2020-04-05
Add API tools & schema generation

robinv85:
- adds docker-compose config for Swagger Editor & Swagger UI
- initial OpenAPI Spec (v3.0) 
- adds script to generate JSON schema's for validation
- adds some examples for generated schema's

Next step is probably replacing Express by Fastify and using the schema's
+ updating the API spec so the schema's are correct


robinv85:
Also did some research about the current status of the OpenAPI and JSON schema standards but the tooling is currently not very up to date with the latest specs.
So this is Open API v 3.0 and we have to use JSON format or we can maybe use YAML and convert it to JSON with an extra step in the script, maybe good to ask people's preferences... might be easier to edit the API spec. We also need to check compatiblity with JSON schema and validation in Fastify if we add that.

neemiasjnr:
Looks good for now, maybe the next step is migrate to fastify and search how we can use json schema at the routes level and see if it's possible to convert to openapi yaml

robinv85:
@neemiasjnr yeah there is currently a conversion from OpenAPI JSON to JSON schema but that doesn't do very much at the moment, maybe we can try if it works YAML => OpenAPI JSON => JSON Schema and then use it like that

PR #67, opened 2020-04-05
Done Login/Signup UI

shivanshs9:
### Login:
image
### Signup
image

## Pending
- Connecting functionality to backend auth

**PS:** Ignore the ...] icon in the input fields in the screenshot. That's due to LastPass extension.


client/src/components/Icon/Facebook.js
```diff
@@ -0,0 +1,7 @@
+import React from "react";
+
+import IconFb from "~/assets/icons/social-facebook.svg";
+
+export default (props) => {
+  return <img {...props} src={IconFb}></img>;
```
robinv85, line 6: you can probably make it a bit more DRY by having one icon or social-icon component or similar and just render the SVG... for example SVGicon component with svg  / src prop

client/src/components/Input/LabelledInput.js
```diff
@@ -0,0 +1,15 @@
+import React from "react";
+import styled from "styled-components";
+
+const Label = styled.label`
+  color: #425af2;
+`;
+
+export default ({ label, children, style }) => {
```
robinv85, line 8: this one doesn't really render an input but is a wrapper so maybe it can be refactores to a HOC (higher order component)  like `withLabel()` or actually embed the BaseInput component

client/src/components/Input/PasswordInput.js
```diff
@@ -0,0 +1,54 @@
...
+const StyleEye = {
+  float: "right",
+  marginLeft: "-25px",
+  marginTop: "-10px",
+  position: "relative",
+  zIndex: 2,
+  opacity: 0.5,
+};
+
+const EyeIcon = styled(VisibilityIcon).attrs((props) => ({}))``;
```
robinv85, line 25: why empty `{}`, can omit `attrs()` ?

client/src/components/Input/PasswordInput.js
```diff
@@ -0,0 +1,54 @@
...
+          style={inputStyle}
+          placeholder={placeholder}
+          ref={passwordRef}
+          isVisible={isVisible}
+          {...props}
+        />
+        {isVisible ? (
+          <VisibilityOffIcon style={StyleEye} onClick={onToggleVisibility} />
+        ) : (
+          <VisibilityIcon style={StyleEye} onClick={onToggleVisibility} />
```
robinv85, line 49: also small nitpick but if only component changes and props are the same you could also before the return do something like `const iconComponent = isVisible ? VisibilityOffIcon  : VisibilityIcon `

PR #72, opened 2020-04-06
Map implementation

PriyalPatney:
I Have implemented the map component as per manualzuru/FightPandemics.com#18.
this will do the following:-
1. It will fetch the current position of the user, and zoom it.
1. placeholders for the other markers on the map, which can be populated through backend
1. UI for cards which will show the places details


client/src/pages/nearest-hospital/NearestHospital.js
```diff
@@ -0,0 +1,53 @@
...
+          Your Nearest Hospital
+        </h5>
+        <div>
+          <div>
+            <div className="card">
+              <div className="card-body">
+                <h5 className="card-title">United Hospital</h5>
+                <div className="scrollable">
+                  <p className="card-text">
+                    HEALTHSOUTH REHABILITATION HOSPITAL OF CHARLESTON,Lorem
```
robinv85, line 20: so is this dummy data to just test some markup ?

client/src/pages/nearest-hospital/NearestHospital.js
```diff
@@ -0,0 +1,53 @@
+import React, { Component } from "react";
+import NrMap from "./NrMap.js";
+import "./NrMap.css";
+
+class NearestHospital extends Component {
+  render() {
+    return (
+      <div>
+        <br />
+        <h5 style={{ display: "flex", justifyContent: "center" }}>
```
robinv85, line 10: please avoid inline style props and .css files
I realise I also started like that in the hackathon but that was just to get something working fast
so now that we have StyledComponents, try to use that as much as possible :)

PR #75, opened 2020-04-06
Update API spec & schema's (wip)

robinv85:
Updated API schema's to use fluent-api and generate Open API spec automatically
Updated the schema's for users, posts, and geo routes (initial draft version)


robinv85:
@shivanshs9 yeah that's a good suggestion, I actually saw it when I was rewriting the code to use Fastify, but it works the other way around, so currently what I had in mind was "design first" and start with the OpenAPI schema and generate the schema's that can be used for the route validators.

I'm not sure if writing the routes first would be beneficial, we also have to be careful with the compatibility, for example I used OpenAPI v 3, which will soon be compatible with the JSON schema standard but currently there are some small differences so it needs to be converted, Swagger itself is only used for the UI and the Editor with the docker images...

robinv85:
@shivanshs9 do you think it would be a benefit to use fastify-swagger instead?
It might be a little easier to implement, and then we write schema's first, but it actually serves the swagger docs on the fastify server, while we are now only using that in development with docker-compose if we run the Swagger UI container, ... and we have to run the npm script to update the schema's.

robinv85:
We could also use this to generate fastify routes from the API spec https://github.com/seriousme/fastify-openapi-glue

robinv85:
I will try to replace it with https://www.npmjs.com/package/fastify-oas this seems like the most recent Open API documentation generator for Fastify, I will check if we can use it with the docker swagger UI or maybe we can just serve it from the fastify plugin (and enable only in development).
This way it will be easier for other people to collaborate even if they don't know OpenAPI spec, they just need to update the Fastify schema's.

backend/lib/endpoints/schema/geo.js
```diff
@@ -1,4 +1,11 @@
-// const geoSchema = Joi.object().keys({
-//   latitude: Joi.number().min(0).max(90).required(),
-//   longitude: Joi.number().min(-180).max(180).required(),
-// });
+const S = require("fluent-schema");
```
mannykary, line 1: Not really a fan of one letter variable names as it can make code less readable, although if this is the convention for this particular library then I am okay with keeping it this way. If it isn't convention, perhaps we can rename this to `fluentSchema`?

backend/lib/endpoints/users.js
```diff
@@ -6,39 +6,31 @@ const httpErrors = require("http-errors");
...
-      }
-      return {
-        id: result._id,
-        firstName: result.firstName,
-        lastName: result.firstName,
-        email: result.email,
-      };
-    },
-  );
+  app.get("/current", { preValidation: [app.authenticate] }, async () => {
```
mannykary, line 9: Minor: might want to use the same formatting convention as [here](https://github.com/manualzuru/FightPandemics.com/pull/75/files#diff-dfa0a0cc233d1b2aef3fb1520d8b6507R8), to keep code style consistent.

mannykary: Or even adopt this style instead. It doesn't really matter for me, as long as it's consistent.

backend/lib/endpoints/users.js
```diff
@@ -6,39 +6,31 @@ const httpErrors = require("http-errors");
...
+    // todo: get current user from JWT
+    const result = await User.findById("");
+    if (result === null) {
+      return new httpErrors.NotFound();
+    }
+    return {
```
mannykary, line 15: Not too familiar with Mongoose, but is there a reason for not returning the `result` object as is? Does it contain extraneous properties that we would not wish to return as part of the response?

backend/lib/endpoints/schema/geo.js
robinv85: @mannykary heh... yeah that's exactly what I thought
Although this is what I found in the Fastify documentation https://www.fastify.io/docs/master/Fluent-Schema/
I have usually always written full JSON schema documents in external files following the official standards, but the concept of "fluent API's" intends to make it possible to write these schema's much faster, I think that's also why they choose to use a single letter.

Personally I've learned to embrace short variable names in certain contexts such as a function that maps over a collection, since it can reduce cognitive overhead. And certain languages like Golang even consider this best practice: https://github.com/golang/go/wiki/CodeReviewComments#variable-names.

So I thought, if we use separate files for the schema's, and use this "fluent-schema" library, then why not follow the Fastify and fluent-schema documentation and use the short name?

backend/lib/endpoints/users.js
robinv85: Yeah actually we use `Prettier` to automatically format the code style, and I think this depends on the line length, so it's not really something we can really choose in these handlers I think...

robinv85: but maybe we can always put it on the next line...
I was actually looking to move the `{ preValidation } ` hook to the containing function so we can have a secured endpoint with all the subroutes, which is possible in other Node.js frameworks such as express with "nested Routers"... but I didn't find a way to do that yet with Fastify, so definitely open to suggestions!

backend/lib/endpoints/users.js
robinv85: Yes, I just changed thsi since I read in the Fastify documentation that it's recommended, it has nothing to do with Mongoose. 
It's a convention for async routes in Fastify handlers and it's done this way because of how JavaScript workes with Promises and async/await, you can read more in the docs: https://www.fastify.io/docs/latest/Routes/#promise-resolution

robinv85: And ah maybe I misunderstood your question, I was still thinking about some of the responses, sometimes it might not be the best to use a NotFound error but we return an empty list, depending on the request.
For the `result` we definitely need to be careful because we might expose data that we don't want, in this case `/current` only requests your own  logged in user data so I guess we can probably return the full user object.

robinv85:
Yeah @mannykary thanks for the feedback! I would like to add more schema's for all the routes now, but we will have to go through them one by one to make sure the back-end have the right validation, handler logic  etc.

robinv85:
@mannykary went thrugh your comments, seems like Prettier doesn't format the multi-line things automatically so we can still choose to use either format, currently made it consistent with route options on 1 line.
Also it seems like everything works locally now but inside the docker-container i can only access the `/api/version` route so something is still wrong there

backend/lib/endpoints/schema/geo.js
mannykary: Fair enough, as I said I don't have strong opinions about it, especially if it's by convention. Lodash is another library that embraces that use of `_` 😄

backend/lib/endpoints/users.js
mannykary: Yeah, sometimes Prettier and other code formatters don't always catch things like that. I personally prefer the `{ preValidation }` to be on a separate line, but again, I have no strong opinions about it.

backend/lib/endpoints/users.js
mannykary: Makes sense, I was just curious about returning only a few fields in the result rather than all the fields, but from a security standpoint it does make sense to whitelist a few fields that we want to return as part of the response.

backend/lib/endpoints/schema/posts.js
```diff
@@ -0,0 +1,66 @@
+const S = require("fluent-schema");
+
+const getPostsSchema = {
+  querystring: S.object()
```
mannykary, line 4: I think I see why the FluentSchema docs encourage the use of a single `S` letter. It would have made the code more verbose than necessary. This looks good to me 👍

backend/lib/endpoints/schema/posts.js
```diff
@@ -0,0 +1,66 @@
...
+  params: S.object().prop("postId", S.string().required()),
+};
+
+const addCommentSchema = {
+  body: S.object().prop("comment", S.string().required()),
+  params: S.object().prop("postId", S.string().required()),
+};
+
+module.exports = {
+  getPostsSchema,
```
mannykary, line 60: Minor: perhaps these exports could be ordered in alphabetical order?

backend/lib/endpoints/users.js
```diff
@@ -1,36 +1,34 @@
 const httpErrors = require("http-errors");
 
+const { getUserByIdSchema } = require("./schema/posts");
```
mannykary, line 3: Is the path here correct? I think this is supposed to require `./schema/users` since that is where `getUserbyIdSchema` is located, right?

mannykary:
@robinv85 Looks good, just one comment in `users.js` which I believe to be an error in the require path.

mannykary:
@robinv85 Actually, in case my request for changes is a blocker, I'm just going to approve, so that you can go ahead and merge this PR in once you've make your fixes.

backend/lib/endpoints/users.js
robinv85: Alright well now it's consistent, we can still consider later to change it to multi-line as you prefer, I also don't have strong preferences about this

backend/lib/endpoints/schema/posts.js
robinv85: yeah exactly since it can be references multiple times in a schema definition so it makes it better readable and faster to type...

backend/lib/endpoints/schema/posts.js
robinv85: Great idea, I actually prefer that myself, the alternative is "logical" ordering but usually this gets hard to stay consistent.
I propose to add an ESLint rule for this for all the imports / exports and also for object properties and such, but then I think it's better to create a separate pull request for this

robinv85: @MohamedEl1 just so you know, I'll make a PR for this later today

backend/lib/endpoints/users.js
robinv85: No this was a bug, well spotted, fixed now. Actually strange that I didn't notice the name `import` didn't exist. One of these examples where TypeScript would definitely help.

mannykary: Yes, I'm actually a big fan of TypeScript, although it does become a pain to have to rebuild to transpile from TypeScript to JavaScript. We could potentially set up a gulp script combined with nodemon to automatically rebuild on changes though.

PR #76, opened 2020-04-06
feat: authentication api

neemiasjnr:


backend/lib/components/Auth0.js
```diff
@@ -1,47 +1,69 @@
...
+const errorHandler = (err) => {
+  const { response } = err;
+  const { error, message } = response.data;
+  const statusCode = response.status || response.data.statusCode;
+  // eslint-disable-next-line no-throw-literal
+  throw { statusCode, error, message };
+};
+
+const buildOauthUrl = (provider) => {
+  const qParams = [
```
mannykary, line 14: It's probably better to use `querystring.stringify` to pass in an object of query parameters; see https://nodejs.org/api/querystring.html#querystring_querystring_stringify_obj_sep_eq_options

backend/lib/components/Auth0.js
```diff
@@ -1,47 +1,69 @@
...
+const authenticate = async (grantType, payload = {}) => {
+  const body = {
+    grant_type: grantType,
+    audience: `${config.auth.domain}/api/v2/`,
+    client_id: config.auth.clientId,
+    client_secret: process.env.AUTH_SECRET_KEY,
+    ...payload,
+  };
+  return axios
+    .post(`${config.auth.domain}/oauth/token`, qs.stringify(body), {
```
mannykary, line 35: Is it necessary to stringify the body for post? I think the object can be passed in directly.

neemiasjnr: This is recommended by Auth0 documentation, do you have a reason for me to revert that?

neemiasjnr: https://auth0.com/docs/api-auth/tutorials/authorization-code-grant#1-get-the-user-s-authorization

backend/lib/components/Auth0.js
neemiasjnr: Good point! Will do

backend/lib/components/Auth0.js
neemiasjnr: Although I agree that we could remove it, I prefer to leave it as it is for now, because I tested a lot using this code and It's better to refactor after we add some tests

backend/lib/endpoints/posts.js
```diff
@@ -1,53 +1,57 @@
 const httpErrors = require("http-errors");
+const Post = require("../models/Post");
```
robinv85, line 2: why did you remove this? I set it up to have a mongo connection as plugin in fastify so it connects to specific database and can be used on the server instance
This comes straight out of the Fastify documentation https://www.fastify.io/docs/latest/Getting-Started/#getting-started

neemiasjnr: > I can't approve this if it removes the work I did on previous PR with Fastify and Mongoose integration

Really? You merged your PR ignoring eslint rules (clearly wrong and I said you to don't merge) and you can't give an approval because I changed your code?

To be honest doesn't make any sense use a plugin for mongo, because we can have less code connecting directly to the model we want. Why did you change the models folder that was working before? Complicated code should always be avoided. I have more experience in fastify and we need to be mature to do code reviews. In my PR we only have one plugin file because that will be needed for all private endpoints and it makes sense to do that. The mongo connection should always be on server file, and the code should connect to an existent mongo connection via model. Less is more, we need always to remember that.

neemiasjnr: What do you mean by "Furthermore it looks like it's not using the correct Auth0 flow but it should use the one for Single Page Apps"? Did you read the articles I mentioned on slack? SPA flow is very insecure (Implicit Flow) and the authorization flow on the server side is the most secure flow because doesn't expose the keys on the client, and I'm very happy to debate about authentication with you

robinv85: Which ESLint rule was broken? It cannot be committed if it break an ESLint rule, and the rules can be decided by us, it's important to understand why the rules are there in the first place.
You have approved my PR and now you just removed all the work I've done to integrate Mongoose with Fastify in a much better way ? That is unacceptable.

This previous method of starting Mongoose used a global connection and needed an event emitter but I have managed to integrated it with the `createConnection()` method which nicely integrates with promises and async/await, so much better than having to rely on the event emitter separately, and it also supports multiple connections to a specific database if needed.

Furthermore I also added code to auto-load all the Mongoose schema's and instantiated them as a model based on the filename so for my that looked as a very convenient way to integrate that, you have just removed that? 

The Fastify docs clearly staate that they recommend to use plugins as much as possible so your claim "that you have more experience"  is totally irrelevant here and not a good way to discuss a code change.

neemiasjnr: I approved your first version, then you changed the code, I said you to don't merge it and you merged. I won't discuss it furthermore with you, please @manualzuru can you help here?

robinv85: @neemiasjnr this is not true, I even had a PR with 1 commit https://github.com/manualzuru/FightPandemics.com/pull/73
And while my PR clearly states it was working on the database integration with Fastify, this change that you did is totally out of scope of the authentication and authorization.

neemiasjnr:
@shivanshs9 The change is that he commented a lot of files, and merged a not ready commit. If you go to master and run `npm run lint` you will see we have an error and it should be fixed on the PR itself and I told to Robin to change that before he merge, and it didn't happen. Actually the PR about fastify was not following any of the rules you are talking about, he changed the way we connect to mongodb, in this PR actually I'm restoring the previous code that shoudn't be touched by Fastify PR. Our codebase on master is a mess, a lot of commented files. I just approved to be able to work on the authentication without many conflicts. I suggest you to navigate through the master code and see what you think about it. I'm looking forward to see a review to state a problem or bug present in my PR, I'm happy to discuss the problem you see.

backend/lib/endpoints/posts.js
codegoalie: I'm going to jump in here just to try and make progress.

One of the main benefits of only having changes focused on a single feature/bug/etc. in a single PR is to avoid this exact situation where a controversial, unrelated change holds up other changes.

I would propose we (@neemiasjnr) removes the model related changes from this PR and opens a new PR with just those changes. We can focus #76 solely on authentication and have a rigorous debate about the other changes in their own PR. I believe other folks are waiting on the authentication changes here so getting them into `master` as soon as possible is a priority.

How does that sound to everyone?

neemiasjnr:
@manualzuru Can we have a meeting talking about this? It seems this is going far beyond that should have gone.

backend/lib/endpoints/posts.js
robinv85: Sounds good to me, I also had some other remarks about the auth part but guess I;ll keep that for the other PR.

manualzuru:
Hi guys, let's definitely have a chat tomorrow. We are in this together💪🏽 All of you have been doing an amazing job and this type of discussion is not the first nor the last (like pandemics😆), just remember that if we are united we can do amazing things!!!

robinv85:
I've rebased this branch with master and fixed a few small issues and inconsistencies
Maybe we can review this again and merge? There will probably be some more work to finalize the whole auth flow.

backend/lib/components/Auth0.js
```diff
@@ -1,47 +1,69 @@
 const axios = require("axios");
-const {
-  config: { auth },
-} = require("../../config");
-const User = require("./User");
+const qs = require("querystring");
+const { config } = require("../../config");
 
-const oauth = async () => {
```
neemiasjnr, line 7: Why did you remove oauth?

backend/lib/endpoints/posts.js
```diff
@@ -1,11 +1,14 @@
 const httpErrors = require("http-errors");
-const Post = require("../models/Post");
-const Comment = require("../models/Comment");
 
 /*
  * /api/posts
  */
 async function routes(app) {
+  const Post = app.mongo.model("Post");
```
neemiasjnr: This is bad and still don't agree with this change. Why did you change this on Fastify's PR changing the way we interact with mongo? If we are following the rule of "one feature per PR" this shouldn't be changed in the first place without any discussion.

backend/lib/index.js
```diff
@@ -1,22 +1,26 @@
...
+const users = require("./endpoints/users");
+const version = require("./endpoints/version");
 
-function createServer(app) {
+module.exports = function createApp(config) {
+  const app = fastify({
+    logger: true,
+  });
+
+  app.register(require("./plugins/mongoose-connector"), config.mongo);
```
neemiasjnr, line 15: The same, still don't like this change

backend/lib/models/index.js
```diff
@@ -6,7 +6,9 @@ const PATH_TO_SCHEMAS = join(__dirname, "./schemas");
 module.exports = function registerModels(mongoConnection) {
   readdirSync(PATH_TO_SCHEMAS).forEach((fileName) => {
     const filePath = join(PATH_TO_SCHEMAS, fileName);
+    console.log({ fileName, filePath });
```
neemiasjnr: This file is one of the worst in the whole project, using IO to read mongo model for no reason/benefit

backend/lib/plugins/mongoose-connector.js
```diff
@@ -0,0 +1,13 @@
+const fp = require("fastify-plugin");
+const mongoose = require("mongoose");
+
+const registerModels = require("../models");
+
+async function dbConnector(app, config) {
+  const connection = await mongoose.createConnection(config.uri, config.params);
```
neemiasjnr: Unnecessary

backend/server.js
```diff
@@ -1,20 +1,12 @@
...
-
-mongoose.connect(config.mongo.uri, config.mongo.params);
-mongoose.connection.on("error", (err) => {
-  server.log.error(err);
-  process.exit();
-});
+const server = createServer(config);
 
-server.listen(config.server.port, (err, address) => {
+server.listen(config.server.port, "0.0.0.0", (err, address) => {
```
neemiasjnr: Why this "0.0.0.0", this totally unnecessary and you are adding more code for no reason. Don't agree with this change

backend/package.json
```diff
@@ -6,7 +6,7 @@
   "main": "server.js",
   "scripts": {
     "start": "node server.js",
-    "dev": "./node_modules/nodemon/bin/nodemon.js -L server.js",
```
neemiasjnr, line 9: This was added to be compatible with docker-compose on MacOS. Please revert this change

backend/package.json
```diff
@@ -38,10 +38,8 @@
     "fastify-plugin": "^1.6.1",
     "http-errors": "^1.7.3",
     "jsonwebtoken": "^8.5.1",
-    "jwks-rsa": "^1.7.0",
     "mongoose": "^5.9.6",
     "node-cache": "^5.1.0",
```
neemiasjnr, line 42: We need both jwks-rsa and querystring, it's being used on authPlugin. Again, you changing code for no reason without proper context

backend/lib/components/Auth0.js
robinv85: Hmm don't think I removed this on purpose, maybe merge mistake.
But if I look at your branch it wasn't there also? https://github.com/manualzuru/FightPandemics.com/blob/dac1404bc9aa1d1b1af4af1837b2ad5cd0461f5e/backend/lib/components/Auth0.js

The `/api/auth/oauth` route only calls  `Auth0.authenticate()`

neemiasjnr: The oauth is the most important function in this PR, you shoudn't touch something you don't know without communicating before. This is totally disrespect and only shows me the lack of professionalism you have

backend/lib/models/index.js
robinv85: This is a convenient way to auto-load the mongoose models.
I've used similar methods before to remove some boilerplate and often repeating code, I find it also better to inject the dependencies this way for testing and such.
There is even an official Fastify plugin doing the same for plugins https://github.com/fastify/fastify-autoload

robinv85: Also, this is out of scope of the current Pull Request and was already merged to master.

backend/lib/index.js
robinv85: "Not liking something" is not  very useful code review feedback.
You can state that you have a different personal preference, but it's better to back up your opinion with some explanation for your reasoning.
There are many useful resources to improve your code review comments, like this one from Google https://google.github.io/eng-practices/review/reviewer/comments.html

backend/lib/endpoints/posts.js
robinv85: This is out of scope of the pull request, was already approved by you and merged to master.
If you want to discuss this change please open another ticket.

backend/lib/components/Auth0.js
robinv85: @neemiasjnr again, I did not remove this on purpose,. I need to double check what's going on.
Also, this is a pull request, it hasn't even been merged yet, it seems like you are trying to turn things around, it's not needed to call me out like this since there is absolutely no ground to accuse me like this.

backend/package.json
robinv85: querystring is a Node.js module https://nodejs.org/api/querystring.html
I didn't find jwks-rsa in use, maybe removed that oauth method

backend/lib/components/Auth0.js
robinv85: See this was renamed to 
` const authenticate = async (grantType, payload = {}) => { 
...
eturn axios
    .post(`${config.auth.domain}/oauth/token`, qs.stringify(body), {
```
which is making the same request to Auth0

PR #77, opened 2020-04-07
Style Posts in Feed

anniegiang:
Will need to fix some positioning, clean up some code, and replace the icons but the overall structure is there! 

Will work on state and might use Context to reduce prop threading

Screen Shot 2020-04-07 at 12 04 20 AM
Screen Shot 2020-04-07 at 12 04 30 AM


client/src/pages/Feed.js
```diff
@@ -1,17 +1,74 @@
...
+    setModal(isOpen);
+    setActivePanel(panelIdx > -1 ? `${panelIdx}` : null);
+  };
+
+  const handleQuit = (e) => {
+    e.preventDefault();
+    setModal(false);
+    setActivePanel(null);
+    setSelectedFilters({});
+    setLocation("");
```
robinv85, line 30: it could be nice to work with "initial state" and reset to that, it used to be a React pattern few years ago with `initialState()` and then there were more class components but now with hooks I am not sure if the best way is to use one initial state object or have a separate hook for every state property, seems it could also work better with `userReducer` https://reactjs.org/docs/hooks-reference.html#usereducer

anniegiang: Good point! I haven't used `userReducer` before but it sounds like a good fit. There's also a pattern to use dynamic hooks for specific functionalities like toggling state or handling input. I'll look into the options

PR #78, opened 2020-04-07
Show answers selected by user on results page

mutebironald:
- Add a results page to display the user answers.
- the recommendation bit is missing as per the mockup. This is from the user answers, need guidance on this.
- The Done button will redirect to the feed after the feed feature is completed.

Am currently fixing the styles on the Done button and digits.

### DIFF
--- a/client/src/pages/SymptomsCheck.js
+++ b/client/src/pages/SymptomsCheck.js
@@ -12,7 +12,7 @@ import {
 } from "../components/StepWizard";
 
 const INITIAL_STATE = {
-  answers: [],
+  // answers: []
 };
 
 const Welcome = (props) => {
@@ -48,6 +48,7 @@ const Welcome = (props) => {
 const Step1 = (props) => {
   const onSelectAnswer = (answer) => {
     props.update("age", answer);
+    console.dir(props);
     props.nextStep();
   };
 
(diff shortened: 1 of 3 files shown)


client/src/pages/SymptomsCheck.js
```diff
@@ -305,16 +305,9 @@ const Step7 = (props) => {
...
-    setState({ ...state, updatedAnswers });
-    if (key === "careFacility") {
-      localStorage.setItem(
-        "symptomsCheckAnswers",
-        JSON.stringify(updatedAnswers),
-      );
-    }
+    setState({ ...state, [key]: value });
   };
+  localStorage.setItem("symptomsCheckAnswers", JSON.stringify(state));
```
robinv85, line 310: Hmm looks good that you simplified the `updateAnswers` logic because we only currently need the answers "map" ( could also use Map() instead of object) but updating local storage here on every render of the page seems wrong?
Before I only did this at the last step which was also kind of a temporary solution, but this might not be good either, maybe update after every call to `updateAnswers`? 
And since it's a side effect it probably belongs in a `useEffect()` hook https://egghead.io/lessons/react-store-values-in-localstorage-with-the-react-useeffect-hook

mutebironald: @robinv85 , do we even need the localstorage bit? Was thinking that since data is being stored in state, we could do away with this. Your thoughts?

PR #80, opened 2020-04-07
Drawer menu style tweaks

luxumbra:
A few style updates and added a close icon so the menu matches the design.
Also added a root font size for relative units.

image

image

### DIFF
--- a/client/src/index.scss
+++ b/client/src/index.scss
@@ -2,7 +2,7 @@
 @import "./styles/theme.scss";
 
 html {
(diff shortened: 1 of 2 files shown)


luxumbra:
There's a clear difference in weight of the type but I have used the same font and weight as per the figma designs. I have tried different weights and `bold` but this is the closest achieved. Perhaps my OS/Browser (Pop_OS/Chrome)?

luxumbra:
Thanks guys. :)

PR #87, opened 2020-04-09
Incorporate useReducer and useContext to mange Feed state

anniegiang:


client/src/reducers/feedReducers.js
```diff
@@ -0,0 +1,41 @@
...
+  TOGGLE_STATE,
+  SET_VALUE,
+} from "../actions/feedActions";
+
+export const feedReducer = (oldState, action) => {
+  let newState = Object.assign({}, oldState);
+  const { type, key, value } = action;
+  switch (type) {
+    case TOGGLE_STATE:
+      newState[key] = !newState[key];
```
robinv85, line 14: I would probably write it a little shorter too like `return { ...state, [key]: !state[key] };`

client/src/reducers/feedReducers.js
```diff
@@ -0,0 +1,41 @@
...
+  }
+};
+
+export const optionsReducer = (oldState, action) => {
+  const { option, label } = action.payload;
+  let newState = Object.assign({}, oldState);
+  newState[label] = newState[label] || [];
+  switch (action.type) {
+    case ADD_OPTION:
+      newState[label].push(option);
```
robinv85, line 30: if you clone the `oldState` here and it keeps reference to the `oldState[label]` array your should also immutably update the nested object and arrays, so not using `[].push()` but e.g. 
```
return { ...oldState, [label]: [ ...oldState[label], option ] };
```
more examples here https://redux.js.org/recipes/structuring-reducers/immutable-update-patterns

anniegiang:
@robinv85 made the changes!

robinv85:
Ok 👌🏼

Issue #88, opened 2020-04-09
Add correct fonts and homepage tweaks


luxumbra:
1. The correct fonts for the App are Poppins and Work Sans, neither are being used at present.
2. The Menu icon is an ellipsis but in the designs it's showing a hamburger. It's also not vertically aligned with the logo. 
3. Vertical white space on the homepage needs tweaking

luxumbra:
1. I've added in imports the same way we were importing Roboto (`typeface-roboto`) with `typeface-poppins` and `typeface-work-sans`. Poppins is working a treat but despite `typeface-work-sans` being installed in `node_modules`, I am getting the following error:
`./src/index.js
Module not found: Can't resolve 'typeface-work-sans' in '/app/src'`
2. I've changed the icon for the menu and is now the correct color and aligned with the logo.
3. I've tweaked the spacing on the homepage to match the designs.

robinv85:
@luxumbra seems like this was also updated in #89 see https://github.com/manualzuru/FightPandemics.com/pull/89/files#r406724649

Any idea about current status? Seems to me that Roboto can be removed, and that "Poppins" is now added as a Google font, so maybe can be removed from npm depencencies?

luxumbra:
I have some tweaking to do after feedback from design, then the other wizards will need to have the new styled components that I've created added to them, which should bring them all into line with designs. I'll remove roboto and the poppins dependency. Thanks

luxumbra:
The work this issue was created for has been completed and merged, so I am closing this. I'll open a new issue for the work required for integrating the styled components I created for the wizards. Any additional font / theme oriented work can be done under issue #74. @robinv85 @anniegiang does this seem reasonable?

PR #89, opened 2020-04-10
Feed Frontend: style fixes and render nested comments

anniegiang:
- Fixed sizing and positioning of elements across the Feed
- The Create Post icon floats while scrolling the feed
- Fixed font weight issues by adding more weights when importing Poppins
- Incorporated copy-to-clipboard functionality using a react library (onClick it will show a modal alerting "Link Copied!")
-  Installed `antd` in order to use the Comment component
- Added like/unlike functionality
- The count for share will increase only once onClick
- Clicking comments will show comments
- Rendered nested comments (still need to style them)

WIP:
- Styling comments
- Add functionality to reply to comment
- Once the backend for auth, posts, comments, and likes are done I'll work on API integration

### DIFF
--- a/client/src/index.js
+++ b/client/src/index.js
@@ -2,6 +2,7 @@ import React from "react";
 import ReactDOM from "react-dom";
 import "typeface-roboto";
 import "antd-mobile/dist/antd-mobile.css";
+import "antd/dist/antd.css";
 
 import "./index.scss";
 import App from "./App";
--- a/client/src/styles/theme.scss
+++ b/client/src/styles/theme.scss
@@ -1,8 +1,8 @@
 // Import Colors
 @import "./colors.scss";
-@import url('https://fonts.googleapis.com/css2?family=Poppins:wght@400;700&display=swap');
+@import url("https://fonts.googleapis.com/css2?family=Poppins:wght@200;300;400;500;600;700&display=swap");
 
 $body-bg: $white;
 $primary: $royal-blue;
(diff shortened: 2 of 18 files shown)


client/src/index.js
```diff
@@ -2,6 +2,7 @@ import React from "react";
 import ReactDOM from "react-dom";
 import "typeface-roboto";
```
robinv85, line 3: out of scope here but seems like we still have to remove this one

client/src/styles/theme.scss
```diff
@@ -1,8 +1,8 @@
 // Import Colors
 @import "./colors.scss";
-@import url('https://fonts.googleapis.com/css2?family=Poppins:wght@400;700&display=swap');
+@import url("https://fonts.googleapis.com/css2?family=Poppins:wght@200;300;400;500;600;700&display=swap");
```
robinv85, line 3: didn't we also have an npm dependency for the Poppins font?

anniegiang: Yea it's called `typeface-poppins`. Before adding the weight values the bold text were off so not sure if the dependency works ...

luxumbra: I'm glad to see @anniegiang using the @import for the google fonts as I've just added Work Sans and DM Sans to the index.scss file - I'll move it to theme.scss. I'll remove poppins dependency in my work for #88

PR #90, opened 2020-04-10
added map implementation as per manualzuru/FightPandemics.com#18

PriyalPatney:
I Have implemented the map component as per manualzuru/FightPandemics.com#18.
this will do the following:-
1. It will fetch the current position of the user, and zoom it (can be changed).
1. populating nearby hospital location on the map with markers.
1. UI for cards with hospital details.


client/src/styles/NrMap.css
```diff
@@ -0,0 +1,27 @@
+.card {
```
robinv85, line 1: please avoid .css files and use Styled Components instead (see other existing code for reference)

client/src/pages/NrMap.js
```diff
@@ -0,0 +1,165 @@
...
+
+              createMarkers(results);
+              pagination.hasNextPage &&
+                function () {
+                  pagination.nextPage();
+                };
+            },
+          );
+
+          function createMarkers(places) {
```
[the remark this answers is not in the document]
PriyalPatney: createMarkers(places) is called for the nearbySearch() which will place the markers on the map.

client/src/pages/mapexample.js
```diff
@@ -0,0 +1,136 @@
...
+      function (results, status) {
+        if (status !== "OK") return;
+        //locations are in this variable
+        console.log(results);
+        //this is not working .........
+        //Cannot read property 'setState' of undefined
+        //  this.setState({rec: results})
+      },
+    );
+
```
PriyalPatney, line 83: following comments are irrelevant, missed to remove it before the commit.

client/src/pages/mapexample.js
```diff
@@ -0,0 +1,136 @@
...
+        location: {
+          lat: this.state.latitude,
+          lng: this.state.longitude,
+        },
+        radius: 3000,
+        type: ["hospital"],
+      },
+      function (results, status) {
+        if (status !== "OK") return;
+        //locations are in this variable
```
PriyalPatney, line 76: here we can see the object populated with the nearby query. but setState is not working(scope issue i guess)

j-collier: I think it's because you're trying to call this function within the event listener above at line 48

robinv85: ah probably need to bind "this" to the react component
what you can try is instead of `function(results, status) {}` type `(results, status) => {}` and it will have "this" binding to the parent context

PR #91, opened 2020-04-10
Removed unused imports & variables and Made small changes on About page

linnnruoo:
- Removed all the unused imports and variables because of the warnings thrown, pardon me if I'm a little too OCD about this 😢 
- Made small ui and styling changes on the About page in `About.js` so that its responsive
- Added `<Emoji />` component, because of the warnings thrown for using emojis in text too
- Added `cursor: pointer` to the social icons button on the login page. Since its a button, I guess pointer is good to represent that it's clickable

Apr-10-2020 13-35-44

### DIFF
--- a/client/src/components/Accordion/FilterAccordion.js
+++ b/client/src/components/Accordion/FilterAccordion.js
@@ -1,4 +1,4 @@
-import styled, { css } from "styled-components";
+import styled from "styled-components";
 import { Accordion } from "antd-mobile";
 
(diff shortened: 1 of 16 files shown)


linnnruoo:
> hah this is great :) was annoying me also! Thanks @linnnruoo
> Also I wonder if the design of the about page is even like it's supposed to? (might need to check with a designer)

Actually I was a little puzzled too because I could not find anything on figma :')

PR #92, opened 2020-04-10
Fix docker compose formatting and volumes

mannykary:
@robinv85 I based this off of your `api-spec-schemas` branch. 

I fixed up the quotes; I do not believe that they are necessary for the volume mounts, based on my experience at work with Docker Compose files. I also added the quotes for port mappings, although there seems to be conflicting documentation about this as you've no doubt seen. I tend to stick to their recommendation of always quoting them as a force of habit to avoid accidentally running into that weird base-60 gotcha.

I also fixed the volume mounts causing issues with `node_modules`. I don't think the explicit `node_modules` mount is necessary as it is already contained within the first volume mount (`./client:/app` for the frontend, and `./backend:/app` for the backend). I did a quick test of this by entering the bash shell for each of these containers, and then created an empty file in the `node_modules` folder on my host. I was able to see the empty file from the container. Do note that `npm install` should be run on the host in both the `client` and `backend` folders to create the `node_modules`, prior to running `docker-compose up`. It might be worthwhile to note this in the readme.


robinv85:
@mannykary probably need to rebase or merge with master since that other branch got merged by now

mannykary:
Yeah no worries, I'll rebase/fix that up when I get a chance, thanks for the heads up.

mannykary:
I will also look into the backend container returning 404 sometime today.

mannykary:
I think there might be issues with my change, as I just found out that some NPM dependencies may not necessarily be [cross-platform](https://stackoverflow.com/questions/38425996/docker-compose-volume-on-node-modules-but-is-empty#comment74662821_43728476) (i.e. installing a dependency built on OS X may not work when used by the Linux host)

robinv85:
Yeah that’s what I thought so you can either just install them separately inside & outside the contained if you sss as nt to run both, or you can install in the container and mount the folder to be able to debug dependencies from outside the container ...

mannykary:
I think your original volume mount syntax for the `node_modules` was correct, as it excluded mounting the host `node_modules` within the container, if it exists.

I think we should add an npm command to run `npm install` within the docker. I am thinking of adding this to the `package.json` for both the backend and client:
```
// /backend/package.json
"install-docker": ""docker-compose run backend-service npm install"

// /client/package.json
"install-docker": ""docker-compose run client npm install"
```

And then update the Readme to mention this caveat. Thoughts?

robinv85:
@mannykary great idea, was thinking something similar but wasn't sure what's possible (and I suppose this is a common pattern that is often used so we probably don't have to reinvent the wheel).

README.md
```diff
@@ -10,13 +10,28 @@ Currently, there is no easy way to find all the information at the local and glo
...
+NPM dependency installs may not necessarily be cross-platform. There are two ways that you can install the dependencies
+within the container:
+
+### Backend
+Run `docker-compose run backend-service npm install`, or `cd` into the `backend` directory and run `npm run install-docker`.
+
+### Client
+Run `docker-compose run client npm install`, or `cd` into the `client` directory and run `npm run install-docker`.
+
+Be sure to also commit any changes to the `package-lock.json` so that dependencies used by third-parties are locked to
```
robinv85, line 36: *too? :p

mannykary:
This branch has diverged a bit from master, so I'm going to close this PR and open up a new one that's closer to master.

README.md
mannykary: Haha, spacing is a little weird, I'll try to fix that up.

PR #96, opened 2020-04-10
Feed Frontend: style comments, change px to rem , mock commenting 

anniegiang:
- Styled comments
- Changed `px` values to `rem`
- Commenting updates UI
- Imported several SVG icons
- Clicking on icons renders a clicked icon
- Updated styles upon discovering how to get exact css values from Figma 😅

WIP:
- Integrate API (@robinv85 Is the backend done for auth, posts, comments, and likes? )
- Add liking and reply functionality to comments
- Add a like icon in the comments
- Truncate post text
- Clean up code

Screen Shot 2020-04-10 at 4 20 45 PM
Screen Shot 2020-04-10 at 4 21 22 PM


robinv85:
@anniegiang API isn't completely done at the moment, partially implemented
Auth has to be verified with the settings on Auth0, including signup. login with email / password and using social media accounts
The posts has a initial implementation that also needs to be double checked, not sure who made the original one but I used the same model and added a schema (you can currently run it locally without docker and go to localhost:8000/documentation to see all the API routes)
For comments and likes we still need to look at that, so plenty of work to do, will be for tomorrow for me :)

anniegiang:
@robinv85 Ok sounds good! Before moving to Fastify I had created the endpoints for Post, Comment, Like already but there were issues with error handling. So I haven't looked at them since we left Express

robinv85:
@anniegiang ah yes right I remember, so probably I only removed `POST /posts/:postId/like` because I thought that wasn't really RESTful and there's probably a better way but the others should be more or less the same I suppose.

PR #97, opened 2020-04-11
refactor and fix some bug of login and signup

markthien:
- added reduc-thunk
- added react-alert
- added react material-ui
- fixed and refactor bug on Login.js
- modified actions


client/src/pages/Login.js
```diff
@@ -1,16 +1,21 @@
-import React from "react";
+import React, { useState } from "react";
 import styled from "styled-components";
 import { Button, Flex, WhiteSpace } from "antd-mobile";
 import { Link } from "react-router-dom";
+import { makeStyles } from '@material-ui/core/styles';
```
robinv85, line 5: I think we would better avoid this
because currently we are using Ant design + Styled components
The aim for the web app is mostly to focus on mobile, so I think it would be better not to use all these different libraries (currently we have Ant Deisgn, Ant Mobile, Material UI ... and probably still Grommet in there), plus our own base components. We have to take one approach and stay consistent.

client/src/pages/Login.js
```diff
@@ -1,16 +1,21 @@
...
+import TextField from '@material-ui/core/TextField';
+import { validateEmail } from '../utils/common.js';
```
robinv85, line 7: good to see you have added email validation, we can also use this for the Need Help and Offer Help components with email inputs

client/src/pages/Login.js
```diff
@@ -1,16 +1,21 @@
...
+import { useAlert } from 'react-alert';
+import { PASSWORD_MIN_LENGTH } from '../config';
+import { loginWithEmail, signup } from '../actions/authActions';
+import { useDispatch } from "react-redux";
```
robinv85, line 11: my initial thoughts about redux were that "we might not need it" and better avoid it to avoid complexity and to increase performance.
For now, I guess it's good to keep it simple and make sure everything works well for the MVP, we could later see if we can just use the React Context API and if we really need a state management solution.

client/src/pages/Login.js
```diff
@@ -56,10 +61,71 @@ const SocialButton = styled(Button).attrs((props) => ({
...
+
+  const handleSignup = (evt) => {
+    evt.preventDefault();
+    if (!validateEmail (email)) {
+      alert.show('Invalid email address!', {
+        type: 'error',
+      });
+      return;
+    }
+    if (password.length < PASSWORD_MIN_LENGTH) {
```
robinv85, line 121: we have to make sure we are using the same validation rules on backend as well

markthien: yea sure. can you provide me the backend validation rules?

robinv85: seems like it's Auth0 validation rules

robinv85: Currently set to "Good" password strength settings : 

    No more than 2 identical characters in a row
    Special characters (!@#$%^&*)
    Lower case (a-z), upper case (A-Z) and numbers (0-9)
    Must have characters in length
    Non-empty password required

robinv85: @mannykary any idea if this is ok? We can also set it to "excellent" and then there is an extra rule:
No more than 2 identical characters in a row

client/package.json
```diff
@@ -25,6 +25,8 @@
     "polished": "^3.5.1",
     "prettier": "^2.0.2",
     "react": "^16.13.1",
+    "react-alert": "^7.0.0",
```
robinv85, line 28: Also, @markthien not that I want to annoy you about changing everything but why did you add an extra component and theme to show the alerts?
There is also an alert component in Ant Design https://ant.design/components/alert/ 
This will probably also be good enough?
It's just that we should try to use less dependencies, so try to use Ant Design / Ant Design Mobile as much as possible... sorry if that wasn't clear enough, maybe we need to update that in the documentation.

client/src/actions/authActions.js
```diff
@@ -12,3 +15,27 @@ export const submitEmail = (userData, history) => (dispatch) => {
       }),
     );
 };
+
+export const loginWithEmail = (payload) => {
+  return (dispatch) => {
+    console.log('payload', payload);
+    return axios.post('/api/users/signup', payload, {}).then(response => {
```
robinv85, line 22: looks great now, only one small change
currently we have only `/api/auth/authenticate` for both login & signup with email and password
I would like to review this and maybe split them up, we can also provide more data when signing up first time perhaps.

robinv85:
Maybe @anniegiang and @luxumbra can also double check this
not unimportant that it adds more redux 
Also, what do you think about toaster of ant mobile and alert of ant desktop, can we just choose one?

PR #98, opened 2020-04-11
Update README with tech stack

anniegiang:
I noticed there were confusions about our tech stack so I decided to update the README. As we have new people join and help we can direct them to the README so they are familiarized. Feel free to make any edits!


manualzuru:
GREAT!!!

PR #100, opened 2020-04-11
User profile

CeesWang:
Done
- Finished styling the user profile,
- created edit-profile / edit-account routes and components

wip
- styling on edit-profile and edit-account components
- handle data flow for my activity in user profile


client/src/pages/EditProfile.js
```diff
@@ -0,0 +1,40 @@
+import React from "react";
+import { Popover, WhiteSpace } from "antd-mobile";
+import { MenuOutlined, EditOutlined } from "@ant-design/icons";
```
robinv85, line 3: I think we should only use material UI icons? might need to check with designer to be sure but when I asked that's what they said.
I just would like to limit the number of dependencies so try to stick to one icon library if possible.

client/src/pages/EditProfile.js
```diff
@@ -0,0 +1,40 @@
...
+  return (
+    <div>
+      <div style={initialsStyle}>{getInitials(firstName, lastName)}</div>
+    </div>
+  );
+};
+
+//styling
+
+const initialsStyle = {
```
robinv85, line 30: try to used Styled Components  if possible

client/src/pages/Profile.js
```diff
@@ -0,0 +1,227 @@
...
+const firstName = "Cees";
+const lastName = "Wang";
+const about =
+  "  Lorem ipsum dolor sit amet, consectetuer adipiscing elit, sed diamnonummy nibh euismod tincidunt ut laoreet dolore magna";
+const email = "ceeswang@Test.com";
+const location = "NY, USA";
+const needHelp = true;
+const Item = Popover.Item;
+
+function getInitials(firstName, lastName) {
```
robinv85, line 19: could be nice to move to utils or some shared file in profile components if used more than once

client/src/pages/EditAccount.js
```diff
@@ -0,0 +1,180 @@
...
+  "  Lorem ipsum dolor sit amet, consectetuer adipiscing elit, sed diamnonummy nibh euismod tincidunt ut laoreet dolore magna";
+const facebookURL = "http://facebook.com";
+const twitterURL = "http://twitter.com";
+const githubURL = "http://github.com";
+const linkedinURL = "http://linkedin.com";
+const personalURL = "http://personal.com";
+
+const editAccount = true;
+
+function getInitials(firstName, lastName) {
```
anniegiang, line 17: I need this function too for Posts! I'll make this a common function we can use. I'll create a file called `userProfile.js` in `client/src/utils` and export this function in there.

client/src/pages/EditProfile.js
CeesWang: Yes I talked to Yu and managed to use the exact same icons now.

anniegiang: There's a way to grab icons on Figma so that we can be sure we're using the correct icons as well being less dependent on external libraries - I added instructions in the README. @CeesWang let me know if you have any questions with that!

CeesWang:
Completed     
- Changed most inline styling to styled Components
- Created edit-profile / edit-account routes and components
- Finished edit profile styling
- Finished account info styling 


wip
- form validations 
- user profile comments sections

anniegiang:
@CeesWang Hi Chen, can you resolve the merge conflicts?

anniegiang:
@CeesWang Please resolve the merge conflicts.

CeesWang:
Merge conflicts should be resolved let me know if there's anything else

client/src/pages/EditAccount.js
```diff
@@ -0,0 +1,226 @@
...
+  padding-left: ${(props) => props.paddingLeft || ""};
+  margin-top: ${(props) => props.marginTop || "1.5rem"};
+  font-size: ${(props) => props.size || ""};
+  font-weight: ${(props) => props.weight || ""};
+};
+`;
+
+export default function EditAccount(props) {
+  // dummy data props,context, redux etc
+  const firstName = "Cees";
```
robinv85, line 20: ok I guess before merging you could replace this with the mapped props from redux user

client/src/pages/EditAccount.js
```diff
@@ -0,0 +1,253 @@
+import React from "react";
+import { useForm } from "react-hook-form";
```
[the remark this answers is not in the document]
CeesWang: I used useForm instead of ant form because I noticed that ant form uses state which results into a lot of re-renders instead of useForm which only uses one.  I can change it to ant form if necessary let me know

client/src/pages/EditAccount.js
anniegiang: @CeesWang Regarding your question from Slack - you could store dummy data like this in Redux and then connect this component with Redux via `mapStateToProps` to consume the different props/dummy data from redux

client/src/pages/EditProfile.js
```diff
@@ -0,0 +1,107 @@
...
+  );
+}
+
+const mapStateToProps = (state) => {
+  return {
+    user: state.user,
+  };
+};
+
+export default connect(mapStateToProps)(EditProfile);
```
anniegiang, line 107: @robinv85 Will this be fine for now until we have a way for testing HOCs?

robinv85: I'd say we don't have to worry about that at the moment.
A convention that is sometimes used is to export both the component and the connected component, maybe one named and the other default... 
We could use a naming convention in that case such as "Component" and "ComponentContainer"  ... because otherwise it's not transparent to know what is exported from a file

This presentational / container component pattern was pretty much standard for redux apps for a few years but nowadays with hooks many project don't use redux... as you see here in the recently updated article https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0

If we look at Redux docs they still suggest to use separate folders like `/components` and  `/containers` https://redux.js.org/basics/usage-with-react#designing-component-hierarchy

PR #104, opened 2020-04-11
Post likes

### LINKED ISSUE #102: Like / unlike a post API

robinv85:
Closes  #102
like & unlike a post
also updated some details of the post schema's


backend/lib/endpoints/posts.js
```diff
@@ -98,6 +100,52 @@ async function routes(app) {
       return updatedPost;
     },
   );
+
+  app.put(
+    "/:postId/likes/:userId",
+    { preValidation: [app.authenticate], schema: likeUnlikePostSchema },
+    async (req) => {
+      const { postId, userId } = req.params;
+      // todo: get user id from JWT
```
anniegiang, line 109: with passport I think you can get that from `req.user.id` for the current user. or did we remove passport?

backend/lib/endpoints/posts.js
```diff
@@ -98,6 +100,52 @@ async function routes(app) {
       return updatedPost;
     },
   );
+
+  app.put(
+    "/:postId/likes/:userId",
```
robinv85, line 105: I guess maybe I need to rename either the schema or the endpoint because I changed schema to `"likedBy"` for clarity and now it doesn't match the route anymore

backend/lib/endpoints/posts.js
robinv85: Yep that’s right, we had that with express but for Fastify we now have a more custom auth middleware which closely resembles passport. Still need to check if the auth is fully functional... wasn’t completely working last time i tried.

backend/lib/endpoints/posts.js
```diff
@@ -98,6 +100,52 @@ async function routes(app) {
       return updatedPost;
     },
   );
+
+  app.put(
```
shivanshs9, line 104: I'm confused a bit... Liking a post is a PUT request while unliking is a DELETE request?
I'd say DELETE is a better contrast to POST method. :thinking:

backend/lib/endpoints/posts.js
shivanshs9: By not matching the route, do you mean when generating the schema or by fastify router?

robinv85: @shivanshs9 I was referring to the name `/likes` compared to the field `/likedBy` ... and it also returns both `likes` and `likesCount` but I guess that's ok, updated now

backend/lib/endpoints/posts.js
robinv85: yes, it's an attempt to stay as close to the REST architectural principles as possible.
POST should create a new item on every request, and it is "non-idempotent" which means if you make the same request 3 times it will create 3 new items.
Whereas, PUT and DELETE are "idempotent" which means if you make the same request 3 times, it will not have a different outcome because it can only change something once.
Here however because of using the Mongo query syntax to check if a post has already been liked I decided to return a `400 Bad Request` in case a post is already liked and you try to like it again ... which may be not 100% according to "REST" but it's just hard to know and follow all these principles perfectly :) 

More info and examples here: https://stackoverflow.com/questions/5665893/the-rest-way-to-check-uncheck-like-unlike-favorite-unfavorite-a-resource

robinv85: Actually, maybe I should return a 404 if you try to "delete" a like that doesn't exist, which may be even closer to REST & HTTP principles.

Issue #22, opened 2020-03-31
Organization profile - Backend


robinv85:
- [x]  backend / API support to get organistion info

PriyalPatney:
I’d like to write this API.

pavanmt:
Will start working on API dev.

MohamedEl1:
Sure

pavanmt:
PR:
https://github.com/manualzuru/FightPandemics.com/pull/68

mannykary:
@pavanmt Hope you don't mind me picking up this issue. I was asked by @robinv85 to work on implementing this in Node.js/Fastify, rather than Python/Flask as you had done. If you would like to work with me on this, let me know.

PR #105, opened 2020-04-12
Organization API backend

mannykary:
Organization API, implemented using Fastify. Issue #22


backend/lib/endpoints/organizations.js
```diff
@@ -0,0 +1,95 @@
...
+    },
+    async (req, reply) => {
+      // TODO: make sure user can only update organizations they own
+      const organization = await Organization.findById(
+        req.params.organizationId,
+      );
+      if (organization === null) {
+        return reply.send(new httpErrors.NotFound());
+      }
+      Organization.schema.eachPath((key) => {
```
mannykary, line 74: @robinv85 something I noticed when I had the logic similar to what was in the the `posts#patch` endpoint...any new keys that are set in the patch operation were skipped over and not saved to the document. So I had to alter the logic here to iterate over each attribute in the `Organization` model, and check that the attribute key exists in the request body. Just wanted to give you a heads up, so that we should fix it in the `posts` endpoint as well.

Since this pattern will be used by all patch operations, perhaps we could add a reusable function in `utils.js`?

mannykary: Actually, I just realized that the issue was that the code in `posts#patch` is just checking that the key exists in the BSON document. Removing that check should be sufficient. See my latest [commit](https://github.com/manualzuru/FightPandemics.com/pull/105/commits/3d8766a1486d9daaedf44450b1d14e0f57c0a4e5).

backend/lib/endpoints/organizations.js
```diff
@@ -0,0 +1,95 @@
...
+      preValidation: [app.authenticate],
+      schema: getOrganizationsSchema,
+    },
+    async (req, reply) => {
+      const { ownerId } = req.params;
+      const filter = ownerId ? { ownerId } : {};
+      const sortedOrganizations = await Organization.find(filter).sort({
+        name: 1,
+      });
+      return reply.send(sortedOrganizations);
```
anniegiang, line 41: Not super familiar with Fastify but is there a reason why we're sending instead of returning `sortedOrganizations` alone?

backend/lib/endpoints/organizations.js
```diff
@@ -0,0 +1,95 @@
...
+    async (req, reply) => {
+      // TODO: make sure user can only update organizations they own
+      const organization = await Organization.findById(
+        req.params.organizationId,
+      );
+      if (organization === null) {
+        return reply.send(new httpErrors.NotFound());
+      }
+      Object.keys(req.body).forEach((key) => {
+        if (organization[key] !== req.body[key]) {
```
anniegiang, line 75: what do you think about also checking if `organization[key]` exists? So something like: 

`if (organization[key] && organization[key] !== req.body[key]) `

though now that I think of it, it might not be a huge difference

backend/lib/endpoints/organizations.js
```diff
@@ -0,0 +1,95 @@
...
+
+  app.get(
+    "/:organizationId",
+    {
+      preValidation: [app.authenticate],
+      schema: getOrganizationSchema,
+    },
+    async (req) => {
+      const result = await Organization.findById(req.params.organizationId);
+      if (result === null) {
```
anniegiang, line 53: not major but we could make this a one liner, maybe something like 

`!result ? new httpErrors.NotFound() : ""`

backend/lib/endpoints/organizations.js
shivanshs9: @anniegiang actually, I think the check is pretty important since we're directly using client's request body which may have more number of keys (like a malicious attempt) which will cause server error.

shivanshs9: Also, I was thinking instead of directly assigning the request body keys to the model keys, it's better to first filter out only the desired keys by maybe validating the schema.
Otherwise, what if a malicious user changes a private (like maybe created timestamp) field of the Organization object?

backend/lib/endpoints/schema/organizations.js
```diff
@@ -0,0 +1,57 @@
...
+    .prop("ownerId", S.string())
+    .prop("skip", S.integer()),
+};
+
+const updateOrganizationSchema = {
+  body: S.object()
+    .prop("address", S.string())
+    .prop("androidUrl", S.string())
+    .prop("description", S.string())
+    .prop("email", S.string().required())
```
robinv85, line 31: i don't think update has required properties since these are PATCH requests that only use what's in the request?

backend/lib/models/schemas/Organization.js
```diff
@@ -0,0 +1,71 @@
+const { Schema } = require("mongoose");
+
+const OrganizationSchema = new Schema({
+  address: {
+    required: false,
```
robinv85, line 5: maybe can just omit required if false

backend/lib/endpoints/organizations.js
robinv85: yeah commented on that link

robinv85:
> 
> 
> I believe there should be schema validation, like maybe using [AJV](https://ajv.js.org/). Also if possible, there should be a mechanism of cleaning the request data rather than directly accessing it when updating records.

@shivanshs9 there is validation built into Fastify which we are using for all the API requests, it's using the fluent-schema library to generate schema's and internally Fastify uses AJV.

backend/lib/endpoints/organizations.js
mannykary: So, I originally had the check for `organization[key]` but the way MongoDB works is that if you don't set some of the fields, the keys won't even exist in the record. So for instance, if you do a POST with the following body (shortened for clarity):

```
{
  "name": "Test org",
  "email": "test@example.com"
}
```

And then later you do a PATCH to set another field:
```
{
  "description": "An awesome organization"
}
```

When we fetch the organization record from MongoDB, the `description` key won't exist, so it will be skipped, and consequently won't be set.

backend/lib/endpoints/organizations.js
mannykary: Thanks for catching this, I actually copied this from `posts.js`. I agree with the suggestion that we just return the object literal or error object directly rather than use `reply.send`. Will fix this.

backend/lib/endpoints/schema/organizations.js
mannykary: Right, that's a good point, I'll remove the required checks.

backend/lib/endpoints/organizations.js
mannykary: I don't have a strong preference, I was just keeping it consistent with the route handlers in the other endpoints.

backend/lib/endpoints/organizations.js
mannykary: I do like the idea of whitelisting fields that can be updated, so I can add that.

mannykary: Another thing I was thinking about - should we set any unset fields coming into a POST create request to `null` so that there is at least a key set on the MongoDB object?

backend/lib/endpoints/schema/organizations.js
mannykary: Fixed in 9ced945

backend/lib/models/schemas/Organization.js
mannykary: Fixed in 943d01b

backend/lib/endpoints/organizations.js
mannykary: Fixed in e17fb45

backend/lib/endpoints/organizations.js
mannykary: @robinv85 @anniegiang @shivanshs9 I decided to check the properties defined on `updateOrganizationSchema` instead of on the `organization` object returned from MongoDB. As I mentioned, the `organization` object may not have all fields set, so patching of any missing fields will not work. I also added error checking so that we can return a 400 bad request error if any extra fields are passed in through the patch request. See 34a650c

Regarding my comment about system fields, if we do decide to add them, we can do that in a separate ticket/PR as it's kind of outside the scope of this ticket.

mannykary:
@robinv85 @anniegiang @MohamedEl1 @shivanshs9 I addressed most of your comments, but there is still one more thing that I would like to do, which is add a reusable function to `utils.js` that sanitizes/validates a patch request. Once I've received feedback for the code [here](https://github.com/manualzuru/FightPandemics.com/pull/105/files#diff-9bc926ee1cc9b662c5556094ecd82af1R80-R93), I can go ahead and abstract that logic out to be more generic and reusable. I did also add logic to return a 400 error if extraneous fields are passed in, but if we'd rather just silently filter them out I'd be fine with that. IMO I think it's better to return 400 so that it's easier for the user of the API to understand that their extraneous field was not accepted.

backend/lib/endpoints/organizations.js
```diff
@@ -0,0 +1,110 @@
...
+      const organization = await Organization.findById(
+        req.params.organizationId,
+      );
+      if (organization === null) {
+        return new httpErrors.NotFound();
+      }
+      const updateProperties = updateOrganizationSchema.body.properties;
+      const errors = [];
+      Object.keys(req.body).forEach((key) => {
+        if (!Object.prototype.hasOwnProperty.call(updateProperties, key)) {
```
robinv85, line 83: Since there's now `Object.prototype.hasOwnProperty()` you probably don't have to use `. call` but you can use this directly on `updateProperties.hasOwnProperty(key)` 

Before it was more like a generic `Object.hasOwnProperty(key)` but now that it's on the prototype it can be used on any instance of an object, also on a literal.

Also... hmm we are kind of doing the same work that the JSON schema has already done when the request is validated because we use the same schema anyway.

mannykary: > Since there's now Object.prototype.hasOwnProperty() you probably don't have to use .call() but you can use this directly on updateProperties.hasOwnProperty(key)

I actually had that initially (`updateProperties.hasOwnProperty(key)` but the linter didn't like it:
```
83:31  error  Do not access Object.prototype method 'hasOwnProperty' from target object  no-prototype-builtins
```
`Object.prototype.hasOwnProperty.call` was the recommended by [ESLint](https://eslint.org/docs/rules/no-prototype-builtins). ¯\_(ツ)_/¯ 


> Also... hmm we are kind of doing the same work that the JSON schema validation has already done when the request is validated because we use the same schema anyway.

Based on my testing, when I had it the way it was before, it just ignored extra properties not defined on the schema, and it did not return a 400 bad request error. Is this what we want? I also noticed that sending a post request with extra properties also ignored them.

robinv85: Hmmm... 
Besides, maybe we should just make sure every schema has the "additionaProperties: false" setting, I forgot about that, in JSON schema you probably have to set that explicitly.
Unless we use something like this https://github.com/greguz/fastify-no-additional-properties which enables it by default for all schema validation in fastify.

mannykary: I think I tried adding that the FluentSchema but it didn't work. But I'll try again tonight to confirm.

mannykary: Okay, finally figured this out. So by default, Fastify removes additional properties by setting the `removeAdditional` ajv config option to `true` (see https://www.fastify.io/docs/latest/Validation-and-Serialization/). So by the time it reaches the route handler, the property is no longer there.

If I override the default configuration to set the `removeAdditional` config option to `false`, and also add `additionalProperties(false)` to the Fluent Schema, I now get the 400 bad request error:

```
{
    "statusCode": 400,
    "error": "Bad Request",
    "message": "body should NOT have additional properties"
}
```

mannykary: @robinv85 check out my latest commits. I reverted my commit where I was checking for additional properties in the route handler, added an overridden ajv config, and also set `additionalProperties` to false on the schemas.

I did try that library you mentioned where it would automatically set `additionalProperties` to false for all schemas, but it didn't work at all. Maybe we can try to roll our own, but I'll leave that for another PR.

robinv85: Hmm ok so I guess that's more explicit to return a bad request error instead of removing the additional properties. 
Only disadvantage is that it's a risk to forget this in the schema. For me this is ok but would be nice if we can set it as a default without risking to forget it in the schema's.
Can you also add this in the readme?

mannykary: Will add to the readme tonight after work. 

Yes, I think it's easier to debug in case a property was misspelled. Better to fail fast than silently.

mannykary: I updated the readme with this caveat, but we should also look into automatically adding this to all our schemas. I did try the library you mentioned but unfortunately I was not able to get it working. We can perhaps address this in another ticket. I will add a ticket for it.

mannykary:
Still waiting re-review from @shivanshs9 as merging is blocked by his request for changes. Not sure what our best practice is, but @robinv85 if you feel this is good to merge, you can go ahead and merge it in.

PR #107, opened 2020-04-12
Recommendation generation based on answers.

mutebironald:
Add all answers based on the CDC questionnaire.

- Try answering the questionaire, the appropriate recommendations will be given at the end.
ref: [#Issue81](https://github.com/manualzuru/FightPandemics.com/issues/81)


client/src/pages/SymptomsCheck.js
```diff
@@ -311,18 +324,322 @@ export const SymptomsCheck = () => {
...
   if (state.age === "under 18") {
     return <Under18 />;
   }
-  if (state.careFacility === "planning to") {
-    return <PlannedCareFacility />;
+
+  //message 1
+  let condition1 =
+    state.age === "18-64" &&
+    (state.traveledLast14Days === "yes" ||
```
robinv85, line 336: this state logic looks very complex and kind of unwieldy... 
I wonder if it would be possible to simplify, one of the first things you could do here is definitely use destructuring to avoid having to access all the nested properties multiple times, for example:
```
const { age, traveledLast14Days, exposureLast14Days ... } = state
```


And then there is still a lot of logic here, I wonder if it's needed to repeat all these answers exactly every time? Maybe this can be improved by using keys to define the answers, which will make the code more DRY and reusable, and can also be used for translations, for example

```
createStep => (question, answers) ...

question = "Have you traveled internationally during the last 2 weeks?"

answers = ["yes", "no"] 
answers = {
  yes: : "Yes, I have",
  no: "No, I haven't",
}
```
either array with strings and use the indexes or use keys and values depending on how we will use it...


We might also be able to use a state machine or something similar to improve this functionality.

robinv85: just thinking out loud, maybe it can be turned into a giant switch case inside a function with returns to simplify, so something like 
```
getCondition(state) => {
  switch (state.answers) {
|    case [Q1[0], Q2[2], Q2[1] ...:
       return {}
      ...
  }
}
```

mutebironald: @robinv85 ➕ on the destructuring. I have done that. In regards to the switch statement mentioned, the questions are different and a switch statement might not be the best for that since there is absence of repitition.

PR #108, opened 2020-04-12
Update README.md

anniegiang:


anniegiang:
@robinv85  antd has more components available than the antd mobile. Some of the components in the antd mobile are specific to mobile screens, which can look slightly off when you view them in larger screen size (accordion for example). I just started using the regular version and some of the components look fine regardless of screen size. Maybe there are performance differences between the two libraries. In general I  guess we should've been more clear about our tech stack from the start, as it seems too late to redo what's done if we want to stick with just one library....

robinv85:
@anniegiang yeah I guess it's great that you added this in the documentation.
We started during hackathon with bootstrap and the outdated desktop designs, then later tried some stuff and were told to focus on mobile. I think ant-mobile was mostly added for the "drawer menu" component and everything else doesn't really matter that much. Except for the icons from Material perhaps. So ideally we now have enough information to make the right decisions about this.
Probably Webpack will also optimize the bundles anyway with tree shaking for stuff we don''t use from libraries.

PR #113, opened 2020-04-13
Add theme info to Frontend guidelines in README 

luxumbra:
I have added quite a bit to the `theme.js` and it would be good if devs knew about the theme so it can start being used. However, we don't have a ThemeProvider handling the theme and are just importing the theme file at the moment, so I'm not sure if it's worth having this info here now or add it later when ThemeProvider is in use. @anniegiang,  @MohamedEl1, @neemiasjnr  would you mind reviewing this please. :pray:


anniegiang:
lgtm thanks for adding this!

PR #124, opened 2020-04-15
Create CreatePost frontend, use theme variables, clean up code

anniegiang:
- mostly styling updates
- styled the frontend for the post creation page 
-  refactored several components related to the Feed to use theme variables 
- created new files to separate styled components from react components

WIP:
- implement the modal that opens and shows options for `sharesWith` and `postExpiration`
- add state functionality for post creation
- use a larger size bullet button for the selects
- probably will clean up more code 

*colors are blander in this GIF than they actually are
create


client/src/components/Feed/FilterBox.js
```diff
@@ -3,35 +3,40 @@ import styled from "styled-components";
...
-  const { filters, modal, handleModal, handleQuit } = feedContext;
+  const { filters, filterModal, handleFilterModal, handleQuit } = feedContext;
   const renderFilterOptions = (filters) => {
     return filters.map((filter, idx) => (
-      <FilterOptionButton
+      <DownArrowButton
         key={idx}
         label={filter.label}
-        handleClick={handleModal(idx)}
+        handleClick={handleFilterModal(idx)}
```
robinv85, line 24: no need for `() => ` here? or it will assign the return value of `handleFilterModal()` as handler?

anniegiang: Ah it might not be clear lookong from here but `handleFilterModal()` returns a function

PR #128, opened 2020-04-15
NeedHelp wizard styled components

luxumbra:
* Integrated the new styled components I created for the step wizard to make the wizards more consistently styled with each other. There are still issues with the wizards, largely in terms of flow, but this deals with the styling in the NeedHelp wizard.
* fixed bug with the onChange function on the email input using the wrong value. `evt` is all we need on these antd-mobile inputs


luxumbra:
I messed up `rebase` so creating a new branch and PR.

robinv85:
@luxumbra yeah you could just forced push again too after you fix your rebase

PR #130, opened 2020-04-15
Create profile

pishili:
Create Profile API, using Fastify

### DIFF
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -5,6 +5,7 @@ WORKDIR '/app'
 COPY package.json package-lock.json ./
 RUN npm install --production
 COPY . .
+RUN npm run build
 RUN echo "{\"version\": \"$GIT_COMMIT\"}" > version.json
 
 FROM builder
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -5,7 +5,6 @@ WORKDIR '/app'
 COPY package.json package-lock.json ./
 RUN npm install --production
 COPY . .
-RUN npm run build
 RUN echo "{\"version\": \"$GIT_COMMIT\"}" > version.json
 
 FROM builder
--- a/backend/lib/endpoints/schema/users.js
+++ b/backend/lib/endpoints/schema/users.js
@@ -7,11 +7,14 @@ const getUserByIdSchema = {
 
 const createProfileSchema = {
   body: S.object()
+    .prop("firstName", S.string().required())
+    .prop("lastName", S.string().required())
     .prop("email", S.string().required())
     .prop("name", S.string().required())
     .prop("country", S.string().required())
     .prop("neighborhood", S.string().required()),
 };
+
 
 module.exports = {
   getUserByIdSchema,
--- a/backend/lib/endpoints/schema/users.js
+++ b/backend/lib/endpoints/schema/users.js
@@ -1,23 +1,22 @@
 const S = require("fluent-schema");
 const { strictSchema } = require("./utils");
 
+const createUserSchema = {
+  body: strictSchema()
+    .prop("name", S.string().required())
+    .prop("country", S.string().required())
+    .prop("country", S.string().required())
+    .prop("neighborhood", S.string())
+    .prop("address", S.string())
+    .prop("wants", S.array().maxItems(3).items(S.string()))
+    .prop("needs", S.array().maxItems(3).items(S.string())),
+};
+
 const getUserByIdSchema = {
   params: strictSchema().prop("userId", S.string().required()),
 };
 
-const createProfileSchema = {
-  body: S.object()
-    .additionalProperties(false)
-    .prop("firstName", S.string().required())
-    .prop("lastName", S.string().required())
-    .prop("email", S.string().required())
-    .prop("name", S.string().required())
-    .prop("country", S.string().required())
-    .prop("neighborhood", S.string().required()),
-};
-
-
 module.exports = {
+  createUserSchema,
   getUserByIdSchema,
-  createProfileSchema,
 };
--- a/backend/lib/endpoints/schema/users.js
+++ b/backend/lib/endpoints/schema/users.js
@@ -5,12 +5,6 @@ const getUserByIdSchema = {
   params: strictSchema().prop("userId", S.string().required()),
 };
 
-        // return new User({
-        //   "email": email,
-        //   "name": name,
-        //   "country": country,
-        //   "neighborhood": neighborhood
-        // }).save();
 const createProfileSchema = {
   body: S.object()
     .prop("email", S.string().required())
--- a/backend/lib/endpoints/schema/users.js
+++ b/backend/lib/endpoints/schema/users.js
@@ -5,6 +5,21 @@ const getUserByIdSchema = {
   params: strictSchema().prop("userId", S.string().required()),
 };
 
+        // return new User({
+        //   "email": email,
+        //   "name": name,
+        //   "country": country,
+        //   "neighborhood": neighborhood
+        // }).save();
+const createProfileSchema = {
+  body: S.object()
+    .prop("email", S.string().required())
+    .prop("name", S.string().required())
+    .prop("country", S.string().required())
+    .prop("neighborhood", S.string().required()),
+};
+
 module.exports = {
   getUserByIdSchema,
+  createProfileSchema,
 };
--- a/backend/lib/endpoints/schema/users.js
+++ b/backend/lib/endpoints/schema/users.js
@@ -7,6 +7,7 @@ const getUserByIdSchema = {
 
 const createProfileSchema = {
   body: S.object()
+    .additionalProperties(false)
     .prop("firstName", S.string().required())
     .prop("lastName", S.string().required())
     .prop("email", S.string().required())
--- a/backend/lib/endpoints/users.js
+++ b/backend/lib/endpoints/users.js
@@ -1,4 +1,6 @@
-const { getUserByIdSchema, createProfileSchema } = require("./schema/users");
+const Auth0 = require("../components/Auth0");
+const { getBearerToken } = require("../utils");
+const { getUserByIdSchema, createUserSchema } = require("./schema/users");
 
 /*
  * /api/users
@@ -37,15 +39,21 @@ async function routes(app) {
   );
 
   app.post(
-    "/:userId",
-    { preValidation: [app.authenticate], schema: createProfileSchema },
-    async (req, reply) => {
-      try {
-        return new User(req.body).save();
-      } catch (err) {
-        return reply.code(err.statusCode).send(err);
+    "/",
+    { preValidation: [app.authenticate], schema: createUserSchema },
+    async (req) => {
+      const user = await Auth0.getUser(getBearerToken(req));
+      const { email_verified: emailVerified } = user;
+      if (!emailVerified) {
+        throw app.httpErrors.forbidden("Email address not verified");
       }
-    });
+      const userData = {
+        ...req.body,
+        _id: req.userId,
+      };
+      return new User(userData).save();
+    },
+  );
 }
 
 module.exports = routes;
--- a/backend/lib/endpoints/users.js
+++ b/backend/lib/endpoints/users.js
@@ -35,6 +35,45 @@ async function routes(app) {
       };
     },
   );
+
+  // app.post("/signup/createProfile", async (req, reply) => {
+  //   try {
+  //     const { email, name, country, neighborhood } = req.body;
+  //     console.log(email)
+
+  //     pool.query(
+  //       `INSERT INTO users (email, name, country, neighborhood)
+  //        VALUES ($1::VARCHAR, $2:: VARCHART, $3::VARCHAR)
+  //        `,
+  //        [email, name, country, neighborhood]
+  //     )
+  //     .then((res) => {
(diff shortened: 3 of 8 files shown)


backend/lib/endpoints/users.js
```diff
@@ -37,6 +37,17 @@ async function routes(app) {
       };
     },
   );
+
+  app.put(
+    "/signup/createProfile",
+    { schema: createProfileSchema },
+    async (req, reply) => {
+      try {
+        return new User(req.body).save();
```
anniegiang, line 46: For security concerns, perhaps we could check what's coming form `req.body` before updating the profile?

backend/lib/endpoints/users.js
```diff
@@ -37,6 +37,17 @@ async function routes(app) {
       };
     },
   );
+
+  app.put(
```
anniegiang, line 41: If we're creating a new profile, wouldn't POST be more restful? Unless are we updating an existing user's profile?

backend/lib/endpoints/users.js
```diff
@@ -37,6 +37,17 @@ async function routes(app) {
...
+    "/signup/createProfile",
+    { schema: createProfileSchema },
```
anniegiang, line 43: Relating to my previous comment about using PUT, if we're updating an existing user profile I think we should add a authentication prevalidation so that only logged in users can update their profile (if that's the intention of this endpoint). We would also need to make sure the currentUser is the owner of the profile if that's the case.

backend/lib/endpoints/schema/users.js
```diff
@@ -4,6 +4,18 @@ const getUserByIdSchema = {
   params: S.object().prop("userId", S.string().required()),
 };
 
+const createProfileSchema = {
+  body: S.object()
+    .prop("firstName", S.string().required())
```
mannykary, line 9: I would add `.additionalProperties(false)` so that we don't allow any additional properties that are not defined in this schema.

backend/lib/endpoints/users.js
mannykary: Agreed, this should be POST since it looks like we are creating a new resource, rather than updating an existing one.

backend/lib/endpoints/users.js
mannykary: Actually, since we're using [FluentSchema](https://github.com/fastify/fluent-schema), and the schema has been defined in `createProfileSchema`, the request body will automatically be validated, so there doesn't need to be any extra checking happening here. We should definitely add `additionalProperties(false)` as I mentioned in another comment so that we don't allow extra properties to be passed in and set on the MongoDB object.

backend/Dockerfile
```diff
@@ -5,6 +5,7 @@ WORKDIR '/app'
 COPY package.json package-lock.json ./
 RUN npm install --production
 COPY . .
+RUN npm run build
```
mannykary, line 8: Was this intentional? I don't see a `build` script in `package.json`.

backend/lib/endpoints/users.js
anniegiang: @mannykary Ah thanks for the clarification! Learned something new +1

pishili: Thank you Annie and Manny for the feedbacks. Added additionalProperties(false) to schema

backend/lib/endpoints/users.js
pishili: I am not sure about this as I had discussion with Robin this morning and he mentioned the following: "POST is creating a new entity so i think we already do that because we have a user
you just have to update user info"

backend/lib/endpoints/schema/users.js
pishili: added

backend/Dockerfile
pishili: had to do this  to make the app works locally

backend/lib/endpoints/users.js
robinv85: yeah actually my thoughts about "create profile" in terms of API and schema's,
- user is created with `/auth/authenitcate`  currently but ideally should be split up in login & register... I thought this would mean "signup" is the same as "create profile" but I guess that's the second step and the user already exists when you enter profile data.
- I thought to update profile we might just store user data in `Auth0` database, so I don't think we need a users collection in mongo ... there is a lot of documentation about this on Auth0 https://auth0.com/docs/users/normalized/auth0/store-user-data so we have to closely look at our use case and the data that we store
- I also thought that we should probably have convenient API routes such as here originally `api/users/current` but I think we can even simplify that and use `/api/me` to GET and PUT or PATCH data
- PATCH would be consistent with updating a post, but I wonder if we need a separate call for the initial profile data and later updating a profile? To me that doesn't make much sense but that''s what we first need to verfiy, to understand the data model and schema requirements (which properties are required to set and which are optional)?

backend/lib/endpoints/users.js
robinv85: yeah i said to use PUT or PATCH because we update user data that already signed up, but see my previous comment it depends a bit on the specific use case how we can design it best

backend/lib/endpoints/users.js
robinv85: yeah definitely needs authentication first and extra check that you can only update your own user data, but see previous comments about which HTTP method and endpoint to use

robinv85:
besides discussing what HTTP verb to use and what route we need to understand what exactly needs to be saved, and where, maybe we can just use Auth0 user db instead of rolling our own.

backend/lib/endpoints/users.js
anniegiang: @robinv85 Ah sounds good. I initially thought this was for user registration so I suggested to use POST. But PUT definitely makes more sense since the user already has a profile.

pishili:
Can you Please review the changes?

backend/lib/endpoints/users.js
robinv85: yeah with PUT you have to send the full user object in the request body so for consistency with `updatePost` in `/posts/:postId` we could choose to always prefer PATCH 
I've used `PUT` to add comments and likes since those are very small sub resources that have separate endpoints ...

backend/lib/endpoints/users.js
```diff
@@ -37,6 +37,17 @@ async function routes(app) {
       };
     },
   );
+
+  app.post(
+    "/signup/createProfile",
```
robinv85, line 42: Another thing with `signup/createProfile` as a route
usually with REST it's a convention to use resources as the endpoints and use HTTP verbs, but avoid naming endpoints as actions such as "create..." etc.

backend/lib/endpoints/users.js
mannykary: I'm getting confused. This is now a `PATCH`, so we're now updating an existing `User` resource based on the passed in `userId`? Shouldn't we be first finding the `User`, and then set only the fields that are being passed in through the `PATCH` request, similar to [this](https://github.com/FightPandemics/FightPandemics/blob/master/backend/lib/endpoints/organizations.js#L74-L85)?

mannykary:
Based on the comments here, and in Slack, it sounds like we're still fleshing out the details and trying to figure out what to store in Auth0 vs in our MongoDB user model. So I'm guessing this is still a WIP. I'll check back once we've gotten a better idea of exactly what we should be doing here.

robinv85:
Following up comments by me and @mannykary and also after discussion with @Naraujo13 I've been thinking some more about this, and it looks like we decided to only store the authentication data in Auth0, and keep all other user data in our own database, this also makes it easier for us to query user data while browsing the app for certain views. 

So I think it will be something like this:

- login & signup on the `/api/auth` endpoints, with `signup` through email also with 2nd field to confirm password, and email confirmation has to be done before step 2 "create profile" can be done. 
We check this by not actually creating a user yet in our database until "create profile" is completed

- create profile will use `POST /api/users/` endpoint and add a user to our database, and is an authenticated route that already has a token from Auth0 

- update profile can use either PUT or PATCH, or even both, that is out of scope here but the advantage is that we have a separate schema, with different required/optional fields. 

I would need to confirm with design / product if this makes sense to also have "email verification:" step first after first sign up with email.

I'm currently working on fixing the frontend auth / login flow now since we changed the backend to get the access token instead of using the Auth0 client library. So we can check for access token and set  ìsAuthenticated`  and also have an additional check if `userExists` or something like that...

backend/lib/endpoints/users.js
robinv85: yeah so everything changed in the mean time, current idea is POST /users :) going to resolve these outdated comments

backend/Dockerfile
robinv85: hmm yeah i don't think this is needed indeed since there's no build step for the backend

mannykary:
@pishili looks like the build is failing due to the `npm run build` command you added to the `Dockerfile`: https://github.com/FightPandemics/FightPandemics/pull/130/checks?check_run_id=620493524 

I think we need to remove that line.

```
Removing intermediate container f279286935e7
 ---> 2a8f09d62085
Step 7/13 : COPY . .
 ---> 0b7b028d3764
Step 8/13 : RUN npm run build
 ---> Running in 0fb12d918842
npm ERR! missing script: build
```

client/Dockerfile
```diff
@@ -3,7 +3,6 @@ WORKDIR '/app'
 COPY package.json package-lock.json ./
 RUN npm install
 COPY . .
-RUN npm run build
```
mannykary, line 6: While `npm run build` is not required for the backend, I believe this one is required for the client; @anniegiang can you confirm?

mannykary:
@pishili looks like there are conflicts that need to be resolved.

client/Dockerfile
anniegiang: Yes we definitely need to build the frontend for it to production ready

backend/lib/endpoints/users.js
```diff
@@ -37,6 +37,17 @@ async function routes(app) {
       };
     },
   );
+
+  app.post(
```
robinv85, line 41: So, I think this should change to `POST /api/users` since creating a new POST should not include the id
usually this id is created in the backend, like in the mongo collection
but in this case we get it from the JWT in the `mongo_id` property that we added in Auth0

I think for now the only thing that needs to be added too is that we make sure the user has a valid email address in Auth0, so we can use probably email from `req.user` I think, but email_verified might need an additional request to `Auth0.getUser()`, alternatively we could also add it to the JWT... can you work on this @pishili ?

robinv85:
Hey @pishili it seems like you've pushed some changes to another branch with the same name on this repository: https://github.com/FightPandemics/FightPandemics/tree/create-profile
This pull request was made from your own fork on that branch: https://github.com/pishili/FightPandemics/tree/create-profile
So that's why you can't see the changes on this pull request. 
Would you like someone to help to get this PR merged?

pishili:
@robinv85 It would be great if I can get help for this,

backend/lib/endpoints/schema/users.js
```diff
@@ -5,6 +5,19 @@ const getUserByIdSchema = {
   params: strictSchema().prop("userId", S.string().required()),
 };
 
+const createProfileSchema = {
```
robinv85, line 8: to stay consistent with the other files, schema's etc. (and API route /users) I think this should probably be named `createUserSchema`

backend/lib/endpoints/schema/users.js
```diff
@@ -5,6 +5,19 @@ const getUserByIdSchema = {
...
+  body: S.object()
+    .additionalProperties(false)
+    .prop("firstName", S.string().required())
+    .prop("lastName", S.string().required())
+    .prop("email", S.string().required())
```
robinv85, line 13: I think we should not allow to pass this `email` prop in the request for now.
The email can be used to create an account with Auth0, and then it will verify the email through their service.
For social media profiles, most of them will have confirmed email addresses, except Twitter and maybe some rare Facebook profiles... but I think we need to find an alternative solution for that, to make sure we only have verified email addresses in our database.
I'm even wondering if it's a good idea to have the email in our own database, because it's already stored in Auth0, maybe we could just encode it in the JWT as well, what do you think about this @Naraujo13 ?
For now we could also do `Auth0.getUser(token);` like in the `/auth/login` and then use the email that we get back from Auth0 and store it in our database.

backend/lib/endpoints/users.js
```diff
@@ -35,6 +38,17 @@ async function routes(app) {
...
+    "/:userId",
```
robinv85, line 43: `userId` should not be in the path because the `userId` is geenrated in Auth0 as `mongo_id` and encoded inside the JWT
so I think this should just be the `/api/users/` route

backend/lib/endpoints/schema/users.js
Naraujo13: I agree that there is no need for `email` prop in the request, getting it through Auth0 seems best. 
About storing or not on our side, I see no issue in storing it

robinv85: @Naraujo13 the reasons that we might prefer not to store it in our db:
- avoid duplication (also need to sync when changing etc, potential bugs)
- more secure to let auth0 handle user data... if our db gets hacked they don’t have the emails

PR #135, opened 2020-04-17
Deployment

barathvk:
This PR adds Github Actions to deploy the app to AWS.

The steps in this process are:

1. Build docker images for the backend, client, and geo-service
2. Push these images to Dockerhub
3. Apply a terraform module found [here](https://github.com/FightPandemics/tf-fargate-task) to deploy the app the AWS ECS Fargate


.github/workflows/workflow.yml
```diff
@@ -0,0 +1,108 @@
...
+                name = "GEO_SERVICE_URL"
+                value = "localhost:5000"
+              },
+              {
+                name = "AUTH_STATE"
+                value = "fight-pandemics"
+              },
+              {
+                name = "AUTH_APP_URL"
+                value = "http://localhost:8000"
```
robinv85, line 79: shouldn't this be `fightingpandemics.com` ? I also wonder if we can deploy with a nginx proxy between to run both client and API on same domain / port and avoid the CORS issues...

.github/workflows/workflow.yml
```diff
@@ -0,0 +1,108 @@
...
+      - uses: kciter/terraform-apply-for-aws-action@v3
+        with:
+          access_key_id: ${{ secrets.REVIEW_AWS_ACCESS_KEY_ID }}
+          secret_access_key: ${{ secrets.REVIEW_AWS_SECRET_ACCESS_KEY }}
+          region: us-east-1
+      - uses: ItsKarma/aws-cli@v1.70.0
+        with:
+          args: ecs update-service --cluster review-cluster --service ${{ steps.get-branch-name.outputs.branch }} --force-new-deployment
+        env:
+          AWS_DEFAULT_REGION: us-east-1
```
robinv85, line 106: should be EU west for GDPR reasons I thought

.github/workflows/workflow.yml
```diff
@@ -0,0 +1,108 @@
...
+            fp_context = "review"
+            subdomain = "${{ steps.get-branch-name.outputs.branch }}"
+            backend_env_variables = [
+              {
+                name = "PORT"
+                value = "8000"
+              },
+              {
+                name = "MONGO_URI"
+                value = "localhost:27017"
```
robinv85, line 67: I think it's best if this includes a database name since we now use `Mongoose.createConnection()` and that can connect to a specific database, or maybe we need to add a separate setting, I'm currently using `127.0.0.1:27017/fightpandemics` locally but I guess we will move this to AWS DocumentDB or Mongo Atlas?

PR #138, opened 2020-04-18
Fix client backend proxy

mannykary:
I also partially fixed the callout to the geo endpoint, but we still need to authenticate to actually get a response back.


robinv85:
ok cool yeah i might have changed backend from POST to GET with query params earlier (makes more sense instead of sending JSON post body)
still working on auth

PR #141, opened 2020-04-19
Auth updates

robinv85:
Updated authentication flow, login & signup
- removed the previous client-side Auth0 library logic
- added some redux actions for init / login / logout
 (getting auth token doesn't mean there's user profile in our db yet, so split up into 2 separate actions / API requests for that)
- updated backend calls for login & signup with schema's, better error handling (for now)
- cleaned up some details

To do after this: 
- finish the "create profile" PR
- finish the "forgot / reset password" flow
- update the "signup" UI with a confirm password, using the real PasswordInput component, with password validation, and also update the backend `/signup` route to use "confirmPassword" field
- client-side validation in login / signup forms also (no toast popup)
- update the redirection logic to make it more robust 
 maybe route components: LoggedIn / LoggedOut / regular Route
 maybe state machine or better routing & redirection mechanism
- even better error handling (the "wrapError" logic in the "backend/components" can be done better and also use the error message from Auth0 for example)
...

Feedback welcome


manualzuru:
@mannykary @barathvk  @Naraujo13  @joaofnds would probably have good feedback also.

client/src/reducers/index.js
```diff
@@ -1,21 +1,33 @@
-import { 
-  SET_USER, 
-} from "../constants/action-types";
+import { AUTH_LOGIN, AUTH_LOGOUT, SET_USER } from "../constants/action-types";
 
 const initialState = {
```
anniegiang, line 3: Should `initialState` also store session errors?

client/src/reducers/index.js
```diff
@@ -1,21 +1,33 @@
...
+        accessToken: action.payload.token,
+        isAuthenticated: true,
+      };
+    case AUTH_LOGOUT:
+      return {
+        ...state,
+        accessToken: null,
+        isAuthenticated: false,
+      };
     case SET_USER:
```
anniegiang, line 23: Why is `SET_USER` a separate action, and not a part of login and logout? Not sure when we would use this separate of auth related actions.

client/src/templates/RouteWithSubRoutes.js
```diff
@@ -32,10 +32,19 @@ export const RouteWithSubRoutes = (route) => {
...
-      render={(props) =>
-        // pass the sub-routes down to keep nesting
-        switchLayout(route.layout, { ...props, ...route.props })
-      }
+      render={(props) => {
+        return !props.isPrivate || props.isAuthenticated ? (
+          // pass the sub-routes down to keep nesting
+          switchLayout(route.layout, { ...props, ...route.props })
+        ) : (
+          <Redirect
```
anniegiang, line 40: Good idea to have a redirect route. We could also generalize routes to be either `protected` or `auth` routes. So for instance, protected routes would be paths that require a loggedin user in order to access, otherwise it will redirect the user to login/signup.

client/src/reducers/index.js
robinv85: yeah that's the thing... it's maybe not really standard, we were only using "SET_USER" before, someone added it, maybe @markthien who added redux and thought this was usually what happens when you login, you get the user info
But in this case, we have a call to `/api/authenticate` (which is currently used for both login & signup but I would like to split them up on the backend too), which returns a token from Auth0
Auth0 can also retrieve the full user object back, but after discussing and looking at some documentation, it seems better to have our own User model in mongodb and store most data there, so the idea is that Auth0 will do just the authentication part (as I mentioned in description of the PR) 
This also means that we don't have a "user" object when we call `authenticate` but just a token. 

We could also add that later on the server side, if we get success response from Auth0 that we also check our mongo and return a user, but maybe it's better to make a separate request to `/api/ucers/current` or similar for that... and then it will need separate redux action and state update... this is why i opened it as a draft PR because it's not really set in stone yet.

client/src/templates/RouteWithSubRoutes.js
robinv85: yeah that's what I was thinking, we could make components for that, and we could maybe set a prop on the definitions in the `routes` file... but currently it's a bit mixed into the whole `layout templates` file which is confusing, not sure if we need that and how we can improve that (maybe wrap multiple HOCs).

client/src/reducers/index.js
robinv85: Don't think so, could be useful but currently didn't see much error handling, I think it should happen mostly in the forms and perhaps show a toaster in some cases when errors happen from network requests.

client/src/reducers/index.js
anniegiang: Yea I'm alright about separating the between using our own User model to store user data and having auth0 just returning auth tokens. It just seems like extra work to not use auth0's full user object. I probably missed the discussion but why aren't we using auth0's user object?

I thought we had a route that gets the current user? Either way I agree it would be handy to have one. I guess that's where we would dispatch  `SET_USER` to  `api/users/current` to set the current user in the frontend.

client/src/templates/RouteWithSubRoutes.js
anniegiang: Yea I've made route components before in personal projects and having those components are pretty handy for routing.

client/src/pages/Login.js
```diff
@@ -64,45 +63,44 @@ const SocialButton = styled(Button).attrs((props) => ({
...
 
   const handleLoginWithEmail = (evt) => {
     evt.preventDefault();
-    if (!validateEmail (email)) {
-      Toast.fail('Invalid email address!', 3);
+    if (!validateEmail(email)) {
+      Toast.fail("Invalid email address!", 3);
       return;
     }
     if (password.length < PASSWORD_MIN_LENGTH) {
```
joaofnds, line 85: What about creating a password validation function? Just like you did with email! Maybe return the error message too

client/src/reducers/index.js
Naraujo13: About using our own User model with Auth0 just as authentication, the main thing would be the integration between the user and all the other data. With a User model of our own at Mongo, we can easily query it alongside other data in a single request (such as bring the author's name and location along with the post query for the feed) or even use the user's info in more complex aggregation pipelines. With the user only inside Auth0, we would need a separate request for that and it would need application code for integrating the info with the data that is inside Mongo.

When I was doing some research on how people usually approach this and found [this discussion](https://community.auth0.com/t/how-to-store-user-information/22148) at the Auth0 forum that seems to be similar to our case (trying to build twitter clone and wanted to know the best approach for storing User data). At the post they present 2 approaches, both keeping a User model at our side, one kept both models in sync and the other used Auth0 only for auth. Long story short, we found that keeping only the authentication at Auth0 side may be the best case to ease the backend pain of keeping both models in sync.

robinv85: > I thought we had a route that gets the current user? Either way I agree it would be handy to have one. I guess that's where we would dispatch `SET_USER` to `api/users/current` to set the current user in the frontend.

Yeah there was already a route made initially but we probably have to clean that up a bit. And yes like @Naraujo13 mentioned, if we do queries for posts and organizations etc. we might need to fetch some user info regularly and then it seems annoying if we have to call the Auth0 API every time, and it can make things like aggregations for mongo queries also impossible in this way, we could also cache user info in mongo but that's adding also going to add more complexity.

So yea if we have separate /login and /register routes, and use `/api/users` for all the CRUD operations for users... only disadvantage is having to make 2 request sometimes e.g. `/auth/login` gets token and then `/iusers/current` gets user info (if we used GraphQL we could do something like getting user at login easier without breaking the REST "rules"). We could also consider to add extra logic in the backend that we check if user exists in our db on /login or /signup.

So if you login, and Auth0 returns that email is not verified => verify email first
If you login and we can't find your user in our db => create profile first

Just need to check now if it's all going to work like this and if we forgot about something.

client/src/templates/RouteWithSubRoutes.js
robinv85: Maybe we should make some more separate tickets to work on these things.

client/src/pages/Login.js
robinv85: yeah `validatePassword` would be a good idea, just have to make sure it has the same validation rules as backend / Auth0.
I consider this out of scope of this PR though.

robinv85: About that, it seems like there was first a `PasswordInput` too but it wasn't working properly so currently it's unused. We should clean that up as well.

client/src/reducers/index.js
robinv85: so one caveat right now is that calling `/signup` and then `authenticate` gets an OAuth token but not the user info from Auth0 so we have to add an additional `getUserInfo` call to make sure email is confirmed after login / signup... so that's a drawback of not using same User db, and i have to check some more of the security issues but that's out of scope here

client/src/templates/RouteWithSubRoutes.js
```diff
@@ -1,41 +1,66 @@
...
+  return (
+    <Route
+      path={path}
+      render={({ layout, location, ...rest }) => {
+        const Layout = layout === "logo" ? LogoLayout : NavigationLayout;
+        let redirect;
+        // todo: cover all different cases, ensure all the props in the routes config are set correctly
+        //  maybe use LoggedIn, NotLoggedIn, Route components
+        //  or maybe even a state machine to simplify this logic
+        if (loggedInOnly && !isAuthenticated) {
```
anniegiang, line 28: what's the difference between `loggedInOnly` and `isAuthenticated`

robinv85: yeah I might need to rename things here, 
`loggedInOnly` and `notLoggedInOnly` are props that can be set on a route (which could be also using LoggedIn and NotLoggedIn route components as I've mentioned)
While `isAuthenticated` is the state of the logged in user, this means we did login or signup and got a token.

robinv85: Also, as you might've noticed, I had to clean this up a bit to merge the layout switching and the redirection logic, which uses props from the `routes` config, as well as from the `RouteWithSubRoutes` ocmponent with `render()` method and also wrapped in Redux connect().... this is getting quite complex.
What the next step can be here is to add redirection to "create profile" what @pishili is working on, to make sure if profile is not created it redirects to that page first.
But as you can see that can become very complex and unwieldy logic here so we need to find a way to make it better (maybe state machine...)

robinv85: currently looking if we could use https://statecharts.github.io/ to simplify this whole redirection stuff

PR #142, opened 2020-04-19
Remove unused variables 

anniegiang:
Removed unused variables noted by the console warnings.

**Before:** 
Screen Shot 2020-04-18 at 7 02 33 PM
**After:**
Screen Shot 2020-04-19 at 10 59 09 AM

### DIFF
--- a/client/src/pages/OfferHelp.js
+++ b/client/src/pages/OfferHelp.js
@@ -1,10 +1,8 @@
 import React, { useState } from "react";
 import { withRouter, Link } from "react-router-dom";
-import styled from "styled-components";
 
 import { asyncGetGeoLocation } from "../utils/geolocation";
 import {
-  AnswerButton,
   StyledWizard,
   WizardContainer,
   WizardStep,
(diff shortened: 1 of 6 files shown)


client/src/pages/OfferHelp.js
```diff
@@ -50,20 +45,20 @@ const Step1 = (props) => {
     updateState({ ...state, answers: updatedAnswers });
     props.update("helpTypeOffered", checkedAnswers);
   };
-  const toggleNone = () => {
-    const newNone = !none;
-    updateState({ ...state, none: newNone });
-    props.update("helpTypeOffered", newNone ? [] : getCheckedAnswers(answers));
-  };
+  // const toggleNone = () => {
```
anniegiang, line 48: I commented this out instead of removing it all together

robinv85: hmm not sure what is going on here, I remember implementing this in one of the 2 flows for the step wizards, might have been copied here and not needed or i might already have broken the functionality in some places...

PR #151, opened 2020-04-21
Feed Frontend: show airtable data and truncate text

anniegiang:
- Replaced dummy data from `fakePosts.js` with the fetched scraped data from Airtable to render them as posts in /feed
- Installed `react-read-more-read-less` to truncate post description
- Commented out code related to commenting and liking since the data coming from Airtable does contain data to support these functionalities

WIP
- Will need a way to add HTML to post descriptions:
    - You can see in the GIF that some descriptions don't have any line breaks (especially for the really long descriptions)
     - Turn url links in the descriptions into `<a href="" />` tags

*colors are blander than they actually are
post


anniegiang:
Some updates: 
I talked to @manualzuru and we're not going to display Airtable data in the Feed page anymore. The data doesn't have anything related to social interactions like comments and likes, which is what we want. 

So the plan now is to populate Mongo with dummy posts created by one user and just display those for now. In the future once Auth is solid I'll populate Mongo with dummy data but with a user that's authenticated and in out DB.

PR #152, opened 2020-04-21
implement forgot password feature

markthien:
implement forgot password feature on the same page as Login


anniegiang:
Should we close this PR? I've reached out to Mark but haven't gotten a response. It's been open for a while and many changes have been made related to this PR.

robinv85:
Yeah not sure, @markthien do you have time to look at this? 
If not, and if someone else wants to work on this then maybe they can look at the branch and see if they can use the code or would like to start over.

anniegiang:
@markthien Hi Mark, do you have any updates on this?

robinv85:
@markthien my earlier comment about redux and not using `REQUEST, FAILED, SUCCESS` actions is now not valid anymore because we have some updated requirements for the validation and error handling. This means that we will probably have to use that pattern for most of the API requests, do you think you can still update this PR and get it merged?

markthien:
> @markthien my earlier comment about redux and not using `REQUEST, FAILED, SUCCESS` actions is now not valid anymore because we have some updated requirements for the validation and error handling. This means that we will probably have to use that pattern for most of the API requests, do you think you can still update this PR and get it merged?

Yes will do

client/src/reducers/index.js
```diff
@@ -33,6 +38,16 @@ function rootReducer(state = initialState, action) {
         ...state,
         user: action.payload,
       };
+    case SET_ERROR:
```
robinv85, line 41: I'm wondering if there's a way to prevent having to store all the errors in the global state if we don't needed.
This is kind of "the Redux way" but I'm also working on the login / signup form error handling at the moment... and I'm thinking it's not very useful to keep this in the global state if we don't need to persist the errors.
Also, it could lead to bugs and sharing errors from different components here with the same state. At least we could maybe split up the state for multiple components... some examples are listed here in this article too https://dispatch.moonfarmer.com/redux-status-and-error-methodology-ec5c1f1634b7

robinv85:
Hey @markthien I've looked at the login / signup pages for validation and error handling and updated the form to use `react-hook-form` and prefer using the `useReducer` over Redux, only kept the logged in session and user in the redux store for now, see #415 ... do you think you could also work on this to refactor later or maybe we should start over from the updated form?

anniegiang:
Any updates on this?

joshmorel:
Closing in favor of #649 . We do need to start from scratch

PR #156, opened 2020-04-23
Footnote for Mobile and Desktop #149

### LINKED ISSUE #149: Create Footnote for Mobile and Desktop
- [ ] Create Footnote for Mobile and Desktop

- [ ] All the UI's need to include the Footnote

Requirements:
https://www.notion.so/fightpandemics/Footnote-24895e52864547a5be0c365fcccc3595

PriyalPatney:
Fixes #149


client/src/index.js
```diff
@@ -12,14 +12,65 @@ import "./index.scss";
...
+const onRedirectCallback = (appState) => {
+  history.push(
+    appState && appState.targetUrl
+      ? appState.targetUrl
+      : window.location.pathname,
+  );
+};
+
+const Footer = () => (
+  <footer style={footerStyle}>
```
anniegiang, line 47: Let's use styled-components for styling as we would like to stay away from inline-styling. It should be a quick fix with the basic syntax like this: 
```
const StyledFooter = styled.footer`
// paste the styles from footerStyle
`;
```
and then to use `StyledFooter` you can replace line 47 with `<StyledFooter></StyledFooter>`

client/src/index.js
```diff
@@ -12,14 +12,65 @@ import "./index.scss";
...
 import * as serviceWorker from "./serviceWorker";
 import rootReducer from "./reducers";
-
+import zIndex from "@material-ui/core/styles/zIndex";
 const store = createStore(rootReducer, applyMiddleware(thunk));
 
-ReactDOM.render(
+const footerStyle = {
+  backgroundColor: "#fff",
+  fontSize: "10px",
```
anniegiang, line 20: Let's use `rem` values instead of `px`. Currently `1rem = 10px`.

anniegiang: And for styling related to typography and colors, please use the values from `theme.js`.

client/src/index.js
```diff
@@ -12,14 +12,65 @@ import "./index.scss";
...
+  );
+};
+
+const Footer = () => (
+  <footer style={footerStyle}>
+    <span style={{ fontWeight: "bold" }}>
+      Copyright 2020 Fight Pandemic. All rights reserved.
+    </span>
+    <br />
+    <a href="http://www.fightpandemics.com/terms-conditions" style={links}>Terms &amp; conditions</a>|
```
anniegiang, line 52: I think we will have pages for terms-conditions, privacy-policy, and cookies-policies so we'll probably use  relative paths for these links

client/src/index.js
```diff
@@ -12,14 +12,65 @@ import "./index.scss";
...
+      domain={config.domain}
+      client_id={config.clientId}
+      redirect_uri={window.location.origin}
+      onRedirectCallback={onRedirectCallback}
+    >
+      <App />
+    </Auth0Provider>
+  </Provider>
 );
+const WrapperWithFooter = withFooter(Wrapper);
```
anniegiang, line 73: Interesting approach to rendering the Footer. Is there a difference between wrapping the entire App with Footer and just rendering Footer within the `App.js` component?

client/src/index.js
```diff
@@ -12,14 +12,65 @@ import "./index.scss";
...
+    appState && appState.targetUrl
+      ? appState.targetUrl
+      : window.location.pathname,
+  );
+};
+
+const Footer = () => (
+  <footer style={footerStyle}>
+    <span style={{ fontWeight: "bold" }}>
+      Copyright 2020 Fight Pandemic. All rights reserved.
```
anniegiang, line 49: Looks like there is a typo in the designs - there should be an 's' after 'Fight Pandemic'

client/src/index.js
PriyalPatney: I tried this approach first, but it was causing some issue with HOC functions.  I'll try to rectify it👍

client/src/index.js
PriyalPatney: Yea it will be an efficient approach.

client/src/index.js
PriyalPatney: euh yeah I will update it inside features requirement too. I appretiate your keen review.😀

client/src/index.js
```diff
@@ -12,14 +12,65 @@ import "./index.scss";
 import App from "./App";
 import * as serviceWorker from "./serviceWorker";
 import rootReducer from "./reducers";
-
+import zIndex from "@material-ui/core/styles/zIndex";
```
robinv85, line 15: even though we currently don't have a lint rule for this in the frontend, for the imports I think it's best to group them, first all the external dependencies and then the internal own to our own code... makes it a bit more clear, I would even prefer if they are done alphabetically as well so it's easy to find something.

client/src/index.js
robinv85: Yeah please prefer styled components unless you have to pass styles through props...
Also it seems like you might not need a HOC for this? (I assume you mean something like  `withFooter` but if this is only for the full App then we can just put a component in there?)

client/src/index.js
robinv85: will these pages be  static and outside of the react app?

client/src/index.js
```diff
@@ -12,14 +12,65 @@ import "./index.scss";
...
+const withFooter = (WrappedComponent) => () => [
+  <WrappedComponent key="1" />,
+  <Footer key="2" />,
+];
+const Wrapper = () => (
   <Provider store={store}>
-    <App />
-  </Provider>,
-  document.getElementById("root"),
+    <Auth0Provider
```
robinv85, line 63: this was removed, should update from master?

PR #157, opened 2020-04-23
Footnote

nadinejuraschek:
Mobile Footnote styled and implemented


nadinejuraschek:
Sorry for the weird milestone and labels, my toddler was typing...

client/src/components/Footnote.js
```diff
@@ -0,0 +1,68 @@
+import React from "react";
+import { Link } from "react-router-dom";
+
+import styled from "styled-components";
+
+const StyledFooter = styled.div`
+  position: absolute;
+  width: 376px;
```
anniegiang, line 8: Let's use `rem`  values instead of `px`. For example, right now we have `1rem = 10px`. So for line 8, it would be `37.6rem` because 376px/10px = 1

client/src/components/Footnote.js
```diff
@@ -0,0 +1,68 @@
...
+  height: 71px;
+  left: 0;
+  bottom: 0;
+  top: 759px;
+
+  background: #f3f4fe;
```
anniegiang, line 14: For styles related to colors and typography like fonts, please use the values from `theme.js`

client/src/components/Footnote.js
```diff
@@ -0,0 +1,68 @@
...
+
+export default () => {
+  const currentYear = new Date().getFullYear();
+  return (
+    <StyledFooter>
+      <Copyright>
+        Copyright {currentYear} FightPandemics. All rights reserved.
+      </Copyright>
+      <Policies>
+        <FooterLink href="https://fightpandemics.com/terms-conditions">
```
anniegiang, line 54: Let's use relative links instead of absolute links for the footer links. So for example on line 54, we can say `href="/terms-conditions"`

anniegiang:
@robinv85 Some updates-  PR #156 is the nearly the same as this PR but Priyal closed it to keep it simple and also because this PR fixes most of the requested changes that were in #156.

nadinejuraschek:
I noticed something off about theme.js: In colors.js there is a DARK_GRAY that I wanted to use, but in theme.js darkGray is assigned to a different hex value.

client/src/components/Footnote.js
```diff
@@ -0,0 +1,65 @@
+import React from "react";
+import { Link } from "react-router-dom";
+import { theme } from "../constants/theme";
+
+import styled from "styled-components";
+
+const { colors } = theme;
+
+const StyledFooter = styled.div`
+  position: absolute;
```
anniegiang, line 10: Not sure if using position absolute is appropriate for this actually. Using that along with top/bottom` and `left/right` values might not make the footnote responsive, since they are fixed values right now. You could probably see what I mean by changing the screen size and see how the links move in response. I think we could just give the footer-links to have display inline-block so that they are all on same line, and then just mess with the margin/padding to get the right amount of spacing. Let's also remove the width and height values for each of the different links, since those values with be determined by the content. Let me know if you have any questions with that.

client/src/components/Footnote.js
```diff
@@ -0,0 +1,65 @@
...
+  height: 2.1rem;
+  left: 2.8rem;
+  top: 1.8rem;
+
+  font-weight: 500;
+  color: ${colors.darkGray};
+`;
+
+const Policies = styled.div`
+  position: absolute;
```
anniegiang, line 36: Don't think we need position absolute here -  it would naturally be under the `Copyright` component and by default have display block and take up the entire width

client/src/components/Footnote.js
```diff
@@ -0,0 +1,65 @@
+import React from "react";
+import { Link } from "react-router-dom";
+import { theme } from "../constants/theme";
+
+import styled from "styled-components";
+
+const { colors } = theme;
+
+const StyledFooter = styled.div`
```
anniegiang, line 9: Let's use `footer` instead of `div` so that it's semantic

anniegiang:
@nadinejuraschek Regarding the DARK_GRAY concern, yea please fix that if it's not right

PR #159, opened 2020-04-24
Tweaks to the Initial Developer Experience

vsupalov:

This PR is meant to fix a few issues with the initial developer onboarding experience, so people have an easier and more intuitive time getting the project running for the first time.

* The geo service docker image won't require dependency reinstalls on every code change.
* Similar example .env files for backend and client.
* A fix to stop the client dev server from taking over the complete docker-compose output.
* Slight modifications to the "getting started" section of the README.
* A PR template to provide a bit of guidance towards more unified PRs.


client/Dockerfile.dev
```diff
@@ -2,4 +2,5 @@ FROM node:12-alpine
 WORKDIR '/app'
 COPY . .
 RUN npm install
-CMD ["npm", "start"]
+#CMD ["npm", "start"]
+CMD ["npx", "concurrently", "npm:start"]
```
robinv85, line 6: what's the benefit of running this with `concurrently`? Might as well delete the original then, too...
I remember reading for older versions to deploy Node.js in Docker by default using `npm start` wouldn't correctly handle the exit codes in production or something like that https://selwyn.cc/writings/docker-node-exit-codes
but this would probably only be relevant for the backend node.js app

vsupalov: There's a comment missing to clarify it, my bad.

It's an ugly fix for around the fact that you can't stop the react app from clearing the screen, as discussed [here](https://github.com/facebook/create-react-app/issues/2495#issuecomment-488505385). The original line resulted in all other docker-compose outputs being cleared away. Running it using concurrently is one of the proposed workarounds. Not a fan of either, but this seemed to provide a more coherent experience overall.

PR #160, opened 2020-04-24
Auth handlers

### LINKED ISSUE #115: Update requests to update/delete posts to authenticate user - backend
TODO:
* Ensure the JWT is set as authorization header (Bearer token) with user id to secure the API endpoints
* Make sure that a user can only delete or update their own posts
* Make sure that the user is authorized to comment/like/delete likes or comments on a post (not sure what the rules are here around post visibility)

Possible approach:
* Use the JWT to get the current user ID, and confirm that userId = post.authorID
* Check the TODOs in backend/lib/endpoints/posts.js

Questions for engineering:
* How are we representing different levels of post visibility?

robinv85:
Closes #115 
Added shorthand for `userId` from the decoded JWT on the request object for authenticated endpoints.
Checks for userId in the api handlers to get the right author of posts, comments, likes etc. and also checks if user is authorized to update or delete something.

For integration with Auth0 : 
- generate mongo ObjectId and use it as user_id to create Auth0 user (username / password)
- get the Auth0 user_id from the decoded JWT token and convert to an object id (without the prefix)

this way we can use both, might need some more work to link Auth0 with our own users, also need to check for the social media profiles.


backend/lib/endpoints/posts.js
```diff
@@ -215,8 +221,9 @@ async function routes(app) {
     { preValidation: [app.authenticate], schema: likeUnlikePostSchema },
     async (req) => {
       const { postId, userId } = req.params;
-      // todo: get user id from JWT
-      //  check if userId is the same as param, and if authorized to like (based on visibility)
+      if (!req.userId.equals(userId)) {
```
mannykary, line 224: I wonder if we could make this DRY by moving this code to a prevalidation hook? https://www.fastify.io/docs/latest/Hooks/

robinv85: yeah I was also thinking about extracting some of the repeating code blocks here, wasn't sure yet what the best way would be, we started with that `backend/components` folder which seemed to be intended for reusable blocks but those functions are more difficult to handle erorrs (if we have to `throw new Error(...)` or `throw new HttpError(...)` 
so I was kind of thinking about a more systematic approach, perhaps using more hooks will be the best way to do it

robinv85: Also, I think it would be better to use `preHandler` so the order would be :
- check auth first (decode JWT + add user on req object)
- schema validation is done to ensure the request has correct format
- any additional hooks could do things like check if a user is only updating his / her own resource. But I think these patterns are not very established at the moment, in a RESTful way we could use PUT with own user Id in the route param or only use the `req.userId` and always return 200 response which is maybe better for idempotent REST request (PUT, PATCH or DELETE) so the actual auth check can differ for each operation

backend/lib/endpoints/posts.js
```diff
@@ -108,35 +109,48 @@ async function routes(app) {
...
+        return new httpErrors.NotFound();
+      }
+      if (!post.authorId.equals(userId)) {
+        return new httpErrors.Forbidden();
       }
-      Object.keys(req.body).forEach((key) => {
-        if (post[key] && post[key] !== req.body[key]) {
-          post[key] = req.body[key];
+      Object.keys(body).forEach((key) => {
+        if (post[key] && post[key] !== body[key]) {
```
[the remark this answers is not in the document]
robinv85: yeah I didn't change this logic but I don't think this is a good approach
it's a naive way to think that every key in the API request will be exactly the same as the mongo model, we definitely need to make sure the schema validation is very strict and also handle things like nullable etc.

PR #166, opened 2020-04-25
Continuous deployment of review branches and staging

mannykary:
### PR Checklist:

* [X] Have you checked out the guidelines in our [Contributing](../CONTRIBUTING.md) document?

This adds continuous deployment of review branches, with the `feature/` prefix, to the review environment, as well as the master branch to the staging environment. Every time a commit is pushed to a feature review branch, a build will be kicked off and deployed to AWS, where it will be accessible at `<branch_name>.fightpandemics.xyz`, where `<branch_name>` is normalized to replace `/` with `-`. For example, `feature/deployment` will be accessible at http://feature-deployment.fightpandemics.xyz. When the pull request is closed or merged, the review branch app will be torn down in AWS. In addition, a staging build will be kicked off and deployed to the staging environment at AWS, where it will be accessible at http://staging.fightpandemics.work.

* Factored out main.tf Terraform so that it can be reused across workflows
* Sanitize branch name so that any `/` in the branch name will be replaced with `-`, making it a valid subdomain
* Added staging Github workflow 
* Added review branch delete Github workflow
* Updated readme with deployment details
* Only run builds for branches with `feature/` prefix, so that it is not run for branches in which there are no tangible changes to the app itself


README.md
```diff
@@ -87,3 +87,16 @@ additional properties not defined in the schema are passed in through the reques
 * For new pages, create a new route in `client/src/routes.js` and import a page component from `client/src/pages`
 * Refrain from making the Redux state too big as it will affect speed performance. React’s Context API is ONLY helpful for avoiding nested prop threading so stick with Redux where Context fails.
 
+## Deployment
+
+### Review branches
+
+Every time you push code up to the repository, a build based off of your feature review branch will be deployed to AWS. Note that in order for a build to run, the branch name must have the `feature/` prefix. You can view the build logs in [GitHub Actions](https://github.com/FightPandemics/FightPandemics/actions). After the build successfully completes, you can access your built app at `<branch_name>.fightpandemics.xyz`, where `<branch_name>` is the *sanitized* branch name of your feature branch (i.e. if the branch name has a `/` in it, it will be converted to a `-`, e.g. `feature/hot-potato` will be converted to `feature-hot-potato`). Note that it may take a few minutes for the app to be accessible, or for changes to be reflected, since it takes time for AWS to spin up the Docker containers.
```
robinv85, line 94: ok good to know, too bad we can't have other branch names deployed like  `/bugfix` like this though
also wonder if it's automatically removed when the branch is deployed?

mannykary: Actually, I can add `bugfix/`, as well as any other prefix we may want to trigger builds.

And yes, the review build would be torn down after the branch is merged to master, or if the branch PR is closed without merging.

.github/workflows/review-workflow.yml
```diff
@@ -1,6 +1,14 @@
-name: fight-pandemics
+name: review
 
-on: push
+on:
+  push:
+    branches:
+      - feature/**
```
mannykary, line 6: @robinv85 this is where we filter for branches on which we want the review build workflow to run. In addition to `bugfix/`, what other prefixes do you think we should filter for?

mannykary:
I'm going to go ahead and merge this in so that we can start taking advantage of continuous deployment. If we do decide to filter for other branch prefixes such as `bugfix/`, we can open another PR to add that.

Issue #184, opened 2020-04-26
Add filters to /posts endpoint 


anniegiang:
In the `/feed` page, a user can filter posts based on the following filters:
1. Location (will use Google Maps API)
   - Related to Issue #64 
2. From Whom
3. Type
4. Need or give help

After a user selects filters, the Feed will make a `GET` request  to `/api/posts` that: 
 - [ ] Accepts queries of filter values.
 - [ ] Queries into database based on filter values.
 - [ ] Return to the client filtered posts that are paginated (10 posts) and lazy loaded on the client.

More reference:
https://www.notion.so/fightpandemics/Feed-653658508c4649869787c48f3e4eee13

Contact @anniegiang for questions about how the frontend of the Feed was built.

robinv85:
So I guess title of the endpoint can change to "add filter parameters to posts API" or something like that instead of creating a new endpoint, I don't think that's necessary.

anniegiang:
@robinv85 Yea I agree, will update the issue.

PR #186, opened 2020-04-27
Filter posts based on filters

anniegiang:
- Started on a basic route `/api/posts/filters` that accepts filters as query params and queries into the db based on those filters

- Simplified some route schemas to not accept fields that are not absolutely needed (yet)

- Renamed the following fields in the Post model and route schema to match wording in `filterOptions.js`
  - `type` -> `needs`
  - `looking`/`lookingFor` -> `helpType`

Next Steps:
- I wonder if `GET api/posts/` should actually be handling filters instead of creating a new route like in this PR ? 

- I'll probably need to figure out a more efficient way of aggregating... would love any insights on the current implementation!

- Add backend pagination for the lazy loading feature of 10 posts, that are also sorted based on nearest location to the user

- Need to get clarification on the Post model. Right now when a user creates a post, the server gets data for `title`, `description`, `needs`, and `helpType`. Not sure when we get data from the user for the other fields like `fromWhom` and `tags`

Linked to issue #184

### DIFF
--- a/backend/lib/endpoints/posts.js
+++ b/backend/lib/endpoints/posts.js
@@ -107,7 +107,11 @@ async function routes(app) {
 
   app.get("/filters", { schema: getPostByFiltersSchema }, async (req) => {
     const { helpType, needs } = req.query;
-    
+    const aggregates = [];
+
+    aggregates.push({ $unwind: "$needs" });
+
+    return Post.aggregate(aggregates);
   });
 
   app.delete(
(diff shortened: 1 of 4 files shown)


robinv85:
Yeah usually a RESTful filter is more something like
`/api/posts?authorId=abc123&status=published&language=en` but we can design the API based on our requirements
Sometimes you even see JSON stringified inside of these query params if we need more advanced filtering capabilities (for example location could be more complex, given coordinates and within given distance or something like that). 
Maybe we can look at some examples of REST API's / standards for reference, 
For example the JSON:API standard specifies a "filter" query param can be used and implemented however you wish https://jsonapi.org/format/#fetching-filtering
https://github.com/typicode/json-server#filter has some great stuff built-in by default
`?q=` could have a query / search

PR #189, opened 2020-04-28
Terms & Conditions, Privacy Policy, and Cookie Policy added

### LINKED ISSUE #150: Policy Page updates
These are the 3 policy pages:
www.fightpandemics.com/terms-conditions
www.fightpandemics.com/privacy-policy
www.fightpandemics.com/cookies-policy

Please work with Edwin Mata to add content to the page.

nadinejuraschek:
closes #150


client/src/pages/CookiesPolicy.js
```diff
@@ -0,0 +1,197 @@
...
+
+export const CookiesPolicy = () => {
+  return (
+    <PolicyContainer>
+      <TextContainer>
+        <h2 className="text-primary display-6">Cookies Policy</h2>
+        <ListNoIndent>
+          <li>
+            WHAT ARE COOKIES?
+            <ol type="I">
```
anniegiang, line 39: what's the type attribute for?

client/src/pages/CookiesPolicy.js
```diff
@@ -0,0 +1,197 @@
+import React from "react";
+import styled from "styled-components";
+
+import { theme } from "../constants/theme";
+const { colors } = theme;
+
+const PolicyContainer = styled.div`
```
anniegiang, line 7: I notice some of the styles are similar across pages. Perhaps we could create a file that specifically exports styles used across the three components, and then import them in components where needed. What do you think?

manualzuru: I like this idea!

anniegiang: Yea I have no strong preference since the styling for these pages are rather minimal. But it's definitely good practice to reuse code wherever possible

client/src/pages/CookiesPolicy.js
nadinejuraschek: It changes the default ordered list (1., 2., 3.) to either lowercase letter (type="a"), uppercase roman numbers (type="I") or lowercase roman numbers (type="i"). I used the same list types as @EdwinMata in the document he sent me.

anniegiang: Ah didn't know about that! Lgtm then

anniegiang:
@nadinejuraschek Thanks for making the changes, lgtm!

PR #191, opened 2020-04-29
Update Post data model

Naraujo13:
Update Post data model to [v2](https://github.com/FightPandemics/DataModel), while adjusting post endpoints.

- Added v2 models at schemas/v2
- Added new Post model
- Updated all post endpoints to use the new model
- Add expiration date to posts
- Add visibility range to posts
- Add sorting by distance to posts

Endpoints:
- GET /posts
- POST /posts
- GET /posts/:postId
- PATCH /posts/:postId
- DELETE /posts/:postId
- PUT /posts/:postId/:userId (like)
- DELETE /posts/:postId/:userId (unlike)

Missing getting the authenticated user, pagination and filters at GET requests since those are out of scope.


backend/package.json
```diff
@@ -31,6 +31,7 @@
     "axios": "^0.19.2",
     "bcryptjs": "^2.4.3",
     "cors": "^2.8.5",
+    "datejs": "1.0.0-rc3",
```
joaofnds, line 34: What about using a more "popular" date library? I think both [date-fns](https://www.npmjs.com/package/date-fns) and [moment](https://www.npmjs.com/package/moment) are great alternatives!

Datejs haven't seen a release in the past 5 years

backend/lib/models/schemas/v2/post.js
```diff
@@ -1,11 +1,12 @@
 // -- Imports
 const { Schema, model, ObjectId } = require("mongoose");
-const { schema: authorSchema } = require("./author");
+// const { schema: authorSchema } = require("./v2/author");
```
joaofnds, line 3: Can we remove this commented code?

backend/lib/models/schemas/v2/post.js
```diff
@@ -1,11 +1,12 @@
...
 
 // -- Schema
 const postSchema = new Schema(
   {
-    author: authorSchema,
+    // author: authorSchema,
```
joaofnds, line 8: Can we remove this commented code?

backend/lib/endpoints/schema/posts.js
```diff
@@ -1,48 +1,84 @@
...
+  "organizationUser",
+));
+
+const EXPIRATION_OPTIONS = ["day", "week", "month", "forever"];
+const POST_OBJECTIVES = postSchema.tree.objective.enum;
+const POST_TYPES = postSchema.tree.types.enum;
+const USER_TYPES = [].concat(
+  userSchema.tree.type.enum,
+  orgSchema.tree.type.enum,
+);
```
joaofnds, line 18: Is there any need to use `.concat` here? Maybe change it to use the array literal
```js
const USER_TYPES = [userSchema.tree.type.enum,  orgSchema.tree.type.enum];
```

backend/lib/endpoints/posts.js
```diff
@@ -21,14 +22,30 @@ async function routes(app) {
...
 
   app.get("/", { schema: getPostsSchema }, async (req) => {
-    // todo: add limit, skip, visibility filter, needs, tags ...
-    const { authorId } = req.params;
-    const aggregates = authorId ? [{ $match: { authorId } }] : [];
-    aggregates.push(
+    // TODO: get userId from jwt
+    // userId = ?
+    // user = User.findById(userId);
+    var user = await User.findById("5ea6900c0e0419d4cb123611");
```
joaofnds, line 33: I think this is "test/debug code" that you forgot to remove

backend/lib/endpoints/posts.js
```diff
@@ -40,42 +57,88 @@ async function routes(app) {
...
+    },
     async (req, reply) => {
+      // TODO: add pagination
       const { postId } = req.params;
       const post = await Post.findById(postId);
       if (post === null) {
         return reply.send(new httpErrors.NotFound());
       }
-      post.comments = await Comment.aggregate([
+      var commentQuery = await Comment.aggregate([
```
joaofnds, line 141: Shouldn't this bet a `let`?

backend/lib/endpoints/posts.js
```diff
@@ -40,42 +57,88 @@ async function routes(app) {
...
-      return new Post(req.body).save();
+    {
+      preValidation: [app.authenticate],
+      schema: createPostSchema,
+    },
+    async (req, reply) => {
+      // TODO: get userId from jwt
+      // userId = ?
+      // user = User.findById(userId);
+      user = await User.findById("5ea6900c0e0419d4cb123611");
```
joaofnds, line 86: I think this is "test/debug code" that you forgot to remove

backend/lib/endpoints/posts.js
```diff
@@ -40,42 +57,88 @@ async function routes(app) {
...
+        reply.code(201);
+        return post;
+      } else {
+        return reply.send(new httpErrors.BadRequest());
+      }
     },
   );
 
-  // /posts/postId
+  // // /posts/postId
```
joaofnds, line 126: I think this was added by mistake

backend/lib/endpoints/posts.js
```diff
@@ -127,19 +216,102 @@ async function routes(app) {
...
+    {
+      preValidation: [app.authenticate],
+      schema: updatePostSchema,
+    },
     async (req, reply) => {
-      // todo: make sure user can only update their own post
+      // TODO: get userId from jwt
+      // userId = ?
+      // user = User.findById(userId);
+      var user = await User.findById("5ea6900c0e0419d4cb123611");
```
joaofnds, line 227: I think this is "test/debug code" that you forgot to remove

backend/lib/endpoints/posts.js
```diff
@@ -93,27 +156,53 @@ async function routes(app) {
...
+      schema: deletePostSchema,
+    },
     async (req) => {
       const { postId } = req.params;
-      // todo: make sure user can only delete their own posts
-      const deletedPost = await Post.findByIdAndRemove(postId);
+      // TODO: get userId from jwt
+      // userId = ?
+      // user = User.findById(userId);
+      var user = await User.findById("5ea6900c0e0419d4cb123611");
```
joaofnds, line 194: I think this is "test/debug code" that you forgot to remove

robinv85:
@Naraujo13 I've merged the auth checks so maybe you can check what's changed and rebase/merge

backend/lib/endpoints/posts.js
```diff
@@ -127,19 +216,102 @@ async function routes(app) {
...
-      if (post === null) {
-        return reply.send(new httpErrors.NotFound());
-      }
-      Object.keys(req.body).forEach((key) => {
-        if (post[key] && post[key] !== req.body[key]) {
-          post[key] = req.body[key];
+      if (!post) return new httpErrors.NotFound();
+      else if (post.author.authorId != user.id)
+        return new httpErrors.Forbidden();
+      var body = req.body;
```
robinv85, line 234: so I guess here, you could `const { body } = req;`

backend/lib/endpoints/posts.js
```diff
@@ -127,19 +216,102 @@ async function routes(app) {
...
+      if (user === null) return new httpErrors.Unauthorized();
+
       const post = await Post.findById(req.params.postId);
-      if (post === null) {
-        return reply.send(new httpErrors.NotFound());
-      }
-      Object.keys(req.body).forEach((key) => {
-        if (post[key] && post[key] !== req.body[key]) {
-          post[key] = req.body[key];
+      if (!post) return new httpErrors.NotFound();
```
robinv85, line 231: I don't think there is an ESLint rule configured for this and we have Prettier so not sure about the rules, but maybe we should be a bit strict about using if / else with returns... here it looks like it can use some more spacing or indentation with `{ }` etc. for readability

backend/lib/endpoints/posts.js
```diff
@@ -127,19 +216,102 @@ async function routes(app) {
...
+      preValidation: [app.authenticate],
+      schema: updatePostSchema,
+    },
     async (req, reply) => {
-      // todo: make sure user can only update their own post
+      // TODO: get userId from jwt
+      // userId = ?
+      // user = User.findById(userId);
+      var user = await User.findById("5ea6900c0e0419d4cb123611");
+      if (user === null) return new httpErrors.Unauthorized();
```
robinv85, line 228: this won't be needed and should be handled by the auth hook

backend/lib/endpoints/posts.js
```diff
@@ -127,19 +216,102 @@ async function routes(app) {
...
-      Object.keys(req.body).forEach((key) => {
-        if (post[key] && post[key] !== req.body[key]) {
-          post[key] = req.body[key];
+      if (!post) return new httpErrors.NotFound();
+      else if (post.author.authorId != user.id)
+        return new httpErrors.Forbidden();
+      var body = req.body;
+
+      // ExpireAt needs to calculate the date
+      if ("expireAt" in body) {
```
robinv85, line 237: I think it would be better to use  `if (body.hasOwnProperty("expireAt")`)

backend/lib/endpoints/posts.js
```diff
@@ -127,19 +216,102 @@ async function routes(app) {
...
+        switch (body.expireAt) {
+          case "day":
```
robinv85, line 239: same as previous comment with switch case or use object as map
this one can definitely be extracted to a reusable utility function (or maybe we can use something from another date library directly)  Like @joaofnds said probably Moment or date-fn's are widely used and have a lot of features, I've also seen https://moment.github.io/luxon/ which is a smaller and newer alternative to moment... so depends a bit what we need

backend/lib/endpoints/posts.js
```diff
@@ -127,19 +216,102 @@ async function routes(app) {
...
+            break;
+          case "month":
+            body.expireAt = (1).months().fromNow();
+            break;
         }
-      });
-      return post.save();
+      }
+
+      return Object.assign(post, body).save();
```
robinv85, line 251: The point of having `if (post[key] && post[key] !== req.body[key])` was some additional security to prevent this request from simply adding anything coming out of the request.body and overwriting it directly... 
I suppose there could be some edge cases here but I think the schena's should take care of validating the request body so except for some data transformations this could work... but it seems like error-prone to me
The `post` object here is an instance of a Mongo document with prototype methods such as `.save()` so I think we should use something like `post.overwrite(body).save()` instead ( see https://mongoosejs.com/docs/documents.html )

backend/lib/endpoints/posts.js
```diff
@@ -127,19 +216,102 @@ async function routes(app) {
...
+    async (req) => {
+      // TODO: get userId from jwt
+      // userId = ?
+      // user = User.findById(userId);
+      var user = await User.findById("5ea6900c0e0419d4cb123611");
+      if (user === null) return new httpErrors.Unauthorized();
+
+      const { postId, userId } = req.params;
+      if (postId === null || userId === null)
+        return new httpErrors.BadRequest();
```
robinv85, line 270: so I think these things are handled by the auth and schema validation now (and on my recent PR now on master I just checked with req.userId here)
If these aren't currently handled by `null` values in the request we might need to adjust the schema's, I 'm not sure if the default is `nullable`

backend/lib/endpoints/posts.js
```diff
@@ -40,42 +57,88 @@ async function routes(app) {
...
+      // userId = ?
+      // user = User.findById(userId);
+      user = await User.findById("5ea6900c0e0419d4cb123611");
+      if (user === null) return new httpErrors.Unauthorized();
+
+      var postProps = req.body;
+
+      // Creates embedded author document
+      postProps.author = {
+        authorId: user.id,
```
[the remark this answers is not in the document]
Naraujo13: I think it probably should be `author.name`, because here we have a doc to represent the author's info, regardless of its origin.

For the sync issues, I've put a callback when setting the name at the user model that would handle this, making the data to be always in sync. 
Since the user model is not being migrated yet, we don't have that callback at the user model, but that should handle that and I can cover it in the next PR when changing the User model.

About being either `user.name` or `organization.name`, you mean because both could author posts?  I don't know if I got what you meant.
Here I was going for the current v2 definition with both of them having the same base model, this way we wouldn't have to worry about it here since we could only fetch using the base model.

backend/lib/endpoints/posts.js
```diff
@@ -40,42 +57,88 @@ async function routes(app) {
...
+        case "month":
+          postProps.expireAt = (1).months().fromNow();
+          break;
+      }
+
+      // Initial empty likes array
+      postProps.likes = [];
+
+      var post = new Post(postProps);
+      if (post.save()) {
```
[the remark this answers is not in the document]
Naraujo13: Good point for the async.
About the return, setting the response and then just returning the the result of `save()` wouldn't make the request return 201 even if save fails?

backend/lib/models/index.js
```diff
@@ -4,13 +4,15 @@ const { join } = require("path");
...
 module.exports = function registerModels(mongoConnection) {
+  require("./schemas/v2/post");
   readdirSync(PATH_TO_SCHEMAS).forEach((fileName) => {
     const filePath = join(PATH_TO_SCHEMAS, fileName);
 
     if (statSync(filePath).isDirectory()) return;
 
     const schema = require(filePath); // eslint-disable-line import/no-dynamic-require
-    const modelName = fileName.replace(".js", "");
+    var name = fileName.replace(".js", "");
```
robinv85, line 14: forgot a `var`
maybe we should add a linting rule if we don't want to allow this anyway
Also, maybe we should add a ticket to refactor this `registerModels` function, and eikther import the files all manually or update it so we have for every model, both a schema and a model file, what do you think is best @Naraujo13 ?

Naraujo13: Yep, probably a good idea to refactor.
I think that each file should export its schema and model and then `registerModels` can just import them all manually. I think manually would because the looping approach may be harder to control and may cause some undesired side-effects. Also, there are only a few models we should require, not something that lengthy, so I think the simpler approach would be best.

Naraujo13: Also, I forgot to add but I will change to `const` there but will leave the whole refactoring of the `registerModels` to another PR.

robinv85: yeah I would like to work on that, I can make a ticket for it
also we can add this https://eslint.org/docs/rules/no-var

backend/lib/models/schemas/v2/author.js
```diff
@@ -4,16 +4,16 @@ const { schema: locationSchema } = require("./location");
...
     ref: "User",
     required: true,
     type: ObjectId,
   },
-  authorName: {
+  name: {
     required: true,
     type: String,
   },
-  authorType: {
```
robinv85, line 16: I find the word `type` is quote ambiguous here and could mean many things, might be confusing or limiting for future changes.
Also because of the merging of "organization" with "indivudual" as  user... since there are 2 different sub types of a user, maybe those should have the types that only apply to a individual or organization user?

Naraujo13:
> ok this looks definitely good to merge to the other data model branch
> I think it's probably better to merge the whole posts API update as one pull request to master asap so we can continue integrating with frontend and get testing while you work on other parts of the data model ( also, don't think we need `schema/v2` folder? can probably overwrite the old one... it was kind of a draft version, in MVP it will kind of be v1 anyway :)
> I can work on the mongoose integration to load the schema's tomorrow so will help with that

About the v2 folder, this is to avoid the `index.js` trying to create a model out of something that is already exporting it (it does this with all files in the root schema folder). I agree there is no need for the folder, but while we don't refactor the whole `index.js`, unfortunately, it is needed.

PR #197, opened 2020-04-29
Adds pagination to the posts API

### LINKED ISSUE #126: Implement Pagination on the Backend
We need to implement pagination when the user loads posts. We'll implement lazy loading on the frontend in https://github.com/FightPandemics/FightPandemics/issues/127 

@robinv85 thinks this might be best handled using mongo "skip" and "limit" parameters https://fightpandemics.slack.com/archives/C010V0DDHLN/p1586966620489600?thread_ts=1586966336.487500&cid=C010V0DDHLN

Let's shoot for loading 5 posts at a time

Check out `backend/lib/endpoints/schema/posts.js`:
```
const getPostsSchema = {
  querystring: S.object()
    .prop("authorId", S.string())
    .prop("skip", S.integer())
    .prop("limit", S.integer()),
};```

### PR DESCRIPTION
**Pull request by mushketyk**
Closes #126 

### PR Checklist:

* [x] Have you checked out the guidelines in our [Contributing](../CONTRIBUTING.md) document?
* [x] Have you looked if there might be other open [Pull Requests](../../../pulls) for the same update/change?

* To help understand them.

### DIFF
--- a/backend/lib/endpoints/posts.js
+++ b/backend/lib/endpoints/posts.js
@@ -27,15 +27,9 @@ async function routes(app) {
   // /posts
 
   app.get("/", { schema: getPostsSchema }, async (req) => {
-    // todo: add limit, skip, visibility filter, needs, tags ...
+    // todo: add visibility filter, needs, tags ...
     const { authorId, limit, skip } = req.query;
     const aggregates = authorId ? [{ $match: { authorId } }] : [];
-
-    const skipNum = skip || 0;
-    const limitNum = limit || DEFAULT_LIMIT;
-    // Total number of items Mongo will consider, but then will skip the
-    // first "skip" items before returning a result
-    const totalLimit = skipNum + limitNum;
 
     aggregates.push(
       {
@@ -47,10 +41,10 @@ async function routes(app) {
         },
       },
       {
-        $limit: totalLimit,
+        $skip: skip || 0,
       },
       {
-        $skip: skip || 0,
+        $limit: limit || DEFAULT_LIMIT,
       },
       {
         $project: {
--- a/backend/lib/endpoints/posts.js
+++ b/backend/lib/endpoints/posts.js
@@ -33,18 +33,18 @@ async function routes(app) {
 
     aggregates.push(
       {
+        $skip: skip || 0,
+      },
+      {
+        $limit: limit || DEFAULT_LIMIT,
+      },
+      {
         $lookup: {
           as: "comments",
           foreignField: "postId",
           from: "comments",
           localField: "_id",
         },
-      },
-      {
-        $skip: skip || 0,
-      },
-      {
-        $limit: limit || DEFAULT_LIMIT,
       },
       {
         $project: {
--- a/backend/lib/endpoints/posts.js
+++ b/backend/lib/endpoints/posts.js
@@ -14,6 +14,8 @@ const {
   addCommentSchema,
 } = require("./schema/posts");
 
+
+const DEFAULT_LIMIT = 10;
 /*
  * /api/posts
  */
@@ -26,8 +28,15 @@ async function routes(app) {
 
   app.get("/", { schema: getPostsSchema }, async (req) => {
     // todo: add limit, skip, visibility filter, needs, tags ...
-    const { authorId } = req.params;
+    const { authorId, limit, skip } = req.query;
     const aggregates = authorId ? [{ $match: { authorId } }] : [];
+
+    const skipNum = skip || 0;
+    const limitNum = limit || DEFAULT_LIMIT;
+    // Total number of items Mongo will consider, but then will skip the
+    // first "skip" items before returning a result
+    const totalLimit = skipNum + limitNum;
+
     aggregates.push(
       {
         $lookup: {
@@ -36,6 +45,12 @@ async function routes(app) {
           from: "comments",
           localField: "_id",
         },
+      },
+      {
+        $limit: totalLimit,
+      },
+      {
+        $skip: skip || 0,
       },
       {
         $project: {
--- a/backend/lib/endpoints/schema/posts.js
+++ b/backend/lib/endpoints/schema/posts.js
@@ -3,7 +3,9 @@ const S = require("fluent-schema");
 const getPostsSchema = {
   querystring: S.object()
     .prop("authorId", S.string())
+    // Number of items to skip
     .prop("skip", S.integer())
+    // Number of items to return
     .prop("limit", S.integer()),
 };
 

### CODE REVIEW

**Comment by mushketyk** [comment]
@Naraujo13 I've updated my PR. Could you review it again?

File: backend/lib/endpoints/posts.js [hunk 1]
```diff
@@ -37,6 +40,12 @@ async function routes(app) {
           localField: "_id",
         },
       },
+      {
+        $skip: skip || 0,
+      },
+      {
+        $limit: limit || DEFAULT_LIMIT,
+      },
```
Naraujo13, line 48: That's better!

I would just recommend putting these two steps before the lookup stage there.
So the pipeline would look something like this:
```
    aggregates.push(
      {
        $skip: skip || 0,
      },
      {
        $limit: limit || DEFAULT_LIMIT,
      },
      {
          $lookup: {
            as: "comments",
            foreignField: "postId",
            from: "comments",
            localField: "_id",
         },
       },
       // rest of the pipeline as is
    )
```
It's just changing the execution order.
This way the lookup will only be executed for the posts that are within our page range.

backend/lib/endpoints/schema/posts.js
```diff
@@ -3,7 +3,9 @@ const S = require("fluent-schema");
 const getPostsSchema = {
   querystring: S.object()
     .prop("authorId", S.string())
+    // Number of items to skip
```
robinv85, line 6: I don't think there should be comments here, I'm  more in favor of the "clean code" principles to strive for self-documenting code and only add comments when absolutely needed

Naraujo13:
I've just thought that we may need pagination at the `/posts/:postId` endpoint also.
There we retrieve the post alongside its comments, maybe it would be a good idea to paginate the comments? 
That way we can lazy load the comment tree to avoid any performance issues with big threads discussions.

What do you think @robinv85 @mushketyk @anniegiang ?

backend/lib/endpoints/schema/posts.js
mushketyk: Good point. Will remove the comment.

robinv85:
Yeah good point @Naraujo13 I also thought we might need pagination for comments, because lots of comments should probably be lazy loaded. 
However I wonder if there is a good way to make this clear in the REST  API that the parameters apply to a sub resource, something like `api/posts?comments.skip=0&comments.limit=10`

mushketyk:
@robinv85 Since `api/posts/:postId` returns information about the post and comments, this seems a bit counterintuitive. Would it be better to have another endpoint for comments pagination: `api/posts/:postId/comments?skip=0&limit=10`?

robinv85:
@mushketyk yes I think that's the way to go, do do requests for lazy-loading pages of comments from the frontend...
Howver I'm not sure about the params on `/api/posts/:postId` since that endpoint also returns comments, we could add query params to specify how many comments we want (page size, e.g.  `?commentsCount=` ... or just use a default, but the client needs to know where to start, so the easiest would be something like returning page 1 and fetching the next page.

mushketyk:
@robinv85 Maybe this is a bit off topic, but did you consider GraphQL? It can be a good fit for this problem: subscription allows us to fetch multiple item types at the same type, etc.

mushketyk:
@robinv85 For the current approach, could the frontend perform two queries:

* `posts/:postId?commentsCount=10` - to get a particular post and 10 comments
* `posts/:postId/comments?skip=10&limit=10` - to get the next 10 comments

Do you think this is reasonable or would you suggest to simplify the API?

robinv85:
@mushketyk yes I've mentioned GraphQL already a few times on Slack but we decided for MVP not to add any more different technologies, we started in a hackathon with a basic stack with Express, Mongo, etc. and only thing that we changed is Fastify mostly for the async/await functionality.

I think sub resources in REST API's are often more complex and subject for discussions, so I like that solution which is pretty simple to start with something. so the default response could just include the first page of comments and then the client can query for the next pages. Will that be accurate if there are newer comments since previous page was fetched though?

mushketyk:
> Also, I'm thinking about "newest first" sorting
> Will that be accurate if there are newer comments since previous page was fetched though?

What if we use range queries instead of using `skip`? Mongo objects have _id fields that have the following structure:

> a 4-byte timestamp value, representing the ObjectId’s creation, measured in seconds since the Unix epoch
> a 5-byte random value
> a 3-byte incrementing counter, initialized to a random value

We can use the following range query to get *comments after the oldest comment we have*: 

```js
  db.students.find( { _id: { $lt: startValue } } )
             .sort( { _id: -1 } )
             .limit( nPerPage )
```

From [here](https://docs.mongodb.com/manual/reference/method/cursor.skip/).

robinv85:
Yeah sounds like a good idea, @Naraujo13 what do you think about this way for pagination?

Naraujo13:
I like that endpoint suggestion also, no problem with that.

About the newer comments and sorting discussion, I think I'm missing something here.
The comments are sorted from older to the newer right?  At least that is the standard at Facebook, Slack and etc.
If so, comments added while watching the discussion would just be added on the last page. Am I missing something here @robinv85 ?

I was thinking if we would have the same problem at posts, but they are sorted now by distance to the user, so we won't have the same problem there.

Naraujo13:
Also adding up, using the _id for paging when sorting is done by date is usually a good idea though.

mushketyk:
> If so, comments added while watching the discussion would just be added on the last page. Am I missing something here @robinv85 ?

If comments are displayed older to newer, then there is no problem with pagination, and we can either use `skip`&`limit` or `_id`&`limit` parameters. If they are sorted newer to older then we might get duplicates when we iterate through them using pagination since a second API call might skip over a newly added comment.

Using `_id`&`limit` over `skip`&`limit` should make a search a bit faster though.

> I was wondering if we would have the same problem at posts, but they are sorted by distance to the user now, so I think we won't have the same problem there.

I think we might have a problem if a post is added in between two API calls, and we might get duplicated results. This might not be a big issue, and we can just skip it for now. If we want to solve it I think we would need to use a different set of pagination parameters to ensure that we don't have duplicates. Maybe specify `limit`&`minDistance` when iterating over posts?

mushketyk:
@Naraujo13 @robinv85 This discussion turned out to be quite long! Let me summarize the points of discussion and try to see what we agree disagree on:

* We need to have pagination on both posts and comments.
* * I think we all agree here
* We can either have a single `posts/:postId` endpoint to get a post and the first batch of comments, or we can have an additional endpoint `posts/:postId/comments` to paginate over points.
* * Not sure what did we decide on this one
* Ordering and pagination - incorrect pagination can lead to duplicate results
* * If we need to order comments from oldest to newest. In this case, we can use `skip`, `limit` parameters, and this should not cause any problems
* * * Are we agree on this?
* * Do we want to order posts by distance to a user?
* * * Do we need to implement special pagination for this?

Is this all, or do I miss something?
What do you think about these points?

robinv85:
@mushketyk @Naraujo13 Yes seems good, I like the idea to use `_id` instead of `skip` to prevent errors.
Using 2 endpoints also seems ok, separate endpoint to load more comments.
And for comments it's probably oldest to newest so not a problem for pagination, for posts I would think that it's probably location + newest first? And the filters are configurable so maybe sorting as well, but that's out of scope here, just have to think that it has to be a bit flexible.

Naraujo13:
Yeah, I also agree with all.

mushketyk:
Thank you for your comments @Naraujo13 and @robinv85 ! I think we are on the same page.

I'll close this PR, since there are too many comments already and will open another PR with these comments addressed.

PR #200, opened 2020-04-29
home page for desktop 

TenNga:
### PR Checklist:

* [ ] Have you checked out the guidelines in our [Contributing](../CONTRIBUTING.md) document?
* [ ] Have you looked if there might be other open [Pull Requests](../../../pulls) for the same update/change?

responsive home page for desktop 

added new styling to match the design from Figma. Used different unit to make the page responsive to multiple screen size (Desktops)
* To help understand them.


client/src/pages/Home.js
```diff
@@ -28,12 +46,36 @@ const StyledWelcome = styled(Heading)`
   line-height: 3rem;
   margin: 2.5rem auto 0;
   text-align: center;
+
+  @media only screen and (min-width: 600px) {  
+    font-size: 2.8vw;
```
anniegiang, line 51: For typography related styles, please grab them from `theme.js`. 

On a side note, let's stick with using `rem`. We currently have a root font size of `62.5%` to help with with responsiveness and scaling sizes (`1rem = 10px`).  So far we've been using `rem` for almost everything, though perhaps we should probably use `vw` and `vh` where appropriate. 

Also based on [MDN](https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Values_and_units), it looks like they describe that viewport values are for things like width, height, margin, and padding (though I'm not sure how strict this is).

client/src/pages/Home.js
```diff
@@ -43,24 +85,43 @@ const StyledP = styled.p`
...
+    text-align: left;
+    color: white;
+  }
 `;
 
 const OnboardingContainer = styled.div`
   margin-top: 4rem;
+
+  @media only screen and (min-width: 600px){
+    display: "flex";
```
anniegiang, line 101: Don't think the double quotes are needed?

client/src/pages/Home.js
TenNga: I did use rem because of the same reason you mention but when I change the screen from 1440 to 1024 text size stays the same and doesn't match the design. I can write media for each different screen if that's a better option.

client/src/pages/Home.js
TenNga: ok

client/src/pages/Home.js
anniegiang: Ah ok! Can you try the option of writing a media please?  I know it seems like an overkill but I think it's best to stay consistent with using `rem`.

robinv85: I did notice we also have a bunch of mediaquery sizes defined in the `constants/theme.js` and those are set in pixels
maybe these are also reusable?

client/src/pages/Home.js
```diff
@@ -43,24 +85,43 @@ const StyledP = styled.p`
...
+    width: 100%;
+    margin: auto 0;
+  }
 `;
 
 export const Home = (props) => {
   console.log("render home", { props });
   return (
-    <div className="text-center">
+    <MainContainer className="text-center">
```
robinv85, line 111: I also noticed this class while trying to clean up the styles, wonder if we still need these classes defined globally or if there is a better way if we use StyledComponents with something reusable as a component perhaps?

client/src/pages/Home.js
anniegiang: Ah didn't know about that the media queries in `theme.js`. If so let's use them

client/src/pages/Home.js
TenNga: I replaced with styled-component

client/src/pages/Home.js
TenNga: Replaced all the value of media queries with theme.js values

MohamedEl1:
@TenNga there seems to be some merge conflicts

PR #201, opened 2020-04-29
Cookie message

### LINKED ISSUE #188: Cookie Message that displays on the top or bottom of the page when it is first opened
@EdwinMata, our lawyer, asked to open this issue.

TO DO:
* Connect with @Yu on Slack for design.
* Message needs to say something like: "We use cookies. By continuing to browse this site you are agreeing to our use of cookies. Read more."
* Read more should redirect user to /cookies-policy

nadinejuraschek:
closes #188 

* uses localStorage to ensure that the message is only displayed the first time the user visits our website


client/src/components/CookieAlert.js
```diff
@@ -0,0 +1,67 @@
...
+const CloseButton = styled.button`
+  align-self: start;
+`;
+
+export default () => {
+  const [active, setActive] = useState();
+
+  const hideMessage = (event) => {
+    active ? setActive(false) : setActive(true);
+    localStorage.setItem("cookieMessage", "true");
```
anniegiang, line 39: @robinv85 Just wanted to double check - is it fine to use localStorage for production? I remember reading a comment in the code base somewhere about localStorage not working well with production.

client/src/components/CookieAlert.js
```diff
@@ -0,0 +1,67 @@
...
+  color: ${colors.selago};
+  text-decoration: underline;
+`;
+
+const CloseButton = styled.button`
+  align-self: start;
+`;
+
+export default () => {
+  const [active, setActive] = useState();
```
anniegiang, line 35: I think here we can set `active` initially as `true`, and then on line 38, we can replace the one-liner if-statement with `setActive(false)`,  since the the primary function of `hideMessage` is to handle the close button.

client/src/components/CookieAlert.js
```diff
@@ -0,0 +1,67 @@
...
+
+  const hideMessage = (event) => {
+    active ? setActive(false) : setActive(true);
```
anniegiang, line 38: I think we can replace this with `setActive(false)`,  since the the primary function of `hideMessage` is to handle the close button.

client/src/components/CookieAlert.js
robinv85: Not sure what you mean with it not working, I first used local storage for some things like storing the answers to the wizards, email input etc. ... which was just a temporary solution because I didn't know what we were supposed to do with the data and the backend wasn't ready yet...
So that said, there is also a security concern for storing things like a JWT which is not relevant here
And as far as legal implications for the cookie law I'm also not sure if this applices to local storage, but I think it's probably just fine, but maybe we need to keep track of every data we store and where, for the legal team so they know to find everything

robinv85: Also if I think more about it, for anonymous users this probably works best
For a logged-in user we might even want to save it with their user data so if they log in on another browser or device they don't need to accept it again, but that's also something we should check with legal

anniegiang: Ah I didn't mean local storage isn't working, I just wanted to take caution because there's a comment in `authActions.js`:

```
// Note: for production apps, both localstorage & cookies contain risks to store user & auth data
```
But it makes sense now that you mention auth in the picture. So this PR should be ok to merge.

PR #209, opened 2020-04-30
Normalize Redux state shape

anniegiang:
- Setup `combineReducers` to achieve a normalized state shape
  - Moved the auth related reducer from `reducers/index.js` into `reducers/sessionReducer.js`, creating a  `session` slice of state
- Cleared some console warnings about unused variables
- Setup a basic `FeedContainer` to connect `Feed` with Redux
- Printed out the redux state 
  - I tried to setup `redux-logger` but couldn't get it to log in the console... so printing the state out is more of a bandaid fix


client/src/index.js
```diff
@@ -14,6 +14,7 @@ import * as serviceWorker from "./serviceWorker";
 import rootReducer from "./reducers";
 
 const store = createStore(rootReducer, applyMiddleware(thunk));
+console.log("redux state: ", store.getState());
```
robinv85, line 17: forgot log?

client/src/templates/RouteWithSubRoutes.js
```diff
@@ -58,9 +58,9 @@ export const RouteWithSubRoutes = (route) => {
 };
 
 const mapDispatchToProps = {};
-const mapStateToProps = ({ emailVerified, isAuthenticated }) => ({
-  emailVerified,
-  isAuthenticated,
+const mapStateToProps = ({ session }) => ({
+  emailVerified: session.emailVerified,
+  isAuthenticated: session.isAuthenticated,
```
robinv85, line 63: we might need to fine tune this a bit, it's not really a "session:" and the auth & user logic is a bit confusing at the moment but we can still improve that later I guess

client/src/index.js
anniegiang: Left it there on purpose to temporary replace it with redux-logger.. but I can remove it for now.

client/src/templates/RouteWithSubRoutes.js
anniegiang: Ah ok sounds good

client/src/index.js
robinv85: ah ok, i'm used to just check with redux-devtools if needed, i guess adding a logger could also work

PR #215, opened 2020-04-30
Allows to connect debugger to an application in the dev mode

mushketyk:
Closes #215 

### PR Checklist:

* [x] Have you checked out the guidelines in our [Contributing](../CONTRIBUTING.md) document?
* [x] Have you looked if there might be other open [Pull Requests](../../../pulls) for the same update/change?

Allow developers to attach a debugger to a running process in dev. mode

* Adds `--inspect` flag to `nodemon` process to allow to attach a debugger

### DIFF
--- a/backend/package.json
+++ b/backend/package.json
@@ -6,7 +6,7 @@
   "main": "server.js",
   "scripts": {
     "start": "node server.js",
-    "dev": "nodemon -L server.js",
+    "dev": "nodemon --inspect=0.0.0.0:9229 -L server.js",
     "test": "jest",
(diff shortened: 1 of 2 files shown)


robinv85:
Great idea, I also thought this wasn't working yet, do you think it should always run on `npm run dev` or maybe add another `npm run debug` as alternative script?

mushketyk:
@robinv85 I would keep the current command. It seems reasonable to be able to debug an app when we run it locally. And I don't expect that we will use this mode anywhere else.

Issue #223, opened 2020-05-01
Fix Ajv configuration without eslint warning


robinv85:
ESLint marked the line `require("ajv")` because of the `no-extraneous-dependencies` rule
I've added an ignore rule for now to be able to commit changes, but this should be fixed.
Fastify includes Ajv by default and this was imported to adjust the configuration, 

This example was taken from possibly outdated documentation
https://www.fastify.io/docs/v2.10.x/Validation-and-Serialization/

There should now be a better way to just pass a configuration object, might require to upgrade fastify
https://www.fastify.io/docs/latest/Validation-and-Serialization/

joshmorel:
This I think has been resolved maybe with #224 ?

PR #224, opened 2020-05-01
Feature/#210 fastify sensible

### LINKED ISSUE #210: Add Fastify sensible plugin for some more consistency and coding conventions
Fastify Sensible is an official plugin that provides some sensible defaults https://github.com/fastify/fastify-sensible
It's very similar to what we are currently doing already, with the `httpErrors` package built-in on the Fastify instance and the `reply` objects.
It comes with some useful things built-in that we can consider which can improve the codebase
- assertions on the fastify instance
- the `to` async/await wrapper for better error handling
- a custom default error handler, instead of the default which returns all error info to the client

robinv85:
Closes #210 

* Added fastify sensible plugin for more consistency and some default configurations
* Updated users and auth endpoints to use the included httpErrors (slightly different syntax than using the http-errors package directtly)
* Also updated the documentation about the backend

To do: added eslint ignore rule for the Ajv configuration because it's importing Ajv which isn't listed in package.json, but is a dependency of fastify. There is a better way to configure this but it's unclear based on different versions of the docs, so opened a new issue to fix this. #223 

Also to do: I didn't update all other handlers for posts because these are WIP in #191 for the data model update so when everything is updated the `http-errors` package can be removed

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,14 +61,19 @@ while developing. This is a living document, so feel free to add any notes that 
 
 ### Backend
 
-* Set `.additionalProperties(false)` for each schema defined with FluentSchema, to return a 400 bad request error if any
-additional properties not defined in the schema are passed in through the request.
-    * We are using [FluentSchema](https://github.com/fastify/fluent-schema) to validate backend requests. This is the
-    default validator for Fastify, our backend framework. FluentSchema uses ajv under the hood, and compiles to the
-    more standard JSON Schema.
+* API requests are validated using [Ajv](https://github.com/epoberezkin/ajv) and JSON schema's are defined with [FluentSchema](https://github.com/fastify/fluent-schema)
+    * By default, the schema's allow additional properties unless explicitly set that additional properties are not allowed.
+        There's a utility function `strictSchema()` in `lib/endpoints/schemas/utils.js` that can be used to initialize all schema's.
     * Rather than silently suppress additional properties, as is the default behavior for Fastify's ajv configuration,
-    we are instead returning a 400 bad request error if additional properties are passed in. This makes it easier to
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -61,14 +61,19 @@ while developing. This is a living document, so feel free to add any notes that
 
 ### Backend
 
-* Set `.additionalProperties(false)` for each schema defined with FluentSchema, to return a 400 bad request error if any
-additional properties not defined in the schema are passed in through the request.
-    * We are using [FluentSchema](https://github.com/fastify/fluent-schema) to validate backend requests. This is the
-    default validator for Fastify, our backend framework. FluentSchema uses ajv under the hood, and compiles to the
-    more standard JSON Schema.
+* API requests are validated using [Ajv](https://github.com/epoberezkin/ajv) and JSON schema's are defined with [FluentSchema](https://github.com/fastify/fluent-schema)
```
mannykary, line 64: Minor: formatting looks a bit off here, maybe you're adding the line breaks manually?

robinv85: Ah yeah I did that on purpose, thought it was better, similar to limiting max line-length in the code (to e.g. 150 chars...) But I'm not sure if this is a best practice for markdown files
The end result seemed the same when parsed so this means the original would also be more readable, seems like this is allowed https://stackoverflow.com/questions/29955688/in-markdown-is-there-an-official-best-practice-as-relates-to-line-length

Seems like some markdown renderers might have different behavior, but on Github it's ignored
see https://stackoverflow.com/questions/41513565/markdown-and-line-wrapping
so we could perhaps implement it as a best practice

robinv85:
I can't seem to add @mushketyk as reviewer

Issue #221, opened 2020-04-30
Resolve console warning in NrMap.js


anniegiang:
```
./src/pages/NrMap.js
  Line 90:32:   Expected to return a value in arrow function  array-callback-return
  Line 105:22:  Expected '===' and instead saw '=='           eqeqeq
  Line 123:28:  Expected to return a value in arrow function  array-callback-return
```

- [ ] Line 90:32:   Expected to return a value in arrow function  array-callback-return

- [x] Line 105:22:  Expected '===' and instead saw '=='           eqeqeq

- [ ] Line 123:28:  Expected to return a value in arrow function  array-callback-return

nadinejuraschek:
@anniegiang This issue can be closed right?

anniegiang:
> @anniegiang This issue can be closed right?

Not sure, I haven't been following that close to the project progress lately. The last time I checked the two checkboxes that aren't checked weren't fixed yet I think

nadinejuraschek:
> > @anniegiang This issue can be closed right?
> 
> Not sure, I haven't been following that close to the project progress lately. The last time I checked the two checkboxes that aren't checked weren't fixed yet I think

NrMap was refactored since it was our last class component, so there are still warnings for NrMap, but different ones. I opened an issue for the new ones a few days ago: #449 

I'm closing this issue

PR #227, opened 2020-05-01
got rid of three console warnings

### LINKED ISSUE #222: Resolve console warning in utils/common.js
```
./src/utils/common.js
  Line 2:24:  Unnecessary escape character: \[  no-useless-escape
  Line 2:49:  Unnecessary escape character: \[  no-useless-escape
```

nadinejuraschek:
closes #222

also gets rid of one console warning from #221, I think the remaining two can be resolved once this last class components gets converted to a functional component


anniegiang:
Awesome thanks for working on this! Just to double check, removing the warnings doesn't affect the original use cases right?

nadinejuraschek:
It shouldn't. It seemed to work fine when I tested it.

PR #229, opened 2020-05-02
Allow unauthenticated users to access a post

anniegiang:
- Remove preValidation in api/posts/:postId

### DIFF
--- a/backend/lib/endpoints/posts.js
+++ b/backend/lib/endpoints/posts.js
@@ -67,43 +67,39 @@ async function routes(app) {
 
   // /posts/postId
 
-  app.get(
-    "/:postId",
-    { preValidation: [app.authenticate], schema: getPostByIdSchema },
-    async (req, reply) => {
-      const { postId } = req.params;
-      const post = await Post.findById(postId);
-      if (post === null) {
-        return reply.send(new httpErrors.NotFound());
-      }
-      post.comments = await Comment.aggregate([
-        {
-          $match: {
-            parentId: null,
-            postId: mongoose.Types.ObjectId(postId),
+  app.get("/:postId", { schema: getPostByIdSchema }, async (req, reply) => {
+    const { postId } = req.params;
+    const post = await Post.findById(postId);
+    if (post === null) {
+      return reply.send(new httpErrors.NotFound());
+    }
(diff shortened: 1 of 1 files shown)


Naraujo13:
I got 2 hours late here, but was this logic correct? 
If I'm not mistaken, the requirements for an unauthenticated user was for him/her to be able to see only 20 words of the post and no comments. 
Wouldn't that need further changes?

robinv85:
@Naraujo13 I also thought that you were working on the new posts backend, mostly database but also some changes in the endpionts.
I was also wondering what we should do if we have an endpoint that could have a different response for logged-in or non-logged-in users because currently the `authenticated` hook immediately denies access for requests without access token so we will have to find a solution for that if we need to support different responses.

anniegiang:
Ah I didn't think about this. If we add back the preValidation, does this route error silently or immediately if a user is not authenticated? Like would it immediately respond with something like a 401 error, "User is not authorized"? 

Another idea is, we could remove the preValidation and have the frontend handle what to be displayed based on logged-in/logged out, though that would mean our backend would give back more data than needed.
 
Any thoughts?

Naraujo13:
I think for `/posts` we definitely need to just know whether the user is logged in or not in order to know if we cap the posts at 20 words. So for this endpoint the hook can't just deny access, but I don't know the optimal way for this.

Also, @anniegiang this PR changed the `/posts/:postId` endpoint to remove the auth, but can unlogged users use this? I thought that unlogged users couldn't see the comments and therefore, couldn't access this endpoint, but I may be mistaken.

robinv85:
@Naraujo13 I've asked this to @manualzuru a few days ago and he told me this might be a misunderstanding because it's not related to either logged-in or not logged-in user, but we always want to show max. 20 words when showing posts in lists, and then allow to expand to show the full post...

Naraujo13:
@robinv85 makes sense to cap to 20 words at the list to everyone, though the [feature requirements](https://www.notion.so/fightpandemics/a729bd46f001461393a0ad707161f299?v=16cb1b0a7a1c4d5ba0a1aea6184cf734&p=653658508c4649869787c48f3e4eee13) for the feed ask to cap only for unlogged users.

I will re-check with them about this

Issue #148, opened 2020-04-20
User Feedback Backend 


indirabhowmick:
Backend: the creation of a database for collecting feedback that lists user (if registered and logged in) or IP-address together with location based on IP address and answers to feedback questions

Requirements:
https://www.notion.so/fightpandemics/User-Feedback-UI-645a5edc39c44d91843d3b6e56f2eff8
Please contact Valentin Rudloff if there are any questions related to the requirements.

DanielAlexander84:
I would like to do this in addition to the frontend [issue](https://github.com/FightPandemics/FightPandemics/issues/117) associated with it to get to know the project better. If anyone has an issue with this or I am taking too long, please let me know. 😄

indirabhowmick:
@DanielAlexander84 awesome. Thanks for volunteering. I think for this you have to create the API to post the user feedback. The table has already been created to store data and should be present in the data model. I will create another ticket to connect the front end with back end - API connectivity. Any questions related to the data model, please contact Nicholas.

Naraujo13:
Hi @DanielAlexander84, if you need any help with the data model, please reach out.

[Here](https://github.com/FightPandemics/DataModel/blob/master/v2/scripts/Mongoose_Script/feedbacks.js) is the definition of the feedback model, already in mongoose, so you can use it to save some time and focus on the endpoints.

DanielAlexander84:
Thank you both for the info. I plan on working on this during the weekend when I will have more free time. 😄

PR #230, opened 2020-05-02
Create endpoint/backend for feedback form

DanielAlexander84:
Closes [148](https://github.com/FightPandemics/FightPandemics/issues/148)

- Adds feedback data model to mongo
- Adds fluent-schema for feedback data coming from the body of request
- Adds `api/feedback` endpoint to be used for form submission
  - Before saving it ensures that the each user and ip can only submit one feedback form. This was 
    discussed with @robinv85 as a desired check.

### DIFF
--- a/backend/lib/endpoints/feedback.js
+++ b/backend/lib/endpoints/feedback.js
@@ -10,9 +10,9 @@ async function routes(app) {
     const { userId } = req.body;
 
     if (userId) {
-      const [err, userFeedback] = await app.to(Feedback.find({ userId }));
-      if (userFeedback[0] || err) {
-        throw app.httpErrors.internalServerError();
+      const [userFeedback] = await app.to(Feedback.findOne({ userId }));
+      if (userFeedback) {
+        throw app.httpErrors.conflict("Feedback already submitted");
       }
     }
 
--- a/backend/lib/endpoints/feedback.js
+++ b/backend/lib/endpoints/feedback.js
@@ -8,7 +8,7 @@ async function routes(app) {
 
   app.post("/", { schema: createFeedbackSchema }, async (req, reply) => {
     const { userId } = req.body;
-    let ip = "undefined";
+
     if (userId) {
       const userFeedback = await Feedback.find({ userId });
       if (userFeedback[0]) {
@@ -16,28 +16,14 @@ async function routes(app) {
       }
     }
 
-    if (req.ip !== null && typeof req.ip !== "undefined") {
-      ip = req.ip;
-      const feedbackFromIp = await Feedback.find({ ipAddress: ip });
-      if (feedbackFromIp[0]) {
-        reply.conflict("Feedback already submitted with this IP address");
-      }
-    }
-
-    if (userId || ip !== "undefined") {
-      return new Feedback({
-        ...req.body,
-        ipAddress: ip,
-      })
-        .save()
-        .then(() => {
-          reply.code(201).send({ success: true });
-        });
-    }
-
-    return reply.internalServerError(
-      "Authenticated user or ip required to save feedback",
-    );
+    return new Feedback({
+      ...req.body,
+      ipAddress: req.ip,
+    })
+      .save()
+      .then(() => {
+        reply.code(201).send({ success: true });
+      });
   });
 }
 
--- a/backend/lib/endpoints/feedback.js
+++ b/backend/lib/endpoints/feedback.js
@@ -1,0 +1,36 @@
+const httpErrors = require("http-errors");
+const { createFeedbackSchema } = require("./schema/feedback");
+
+/*
+ * /api/feedback
+ */
+async function routes(app) {
+  const Feedback = app.mongo.model("Feedback");
+
+  app.post("/", { schema: createFeedbackSchema }, async (req) => {
+    const { userId } = req.body;
+    let ip = "undefined";
+    if (userId) {
+      const userFeedback = await Feedback.find({ userId });
+      if (userFeedback[0]) {
+        return new httpErrors.Conflict("User already submitted feedback");
+      }
+    }
+
+    if (req.ip !== null && typeof req.ip !== "undefined") {
+      ip = req.ip;
+      const feedbackFromIp = await Feedback.find({ ipAddress: ip });
+      if (feedbackFromIp[0]) {
+        return new httpErrors.Conflict(
+          "Feedback already submitted with this IP address",
+        );
+      }
+    }
+    return new Feedback({
+      ...req.body,
+      ipAddress: ip,
+    }).save();
+  });
+}
+
+module.exports = routes;
--- a/backend/lib/endpoints/feedback.js
+++ b/backend/lib/endpoints/feedback.js
@@ -1,4 +1,3 @@
-const httpErrors = require("http-errors");
 const { createFeedbackSchema } = require("./schema/feedback");
 
 /*
@@ -7,13 +6,13 @@ const { createFeedbackSchema } = require("./schema/feedback");
 async function routes(app) {
   const Feedback = app.mongo.model("Feedback");
 
-  app.post("/", { schema: createFeedbackSchema }, async (req) => {
+  app.post("/", { schema: createFeedbackSchema }, async (req, reply) => {
     const { userId } = req.body;
     let ip = "undefined";
     if (userId) {
       const userFeedback = await Feedback.find({ userId });
       if (userFeedback[0]) {
-        return new httpErrors.Conflict("User already submitted feedback");
+        reply.conflict("User already submitted feedback");
       }
     }
 
@@ -21,9 +20,7 @@ async function routes(app) {
       ip = req.ip;
       const feedbackFromIp = await Feedback.find({ ipAddress: ip });
       if (feedbackFromIp[0]) {
-        return new httpErrors.Conflict(
-          "Feedback already submitted with this IP address",
-        );
+        reply.conflict("Feedback already submitted with this IP address");
       }
     }
 
@@ -31,9 +28,14 @@ async function routes(app) {
       return new Feedback({
         ...req.body,
         ipAddress: ip,
-      }).save();
+      })
+        .save()
+        .then(() => {
+          reply.code(201).send({ success: true });
+        });
     }
-    return new httpErrors.InternalServerError(
+
+    return reply.internalServerError(
       "Authenticated user or ip required to save feedback",
     );
   });
--- a/backend/lib/endpoints/feedback.js
+++ b/backend/lib/endpoints/feedback.js
@@ -10,20 +10,22 @@ async function routes(app) {
     const { userId } = req.body;
 
     if (userId) {
-      const userFeedback = await Feedback.find({ userId });
-      if (userFeedback[0]) {
-        reply.conflict("User already submitted feedback");
+      const [err, userFeedback] = await app.to(Feedback.find({ userId }));
+      if (userFeedback[0] || err) {
+        throw app.httpErrors.internalServerError();
       }
     }
 
-    return new Feedback({
-      ...req.body,
-      ipAddress: req.ip,
-    })
-      .save()
-      .then(() => {
-        reply.code(201).send({ success: true });
-      });
+    const [err] = await app.to(
+      new Feedback({
+        ...req.body,
+        ipAddress: req.ip,
+      }).save(),
+    );
+
+    if (err) throw app.httpErrors.internalServerError();
+
+    return reply.code(201).send({ success: true });
   });
 }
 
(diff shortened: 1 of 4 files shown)


backend/lib/index.js
```diff
@@ -40,6 +41,7 @@ module.exports = function createApp(config) {
   app.register(organizations, { prefix: "api/organizations" });
   app.register(posts, { prefix: "/api/posts" });
   app.register(users, { prefix: "/api/users" });
+  app.register(feedback, { prefix: "/api/feedback" });
```
mannykary, line 44: Minor: the route should be registered in alphabetical order. I guess ESLint didn't catch this for some reason...

DanielAlexander84: No worries. That makes sense. I will fix that really quick.

backend/lib/models/schemas/Feedback.js
```diff
@@ -0,0 +1,39 @@
+const { Schema } = require("mongoose");
+const { locationSchema } = require("./Location");
```
Naraujo13, line 2: This is requiring the old location schema, which was blank. 
We should probably replace this with the defined location schema.

backend/lib/endpoints/feedback.js
```diff
@@ -0,0 +1,36 @@
...
+        return new httpErrors.Conflict(
+          "Feedback already submitted with this IP address",
+        );
+      }
+    }
+    return new Feedback({
+      ...req.body,
+      ipAddress: ip,
+    }).save();
+  });
```
Naraujo13, line 33: Was talking with @robinv85 at the posts PR if we shouldn't be setting the response code to `201 - Created` in case of success here

backend/lib/endpoints/feedback.js
```diff
@@ -0,0 +1,36 @@
...
+
+    if (req.ip !== null && typeof req.ip !== "undefined") {
+      ip = req.ip;
+      const feedbackFromIp = await Feedback.find({ ipAddress: ip });
+      if (feedbackFromIp[0]) {
+        return new httpErrors.Conflict(
+          "Feedback already submitted with this IP address",
+        );
+      }
+    }
```
Naraujo13, line 28: What would happen if there was no IP on the request?
We would end up saving it as `undefined` isn't it?
Maybe missing the `else ` case here?

Naraujo13: Also, adding up to that, the Feedback is required to have an IP address or an authenticated User (from which we can get the location)

DanielAlexander84: If there is no ip then it will be undefined, correct. I was unsure of what a desirable name would be otherwise. I don't need an else, because of the fact that I create ip before this if statement.

backend/lib/endpoints/feedback.js
DanielAlexander84: Shouldn't? It is currently a `200 OK` response.

backend/lib/endpoints/feedback.js
```diff
@@ -0,0 +1,36 @@
+const httpErrors = require("http-errors");
```
robinv85, line 1: please use the updated `httpErrors` that are included int he `fastify-sensible` package (see readme for more info, I haven't updated every endpoint yet because there are already some pull requests in progress)

backend/lib/endpoints/feedback.js
robinv85: yeah 201 would be better indeed

backend/lib/endpoints/feedback.js
```diff
@@ -0,0 +1,42 @@
...
+    let ip = "undefined";
+    if (userId) {
+      const userFeedback = await Feedback.find({ userId });
+      if (userFeedback[0]) {
+        return new httpErrors.Conflict("User already submitted feedback");
+      }
+    }
+
+    if (req.ip !== null && typeof req.ip !== "undefined") {
+      ip = req.ip;
```
anniegiang, line 21: From a security standpoint, should we double check what `req.ip` is before we save it to our db? Also, what if the IP address is from a public device accessible by anyone?

Naraujo13: Good point @anniegiang, maybe we can validate that on this on the schema request with `.format`? It seems that we can have [IPV4 and IPV6 formats](https://github.com/fastify/fluent-schema/blob/master/docs/API.md),

backend/lib/endpoints/feedback.js
DanielAlexander84: I just looked at that and saw that there is the option to use `reply` instead of throwing errors. What do you think about that? It looks a bit slimmer and still express what we need.

backend/lib/endpoints/feedback.js
DanielAlexander84: What kind of checks do you imagine having here? I am just unsure as to what would be best, because it is coming directly from fastify and I don't know what potential security concerns would be present.

anniegiang: @Naraujo13 
Adding a validation for IP addresses using `.format` sounds like a good first step. 

Allowing at most 1 submission for a logged-in user sounds reasonable. 

Regarding public devices - other than validating the IP address, we also need to make sure the user is unique. Perhaps we could generate a unique id for anonymous/not-logged users?

anniegiang: @DanielAlexander84 

I was imagining what @Naraujo13 pointed out - validating the format of the address and to at least make sure what's coming in is not malicious. 

Another security concern I just discovered is [IP spoofing](https://en.wikipedia.org/wiki/IP_address_spoofing), which I realize may be out of scope for a simple feedback form.

backend/lib/endpoints/feedback.js
DanielAlexander84: I updated it to use fastify-sensible.

backend/lib/index.js
robinv85: Actually, ESLint is currently only checking the "object keys" order, there are additional rules that we can add such ass alphabetic imports... not sure if this works for every part in the code such as calling methods here, but it might help.
For plugins it's probably not always possible since the order might matter (and these routes are added the same as plugins for fastify... didn't find any better way to group the routes)

backend/lib/models/schemas/Feedback.js
robinv85: is that schema on master or is that still on a separate branch @Naraujo13 ?

backend/lib/endpoints/feedback.js
robinv85: There is always the option to use `reply` but it's not recommended to use it together with async/await.
If you read through the fastify docs about async/await, promises etc. and route handlers then you will see they recommend to use return and throw, without the `reply`, also because you could combine `return reply...` but it might get complicated.
see https://www.fastify.io/docs/latest/Routes/#async-await 
and https://www.fastify.io/docs/v2.6.x/Reply/#async-await-and-promises
and https://www.fastify.io/docs/latest/Routes/#promise-resolution
Also in fastify-sensible docs there are some examples, and to me it makes most sense to prefer only `return value` or `throw error` ... in this way we can also do proper error propagation, but I'm still trying to figure out the best systematic approach with fastify & async/await, suggestions welcome ofcourse if you think you can improve it.

backend/lib/index.js
mannykary: Ah okay that makes sense.

backend/lib/endpoints/feedback.js
robinv85: yeah I think it's a great idea to use that IP validation, we can use this for many things coming in the request and I believe we should.
The current version of JSON schema has a lot of validation already built-in it seems, such as e-mail addreses and such, so we should definitely leverage that where we can, and we can also add custom regexes etc. to be more specific instead of just requiring "string:". 
We should also try to specify in the schema where null is allowed and where ist isn't.

backend/lib/endpoints/feedback.js
```diff
@@ -0,0 +1,44 @@
...
+async function routes(app) {
+  const Feedback = app.mongo.model("Feedback");
+
+  app.post("/", { schema: createFeedbackSchema }, async (req, reply) => {
+    const { userId } = req.body;
+    let ip = "undefined";
+    if (userId) {
+      const userFeedback = await Feedback.find({ userId });
+      if (userFeedback[0]) {
+        reply.conflict("User already submitted feedback");
```
robinv85, line 15: yeah so this is not a good idea because the code is not returning here inside the if, so it's really not clear what happens below this block of code... if that code will be executed or not. 
This kind of if block would fit better with "early return" and I would still prefer the built-in http errors combined with the return for consistency (and like explained in the comment about fastify best practices with async/await).

robinv85: I do like that this is using the shorter `reply.[httpError]` syntax so maybe we can use this everywhere but maybe use `throw` since it's an error and should not continue on the next line?

backend/lib/endpoints/feedback.js
```diff
@@ -0,0 +1,44 @@
...
+    if (req.ip !== null && typeof req.ip !== "undefined") {
+      ip = req.ip;
+      const feedbackFromIp = await Feedback.find({ ipAddress: ip });
+      if (feedbackFromIp[0]) {
+        reply.conflict("Feedback already submitted with this IP address");
+      }
+    }
+
+    if (userId || ip !== "undefined") {
+      return new Feedback({
```
robinv85: for consistency probably also better to stick with await syntax here, but we do have to find a good way to use the `reply.code()` combined with the `model.save()` 
I think we should try to use this  "to" utility from fastify-sensible since it's exactly what we need and then we can just `reply.code();` and `return ...` on success.
https://github.com/fastify/fastify-sensible#to

backend/lib/endpoints/feedback.js
```diff
@@ -0,0 +1,36 @@
...
+async function routes(app) {
+  const Feedback = app.mongo.model("Feedback");
+
+  app.post("/", { schema: createFeedbackSchema }, async (req, reply) => {
+    const { userId } = req.body;
+
+    if (userId) {
+      const [err, userFeedback] = await app.to(Feedback.find({ userId }));
+      if (userFeedback[0] || err) {
+        throw app.httpErrors.internalServerError();
```
robinv85, line 15: ok this is the same as my other recent changes...
I think this might be the best one, even if it's not as shorts as using `reply.notFound()` etc. but because we also `throw` then it's clear that it's actually an error ... well just a nitpick,

backend/lib/endpoints/feedback.js
```diff
@@ -0,0 +1,36 @@
...
+      new Feedback({
+        ...req.body,
+        ipAddress: req.ip,
+      }).save(),
+    );
+
+    if (err) throw app.httpErrors.internalServerError();
+
+    setImmediate(() => {
+      reply.code(201).send({ success: true });
```
robinv85, line 29: why do you need this `setImmediate`? I would think you could directly `return reply` or maybe because that was not recommended by fastify docs for async/await, just do

```
reply.code(201);
return { success: true };
```

Also, I think usually the newly created item is returned, although for a feedback this is probably not used on the frontend...
but yeah I have to admit I don't really like that this feels like a workaround if we just want to return a different status code, maybe there's a better way, not that important though

backend/lib/endpoints/feedback.js
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    const [err] = await app.to(
+      new Feedback({
+        ...req.body,
+        ipAddress: req.ip,
+      }).save(),
+    );
+
+    if (err) throw app.httpErrors.internalServerError();
```
robinv85, line 26: Yeah I think this is much better, you could even `throw error` with that generic catch handler it would just return a 500 with generic message too, but maybe better to log it. And if we want we could still handle special cases here differently
so I think something like this would give us useful error info
```
if (err) {
  req.log.error("Failed submitting feedback", { err }); 
}
``

backend/lib/endpoints/feedback.js
DanielAlexander84: So the `setImmediate` was something taken directly from the docs with regard to `return reply`. But your suggestion is also pretty nice. Maybe we can find something nicer in the future.

backend/lib/endpoints/feedback.js
DanielAlexander84: It does make more sense that it is immediately recognizable as an error due to `throw` being there. `reply` could theoretically be anything.

DanielAlexander84:
@robinv85  thanks for the suggestion with `findOne()`. That is nice to not have to step into the array. And yeah, that check was checking for a truthy value inside of the array. I will be leaving it as `if (userFeedback)` because with `findOne` it is just an object being returned.

PR #235, opened 2020-05-03
Add environment variable

### LINKED ISSUE #234: Setup environment variables
We currently do not have `ENVIRONMENT` variable in our configuration, and it should be added.

anniegiang:
Closes #234


backend/config.js
```diff
@@ -11,6 +11,7 @@ const configData = envSchema({
     .prop("AUTH_DOMAIN", S.string().required())
     .prop("AUTH_SECRET_KEY", S.string().required())
     .prop("AUTH_STATE", S.string().required())
+    .prop("ENVIRONMENT", S.string().default("dev").required())
```
mushketyk, line 14: I would avoid setting a default value for an environment. If it is not defined, that would seem like an error that we should know about and fix it.

mannykary:
Does this also need to be set in the deployment to the `review` and `staging` environments?

mannykary:
Also, another thought, how is this different from `NODE_ENV`?

anniegiang:
I believe `NODE_ENV` and `ENVIRONMENT` are the same (at least in meaning). Did we have a variable for `NODE_ENV` already? To be consistent should we stick with `NODE_ENV ` ?

mannykary:
I think `NODE_ENV` is more standard. I do see it being used in our repo in a few places.

PR #245, opened 2020-05-04
Convert last class components to functional components #203

### LINKED ISSUE #203: Convert last class components to functional components
These remaining class components need to be refactored:
- [x] ~/pages/AirTableCOVID
- [x] ~/pages/NrMap

agmaal1:
PR Checklist:

[x ] Have you checked out the guidelines in our Contributing document?
[x ] Have you looked if there might be other open Pull Requests for the same update/change?
Quick Description of the PR
Convert last class components ~/pages/NrMap and ~/pages/AirTableCOVID to functional components. closes #203 

An Overview of the Changes
Converted class components to functional components via hooks.
other changes:

~/pages/NrMap

added key attribute in card child components
refactored return mapping plcDTL in place which make it more readable.
~/pages/AirTableCOVID

modified key attribute in card child components for not using index which might lead to render conflicts. Instead applied a unique key using each record id.
Possible issues encountered in ~/pages/NrMap
_ the component is not displaying the user location but the default. getLocation is implemented but is not getting displayed.

_ The component is not displaying a card for each marker in the map. Google MAPS API gives a OVER_QUERY_LIMIT response when requesting details about some of the hospitals.
https://developers.google.com/maps/premium/previous-licenses/articles/usage-limits


client/src/pages/AirTableCOVID.js
```diff
@@ -1,4 +1,4 @@
-import React, { Component } from "react";
+import React, { useEffect, useState } from "react";
```
anniegiang, line 1: I believe the `/AirTableCOVID` page will be gone based on discussion on Slack. Were there changes?

client/src/pages/NrMap.js
```diff
@@ -33,61 +33,75 @@ const Cards = styled.div`
...
+  marker: null,
+  place: null,
+};
+
+const NrMap = () => {
+  const [coordinates, setCoordinates] = useState({
+    latitude: 52.520008,
+    longitude: 13.404954,
+  });
+  // rec-> list of hospitals according to nearbySearch
```
anniegiang, line 47: Instead of comments, what do you think about renaming `rec` and `plcDtl ` for better semantics?

client/src/pages/NrMap.js
```diff
@@ -33,61 +33,75 @@ const Cards = styled.div`
...
+  const [rec, setRec] = useState([]);
+  // plcDtl-> list of hospitals with requested details
+  const [plcDtl, setPlcDtl] = useState(() => []);
```
anniegiang, line 50: Why is the initial state a function?

client/src/pages/NrMap.js
```diff
@@ -33,61 +33,75 @@ const Cards = styled.div`
...
+  useEffect(() => {
     const googleMapScript = document.createElement("script");
     googleMapScript.src = url;
     window.document.body.appendChild(googleMapScript);
 
     googleMapScript.addEventListener("load", () => {
-      this.googleMap = this.createGoogleMap();
-      this.marker = this.createMarker();
-      this.place = this.places();
+      map.googleMap = createGoogleMap();
```
anniegiang, line 60: I'm not sure if it's good practice in React to update state like this, since `map` is just an object right now and not part of React state.

client/src/pages/NrMap.js
```diff
@@ -33,61 +33,75 @@ const Cards = styled.div`
...
+        setRec(results);
       },
     );
   };
 
-  placeDetails = () => {
-    this.state.rec.map((place) => {
+  useEffect(() => {
+    if (rec && rec.length > 0) {
+      placeDetails();
```
anniegiang, line 98: Does the order matter for which function is called first? In the old changes, it creates a marker before getting place details.

client/src/pages/NrMap.js
```diff
@@ -33,61 +33,75 @@ const Cards = styled.div`
...
+      createMarker();
+    }
+  }, [rec]);
+
+  const placeDetails = () => {
+    rec.map((place) => {
       let request = {
```
anniegiang, line 105: Let's always use `const` wherever we can.

client/src/pages/NrMap.js
```diff
@@ -99,81 +113,74 @@ class NrMap extends Component {
...
-      new window.google.maps.places.PlacesService(this.googleMap).getDetails(
+
+      new window.google.maps.places.PlacesService(map.googleMap).getDetails(
         request,
         (req, status) => {
           if (status == window.google.maps.places.PlacesServiceStatus.OK) {
-            this.setState((prevState) => ({
-              plcDtl: [...prevState.plcDtl, req],
-            }));
+            setPlcDtl((prevState) => {
```
anniegiang, line 121: If we change the initial state for `plcDtl ` to be `[]`, let's rewrite this to be `setPlcDtl([...plcDtl, req])`.

anniegiang:
@agmaal1 Thanks for working on this! I left some comments for your review. In addition, please the resolve merge conflicts.

Regarding the `OVER_QUERY_LIMIT ` response that Google Maps give back:

According to the docs, a solution to this is either to lower usage of the API, or increase usage by buying more credits. @manualzuru What are your thoughts?

client/src/pages/NrMap.js
agmaal1: `setPlcDtl([...plcDtl, req]) ` tried this way but only displays one card, it is not getting the previous updated state.

client/src/pages/AirTableCOVID.js
agmaal1: No idea. Should It remove it then?

client/src/pages/NrMap.js
agmaal1: done!

client/src/pages/NrMap.js
agmaal1: it doesn't matter. They are independent, one updates the google map and the other the component state.

client/src/pages/AirTableCOVID.js
anniegiang: Yea let's remove it. Currently the data is directly from Airtable, and we want to eventually use just our backend for serving data. We're in the midst of migrating data from Airtable to our database. Also, that data are Posts, which will be shown in the Feed page instead. So there's not much use for this page.

client/src/pages/NrMap.js
anniegiang: Ok I see. I would think that `p1cD1` would be equivalent to `prevState`. So it's weird that it's not getting the previous updated state.

client/src/pages/NrMap.js
agmaal1: remains from refactoring period. changed to useState([]);

client/src/pages/NrMap.js
agmaal1: I agree. I think that the React way would be by creating a ref with userRef() since in functional components, differently from class components, they are meant to store anything (not just dom nodes refs). Personally I still find it a bit confusing, maybe because I still thing in refs as a dom refs. Anyway I've have changed it, and removed the global map variable and it looks more React :D

client/src/pages/AirTableCOVID.js
agmaal1: removed

client/src/pages/NrMap.js
agmaal1: done

client/src/pages/NrMap.js
agmaal1: Yes, I also thought that way. React and goole libraries interacting with the Dom at the same time makes the behaviour of this component is a bit tricky.

client/src/pages/NrMap.js
agmaal1: Yes, I agree. I think that the react way is by useRef(). I've changed it to a ref. Also got rid of the other two variables since they are not storing anything (createMarker and placeDetails don't return anything).

client/src/pages/NrMap.js
agmaal1: done

client/src/pages/NrMap.js
robinv85: Good to know... I remember sometimes there can be problems integrating other libraries with React because of the lifecycle hooks, didn't know you could use refs for things other than DOM elements... I don't really find this in the docs though
https://reactjs.org/docs/integrating-with-other-libraries.html
https://reactjs.org/docs/refs-and-the-dom.html#when-to-use-refs

agmaal1: it seems that the perspective has changed because of hooks. it's in the useRef hooks docs https://reactjs.org/docs/hooks-reference.html#useref

client/src/pages/NrMap.js
```diff
@@ -33,62 +33,69 @@ const Cards = styled.div`
...
-      rec: [],
-      plcDtl: [],
-    };
-    this.googleMapRef = React.createRef();
-  }
-  componentDidMount() {
+const NrMap = () => {
+  const [coordinates, setCoordinates] = useState({
+    latitude: 52.520008,
+    longitude: 13.404954,
```
robinv85, line 39: where do these defaults come from?  Might be better to declare them as constants and put them at the top of the file or something like that, also easier to find back later when searching for something instead of using "magic" numbers or string values.

yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
robinv85, line 1: good point that you removed the `package-lock` from the root which didn't belong there
but this one we probably don't need it in the root either do we?

anniegiang:
Thanks for making the changes! Looks like there are some minor changes left but otherwise looks ready to merge.

yarn.lock
agmaal1: removed

client/src/pages/NrMap.js
agmaal1: I agree. moved to the top of the file as global variables

PR #247, opened 2020-05-05
Config absolute imports for CRA

### LINKED ISSUE #246: Setup absolute imports for CRA
- [x] Setup absolute imports with a baseUrl of `src` 
- [x] Update all relative paths in `client/src` to use absolute paths

```
before
import Button from "../../components/Button";

after
import Button from "components/Button"
```

Resource:
https://create-react-app.dev/docs/importing-a-component/#absolute-imports

anniegiang:
closes #246

If the config looks alright, I will move on to change all relative paths in `client/src` to absolute paths.

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,7 @@ while developing. This is a living document, so feel free to add any notes that 
     * Select an icon or component -> right click -> Copy/Paste -> Copy as CSS/SVG
     * You can also get exact CSS values on the top right of Figma
     * After you are done with your screen, lock it again!!
+* Use absolute imports. We have a resolved path of `client/src`. For example, import like this: `import Button from "components/Button"`, rather than `import Button from "../../components/Button`.
 * For most components, we can use components from Ant Design and override styling where needed. Please refrain from reinventing the wheel.
     ```
         import styled from "styled-components";
--- a/client/config-overrides.js
+++ b/client/config-overrides.js
@@ -1,6 +1,0 @@
-const { resolve } = require('path')
-
-module.exports = function override(config, env) {
-  config.resolve.alias['~'] = resolve(__dirname, 'src/');
-  return config;
-}
--- a/client/jsconfig.json
+++ b/client/jsconfig.json
@@ -1,0 +1,6 @@
+{
+  "compilerOptions": {
+    "baseUrl": "src"
+  },
+  "include": ["src"]
+}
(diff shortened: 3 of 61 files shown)


robinv85:
Yeah great idea, but there is already a webpack alias also that is set in `config-overrides.js` to `~` so we might want to stick to one method... this one is even shorter without the `~` prefix and it seems recommended by CRA docs so I guesss I'd prefer this, unless we have a good reason to use different aliasses...

anniegiang:
Yea I noticed that we were using the `~` alias but I think we should stick with the recommended setup by CRA. Maybe we could use `config-overrides.js` for different aliases outside of CRA?

anniegiang:
@robinv85 

Just changed most (if not all) relative imports in `client/src` to use absolute imports.

A possible concern is with Docker... I needed to rebuild Docker in order for the configuration for absolute imports to work. If this is merged, I'll update people on Slack about these changes and to rebuild if the config isn't working.

robinv85:
@anniegiang alright, sounds good, don't think we need `config-overrides` at the moment (YAGNI), we can just delete it, at least we should delete the other alias config.

anniegiang:
@robinv85 I tried removing `config-overrides.js` but got a `cannot find module` error from during the Github Actions build stage. The file isn't being imported anywhere and it looks like it's being included during the build process. @mannykary Do you know why this is happening?

Right now as a temporary fix to pass the building stage I added the file back as an empty file. 

I don't have a screenshot of the error message from Github Actions, but it was similar to this in the terminal:

Screen Shot 2020-05-06 at 9 23 40 AM

robinv85:
@anniegiang Ah yeah, it seems like someone added this dependency react-app-rewired https://github.com/timarney/react-app-rewired
I thought it was "ejected" to adjust configuration but using this plugin seems to be an alternative, maybe we can remove that plugin as well or just leave an empty config-overrides for now?

anniegiang:
> @anniegiang Ah yeah, it seems like someone added this dependency react-app-rewired https://github.com/timarney/react-app-rewired
> I thought it was "ejected" to adjust configuration but using this plugin seems to be an alternative, maybe we can remove that plugin as well or just leave an empty config-overrides for now?

Yea I agree, let's leave an empty config-overrides for now. I've never ejected before so maybe we could open an issue regarding this.

PR #249, opened 2020-05-05
Create reusable Button 

### LINKED ISSUE #248: Create a customizable Button
We currently have a `CustomButton` that needs to be refactored to handle different props more cleanly.

The CustomButton should ideally reflect the design theme and be reusable anywhere in order to achieve better style maintenance.

anniegiang:
closes #248 
- Added button themes in `theme.js`
- Refactored some components to use `CustomButton`
- Refactored `CustomButton` to accept the following props:
  - `primary`
  - `primaryLight`
  - `secondary`
  - `tertiary`
  - `inline` (ANTD prop)

Screen Shot 2020-05-04 at 11 11 08 PM

Next steps:

Currently many components uses `SubmitButton`, which is very similar to `CustomButton` in terms of styling. However `SubmitButton` returns a React component instead of a styled-component. 

This isn't ideal because it dynamically creates a `styled-component` within the React component, which can lead to memory leaks in the long run:
Screen Shot 2020-05-04 at 9 41 17 PM

Once this PR is merged, I plan to replace `SubmitButton` with `CustomButton`.


anniegiang:
@robinv85 Yea good point about rarer buttons that are invisible and have icons. I think we could separate styles specifically for those buttons to not bloat and overcomplicate `CustomButton`. Maybe I should rename it to be `LargeButton` to be more specific.

Issue #178, opened 2020-04-26
path from need help / offer help w/o symptoms doesn’t work


dbenyamin:
Choosing either path (need help / offer help) doesnt take you to the feed, per spec:

https://www.notion.so/fightpandemics/Feed-653658508c4649869787c48f3e4eee13#f6992b37a3b1406292c7f54c4cdb33d7

It instead takes you to the emergency contact page. This page should only be shown if one chooses Need Help -> Have Symptoms

PR #255, opened 2020-05-05
Fixed path from Offerhelp issue #178

Khoi8:

Fixes issue #178 where the path from after submitting your email in offer help does not take you to the Feeds page like it should.

* Changes the OfferHelp.js file.
* The path in line 194 was originally set to /medical but I changed it to /Feed so that the next page will be Feed.js.


anniegiang:
@nadinejuraschek @Khoi8 

Right, it should be `/feed` 

from `client/src/routes.js`

Screen Shot 2020-05-05 at 6 01 04 PM

Khoi8:
Oh yes it should be lower case my bad. I was thinking of the file name instead. Changing now.

Issue #213, opened 2020-04-30
Configure an ESLint rule for prefer export default


anniegiang:
When there is only a single export from a module, prefer using default export over named export.

```
// bad.js

// There is only a single module export and it's a named export.
export const foo = 'foo';

```

```
// good1.js


export const foo = 'foo';

const bar = 'bar';

export default 'bar'; // There is a default export.

```

For your reference or inspiration:
https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/prefer-default-export.md

manualzuru:
@robinv85 should we move this to closed?

robinv85:
@manualzuru it's not fixed yet, we have ESLint configuration but if we change the rules then we have to update all the code so this is probably not a good time right now when a lot of development is happening, it will cause a lot of merge conflicts

joshmorel:
I'm going to close this. It's kind of stale & we need to fix #786 first putting effort into adding more rules.

PR #258, opened 2020-05-06
ESLint alphabetize

TenNga:
### PR Checklist:

Hey there! It's great to have you here. Here's a quick checklist to help you make a shiny PR:

* [x] Have you checked out the guidelines in our [Contributing](../CONTRIBUTING.md) document?
* [x] Have you looked if there might be other open [Pull Requests](../../../pulls) for the same update/change?

### Quick Description of the PR

Solve issue https://github.com/FightPandemics/FightPandemics/issues/199

### An Overview of the Changes

*  Alphabetize imports and most of export

PR #268, opened 2020-05-07
add eslint config to prefer default export

aelnahas:
### PR Checklist:

* [ ] Have you checked out the guidelines in our [Contributing](../CONTRIBUTING.md) document?
* [ ] Have you looked if there might be other open [Pull Requests](../../../pulls) for the same update/change?

- Reference ticket: https://github.com/FightPandemics/FightPandemics/issues/213
- Added eslint config to prefer default exports
- Fixed some errors that eslint caught.

- enforces the linting rule, along the way there were small errors, such as using un imported react hooks .

* To help understand them.

### DIFF
--- a/client/.eslintrc
+++ b/client/.eslintrc
@@ -1,7 +1,0 @@
-{
-  "extends": ["react-app"],
-  "plugins": ["import"],
-  "rules": {
-    "import/prefer-default-export": 1
-  }
-}
--- a/client/.eslintrc
+++ b/client/.eslintrc
@@ -1,0 +1,7 @@
+{
+  "extends": ["react-app"],
+  "plugins": ["import"],
+  "rules": {
+    "import/prefer-default-export": 1
+  }
+}
--- a/client/package.json
+++ b/client/package.json
@@ -30,9 +30,9 @@
     "typeface-work-sans": "^0.0.72"
   },
   "scripts": {
-    "start": "react-app-rewired start",
-    "build": "react-app-rewired build",
-    "test": "react-app-rewired test",
+    "start": "react-scripts start",
+    "build": "react-scripts build",
+    "test": "react-scripts test",
     "eject": "react-scripts eject",
     "lint": "eslint src",
     "install-docker": "docker-compose run client npm install",
--- a/client/package.json
+++ b/client/package.json
@@ -30,9 +30,10 @@
     "typeface-work-sans": "^0.0.72"
   },
   "scripts": {
-    "start": "react-scripts start",
-    "build": "react-scripts build",
-    "test": "react-scripts test",
+    "start": "react-app-rewired start",
+    "build": "react-app-rewired build",
+    "test": "react-app-rewired test",
+    "lint": "eslint . --fix",
     "eject": "react-scripts eject",
     "lint": "eslint src",
     "install-docker": "docker-compose run client npm install",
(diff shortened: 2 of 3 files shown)


client/src/pages/NrMap.js
```diff
@@ -1,4 +1,4 @@
-import React, { useState, useEffect, useRef } from "react";
+import React, { useState, useEffect, useRef, useCallback } from "react";
```
anniegiang, line 1: This seems out of scope for this PR ? 

@agmaal1 was working on this - can you double check if this is fine please?

aelnahas: @anniegiang I can revert it, only reason why I did that is when I ran the linter this came up as an error, `useCallback` is being used without an import, same with with render on the functional component

anniegiang: Ah ok, maybe we would fix that error in other PR.

client/.eslintrc
```diff
@@ -0,0 +1,7 @@
+{
```
anniegiang, line 1: This [PR ](https://github.com/FightPandemics/FightPandemics/pull/258) configures ESlint in `client/.eslintrc.json`. We also have `backend/.eslintrc.json`. I think to stay consistent, maybe we could wait for that PR to merge, and then add these rules on top of that. What do you think?

robinv85: Similar to my comment on the other ESLint changes for the frontend, shouldn't we use the create-react-app way to change ESLint config? https://github.com/FightPandemics/FightPandemics/pull/258#discussion_r422545352
> CRA isn't fully configurable unless "ejected", you can read more about this here: https://create-react-app.dev/docs/available-scripts#npm-run-eject but we probably want to avoid that unless absolutely necessary because it also has disadvantages.
> 
> For ESLint there is a setting that can be changed and then we can add some additional ESLint settings, that will be added to the settings that are provided by CRA, see https://create-react-app.dev/docs/advanced-configuration
> 
> So I think that's the way to go, and to stay within the scope of this PR you should probably only add what's needed, which is the `"sort-imports"` rule and `"sort-keys"`

Now I'm not sure how that works with CRA but indeed seems likely that it still needs to extend the `react-app` config.

Also didn't know that this needs the `import` plugin, or it might provide some additional benefits?

aelnahas: @robinv85  I wasn't aware of that, I guess I could wait until the other PR is merged and then just apply this rule to it , and yea according to the guide was posted , it looked like that plugin was needed , cause the rule is part of that package.

aelnahas:
@robinv85 I rebased from master and made the changes . Please take a look

robinv85:
Before this is updated, I suggest we first update the config, I think this is probably best way to go:
- move Airtable keys to `config.js` (probably these will be removed eventually anyway
- add `client/.env` to git (remove from .gitignore)
- make sure the `EXTEND_ESLINT=true` setting is set for everyone

client/package.json
```diff
@@ -56,6 +56,8 @@
     "@testing-library/react": "^10.0.3",
     "@testing-library/user-event": "^10.1.0",
     "dotenv": "^8.2.0",
+    "eslint": "^6.8.0",
```
robinv85, line 59: isn't this already included in create-react-app? Or does it need to be here if you also add a plugin?

manualzuru:
@BMariscal @robinv85 @anniegiang @aelnahas should we merge this?

client/package.json
robinv85: Following the other ESLint-related issues, I don't think this should be added in the package.json or it might have conflicting version with the one that is already included and it might cause problems

idjevm:
Closing this, for now, feel free to reopen one this will be decided on. Also, the issue is still open so we can reopen this when the time comes.

PR #284, opened 2020-05-10
Data model migration v2

### LINKED ISSUE #146: Add Expiration Date to Posts
We need to add `expireAt` to the Post model and the Post endpoints.

In a future ticket, we will add filtering on the backend to filter out expired posts before serving them to the FE.

When editing the post, the `expireAt` should be referenced from the time the person is editing the post, not the original time of posting.

### LINKED ISSUE #204: Change Post data model to v2
Change Post mongoose model to [v2 data model version](https://github.com/FightPandemics/DataModel) and adapt the posts endpoints to reflect those changes.

Naraujo13:
### PR Checklist:

* [X] Have you checked out the guidelines in our [Contributing](../CONTRIBUTING.md) document?

Update Post data model to v2, while adjusting post endpoints.
Add geolocation sorting by proximity to `/posts` with expiration and visibility filters.
Add capping of post size based on whether the user is authenticated or not.

Resolves #146 
Resolves #204 

Refactor Endpoints:
- GET /posts
- POST /posts
- GET /posts/:postId
- PATCH /posts/:postId
- DELETE /posts/:postId
- PUT /posts/:postId/:userId (like)
- DELETE /posts/:postId/:userId (unlike)

Add to `/posts`
- Geolocation sorting based on distance to user
- Filter for expiration date - fetching only posts not expired
- Filter for visibility - fetching only posts with a visibility corresponding to the current user location
- Cap post description at 100 words if the user is not authenticated


backend/lib/endpoints/schema/posts.js
```diff
@@ -1,53 +1,94 @@
...
+const { join } = require("path");
+const PATH_TO_SCHEMAS = join(__dirname, "../../models/schemas/v2");
+const { schema: postSchema } = require(join(PATH_TO_SCHEMAS, "post"));
+const { schema: orgSchema } = require(join(PATH_TO_SCHEMAS, "individualUser"));
+const { schema: userSchema } = require(join(
+  PATH_TO_SCHEMAS,
+  "organizationUser",
+));
+
+const EXPIRATION_OPTIONS = ["day", "week", "month", "forever"];
```
anniegiang, line 13: Should we create something like a json file to store constants and enum data? I'm thinking this may help collaborate with the frontend. What do you think?

backend/lib/endpoints/schema/posts.js
```diff
@@ -1,53 +1,94 @@
...
+    .prop(
+      "externalLinks",
+      S.object()
+        .prop("appStore", S.string().format("url"))
+        .prop("email", S.string().format("email"))
+        .prop("playStore", S.string().format("url"))
+        .prop("website", S.string().format("url")),
+    )
+    .prop("language", S.array().items(S.string()))
+    .prop("objective", S.string().enum(POST_OBJECTIVES).required().required())
```
anniegiang, line 50: I think one `.required()` is enough?

backend/lib/models/schemas/v2/post.js
```diff
@@ -0,0 +1,169 @@
...
+    expireAt: Date,
+    externalLinks: {
+      appStore: String,
+      email: String,
+      playStore: String,
+      website: String,
+    },
+    language: [String],
+    likes: {
+      // TODO: how to guarantee unique ids?
```
anniegiang, line 23: Maybe we can use `enum` to check if ids exist already or not?

backend/lib/endpoints/schema/posts.js
Naraujo13: Sure, only once is enough, just a typo 🤣

backend/lib/endpoints/schema/posts.js
Naraujo13: Well, I agree that we should have only one source for them. 

Initially, I was thinking of using it at the model, accessing it through the schema as I did with the visibility options and etc, and then adding there would update everything. But that covers only what we store at the database, cases like this wouldn't be covered (because the interface options are periods and we store the expire date). 

How do you think we can mix those things?
Maybe the JSON defines all options (e.g. `posts.json` would cover DB and interface fields for posts) and at the models, we would use that to define the field enums?

backend/lib/endpoints/schema/posts.js
anniegiang: LOL I was re-reading it over and over again wondering why there were 2 of them 😅

backend/lib/models/schemas/v2/post.js
Naraujo13: Actually, I've used `addToSet` and `removeToSet` mongo operations at the endpoints for liking, so that will guarantee unicity at DB level, though not validate at the model. With that, if the ID is already there, it won't return an error or a message, and neither will it add the like again.

How could we use `enum` to validate at the model level?

backend/lib/endpoints/schema/posts.js
anniegiang: Yup I agree - having a JSON that defines all the options that are available for the models and frontend. Then we can import them to models and frontend components wherever needed. Does that work or does it sound to simple to work?

backend/lib/models/schemas/v2/post.js
anniegiang: Yea adding a validation at model-level sounds like a good idea for error handling and potentially catching ids that pass the model validation but are not unique.

Now that I think of it `enum` won't work because I believe you need to define the values first, which doesn't sound right for likes. 

Instead, could we use unique indexes? If I remember how it works, we could add an index on both the `userId` and `postId` to prevent duplicate insertions.

backend/lib/endpoints/schema/posts.js
robinv85: Yeah great idea to have a shared file, this is often done with JSON schema's, sometimes for both API validation, database schema's and to share it with the frontend. This is also similar to GraphQL types... but in this case I think it makes sense to have a file that the frontend could use as well, instead of mixing DB and API fields into one file and then use that at the model?

backend/lib/models/schemas/v2/post.js
robinv85: @Naraujo13 sounds good, first I used that `PUT likes/:id` in the API because I found it somewhere and thought it makes sense in a RESTful way, but I think what I did wat return a 400 bad request if the like already existed, which isn't completely valid in REST I think since `PUT` is idempotent so it always has the same result, whether or not the like already existed, after the request it's certainly `true`. 
For DELETE this is similar, after the DELETE the resources is certainly deleted, but I think it makes sense to return a `404` error as well if it was already deleted at the time of the request.

robinv85: @anniegiang I don't think this is needed per se to add at model level as unique index because this is not a separate collection I think, it's just an array inside a document... and the mongo query as I implemented it before would just not add duplicates so that should be ok... not sure if that changed now?

backend/lib/endpoints/schema/posts.js
Naraujo13: Well, we have a J[SON Schema already for the DB fields](https://github.com/FightPandemics/DataModel/blob/master/v2/scripts/JSON_Schema/applicationDB/posts.json) since Mongo uses that. 

But how would accessing the enums through the JSOM schema be any different than accessing it through the schema?

backend/lib/models/schemas/v2/post.js
Naraujo13: It is possible to create a unique index, but that will guarantee only at db level, the same way we are already doing with `addToSet` for example.

@robinv85 currently both `PUT` and `DELETE` are idempotent, though no error if the like is already there or if it has already been deleted.

robinv85: Yeah they both should be idempotent but I'm not sure if the response has to follow a certain rule, we might be open to decide this for our own, but it would be nice to have a consistent approach accros all API handlers. 
Also for the client, I don't think this makes a big difference, if we always return OK then the client will just update the state according to the request too. Iimagine 2 people click delete at same time, both would get "ok" and no error message, and client would just remove the post, so seems good to me.

backend/lib/endpoints/schema/posts.js
robinv85: Well I think the main advantage here can be DRY, avoiding duplication, by having a single source of truth for the values that can be used for certain fields.
So for that matter we could consider to move these to a shared file.
The question is, do we share them with mongo schema & API schema, and also with frontend?

I think for us Mongo schema's and fluent schema's are always made manually, but either you could export things from the model files separately and import it in the API schema's (just JS literal for example for the enums), or a separate file for example in `common/schema` could also export named enums etc.
 
Sharing with the frontend is a little bit more complex

backend/lib/endpoints/posts.js
```diff
@@ -21,135 +22,355 @@ async function routes(app) {
...
+      // Additional filters
+      const { paramFilters } = req.params;
+      if (paramFilters) {
+        // TODO: additional filters
+      }
+
+      // Unlogged user limitation for post content size
+      // TODO: how to check for logged/unlogged user?
+      /* eslint-disable sort-keys */
+      const contentProjection = user
```
mannykary, line 79: I recall a discussion we had in the engineering meeting last weekend where we were going to truncate the post on the frontend rather than the backend, are we not doing that anymore?

mannykary: Also, if we have the authenticate hook on this endpoint, then I don't think unauthenticated users would be able to get this this point since the authenticate hook will throw an error.

Naraujo13: I was also not sure if that was to be done in the backend or frontend. Talking with @anniegiang I decided to leave it done if we decided to go with backend strategy since this endpoint has different behaviors based on whether the user is authenticated or not. 

Anyway, if we decided to go the fronted route we can still just remove this project from the pipeline and the auth hook here.

robinv85: @manualzuru suggested doing this on frontend, I think for the feed for MVP this is fine, also yes I think we already discussed this a few times and it was clear that the auth is not needed for reading posts.

indirabhowmick:
@Naraujo13 - can we please move the ticket to Done so that PM's can test the feature?

PR #300, opened 2020-05-11
Landing Page layout fixes

### LINKED ISSUE #288: fix layout on landing view
flow
- go to https://staging.fightpandemics.work/

layout issues

1 - logo is not aligned on the left with the title (may be a navbar problem)
2 - spacing below logo 
2 - missing title
3 - spacing between title and subtitle
4 - spacing on lines on subtitle
5 - no rounded borders in the blue box
6 - no left margin
7 - incorrect font styles used in the blue box

Screenshot 2020-05-11 at 03 26 37

Styles should be pixel perfect:
Screen Shot 2020-05-10 at 11 14 36 PM

iMac, Chrome

Reference:
https://www.figma.com/file/4cV84H1jyMo0rJEAl3HNnJ/FightPandemics-landing-page?node-id=1377%3A8027

nadinejuraschek:
closes #288 

* Tried to match the style more closely to Figma. The margin and font-sizes do not match Figma exactly because there were different font-sizes set for different Desktop sizes. I tried to make it look similar.

* Fixed colors - some font colors were set to "black", I changed it to darkerGray from our themes.js since that is the darkest color used for text in Figma.

* Fixed min-height on main for Desktop views.

* Navbar height needs to be adjusted to match Figma for Desktop. I didn't change the margin / padding for the Navbar, since I think that should be another issue / ticket.


client/src/pages/Home.js
```diff
@@ -21,18 +21,19 @@ const StyledIntro = styled.div`
   }
 
   @media screen and (min-width: ${mq.tablet.wide.minWidth}) {
-    background-color: ${theme.colors.royalBlue};
-    border-radius: 1rem;
+    background: linear-gradient(337.81deg, #425af2 3.41%, #677cf2 98.66%);
```
anniegiang, line 24: Awesome, didn't realize it was a gradient. Maybe we should add this to `theme.js` and then access it from there?

client/src/pages/Home.js
```diff
@@ -21,18 +21,19 @@ const StyledIntro = styled.div`
   }
 
   @media screen and (min-width: ${mq.tablet.wide.minWidth}) {
```
anniegiang, line 23: I'm not sure which media query it is but the page looks kinda off between 450px and 1024px. I think the desktop version shows until 450px? Right now it stops showing <1024px


Screen Shot 2020-05-10 at 11 52 37 PM

client/src/pages/Home.js
```diff
@@ -102,18 +103,21 @@ const StyledP = styled.p`
     font-family: ${theme.typography.font.family.display}, sans-serif;
```
anniegiang, line 103: I think the code is here but the font styles are different now on mobile

Screen Shot 2020-05-10 at 11 53 43 PM

client/src/pages/Home.js
```diff
@@ -137,6 +141,7 @@ const Home = (props) => {
     <MainContainer className="text-center">
       <StyledIntro>
         <IntroText>
+          <StyledP>FightPandemics</StyledP>
```
anniegiang, line 144: I think the design team just updated the designs ? This doesn't show on Figma

Screen Shot 2020-05-10 at 11 57 37 PM

client/src/templates/layouts/Main.js
```diff
@@ -1,8 +1,14 @@
 import styled from "styled-components";
 
+import { mq } from "constants/theme";
+
 const Main = styled.main`
   margin-left: 25px;
   margin-right: 25px;
+
+  @media only screen and (min-width: ${mq.tablet.wide.minWidth}) {
```
anniegiang, line 9: Not sure if this is good since it'll affect all components inside of Main.. If there's no workaround I guess we could go through and update components that were affected

client/src/pages/Home.js
nadinejuraschek: I'll check the new designs, they are on Figma now

anniegiang:
Quick update - #332 is doing some similar with the media queries for the Landing page.

client/src/pages/Home.js
nadinejuraschek: I'm a bit confused... There are two styles for the landing page on Figma, one says "FightPandemics" above the "A place..." and the other doesnt

client/src/templates/layouts/Main.js
nadinejuraschek: But don't all components inside of main have to have a min-width of 100vh- (footer + navbar) to ensure that the footer is at the bottom if the content is short?

client/src/pages/Home.js
nadinejuraschek: I haven't been able to figure out what's going on here yet...

client/src/templates/layouts/Main.js
anniegiang: Ah sorry I commented this while on mobile and made a mistake, I meant to write this comment for line 10

anniegiang: I meant that adding `margin: 0` may affect other components that are within the `Main` component

client/src/pages/Home.js
anniegiang: Yea, this is also happening in #322. I think it has to do with a media query controlling the blue box. It looks like the media query is hiding the box at 1024px (too early) instead of 450px

client/src/pages/Home.js
anniegiang: Yikes that's confusing... let's clarify with the design team

client/src/pages/Home.js
nadinejuraschek: Someone just opened an issue for that. I'll leave fixing this problem to the other ticket.

client/src/pages/Home.js
nadinejuraschek: This seems to be fixed already

client/src/templates/layouts/Main.js
nadinejuraschek: I don't think there is another way to remove the margin on the landing page? Since the Main component will always set a 25px if we don't remove it and our content is rendered in Main for every page

nadinejuraschek: You're right though, that might effect other pages as well, I just don't know how to fix it without changing the margin here

TenNga:
Fixed this issue here https://github.com/FightPandemics/FightPandemics/issues/343

client/src/pages/Home.js
nadinejuraschek: I reached out on Slack

nadinejuraschek:
Great! I haven't touched the Navbar, so there shouldn't be an issue with #332

anniegiang:
duplicate of #349

manualzuru:
@nadinejuraschek @anniegiang @TenNga should we move this from in progress?

PR #311, opened 2020-05-12
fix the cursor and delete comment

harithzamri:
Close the comment and add pointer


anniegiang:
@harithzamri Please make the changes in the PR that was opened already.

PR #312, opened 2020-05-12
feat: refactor the mongo model & schema files & loading

robinv85:
Moved mongo schema & models together into files so it's also easier to customize models when needed
removed the magical auto loading based on file names
export both schema & model from the file just in case both are needed
require all the models in the mongo plugin to load them
( this changes slightly implementation with global mongoose models instead of on the specific mongo connection ... but for now this doesn't make a difference since we only have one connection).


robinv85:
Yeah @anniegiang I saw that, @Naraujo13 also mentioned it, to me it's the same whether we only export the model  for example `User` and then we can still use `User.schema` 
Only annoyance is that we are still using the old ES5 / Common.js syntax which is `module.exports` and that one doesn't really support both default + named exports from same file in a nice way. So I changed it to 2 named exports... but I will check maybe it's better to use only export like
``` module.exports = model("User", UserSchema);` and then use `User.schema`

robinv85:
I've updated this to use default exports and use `Model.schema` if we need it so should be good and usable for @Naraujo13 in #284

Issue #278, opened 2020-05-10
fix refresh on link


stavares843:
flow
- go to https://staging.fightpandemics.work/
- click on need help
- refresh
- returns 404

<img width="1863" alt="Screenshot 2020-05-10 at 21 59 13" src="https://user-images.githubusercontent.com/29093946/81510427-87c45080-9309-11ea-8f80-3486ec977141.png">

```
Request URL: https://staging.fightpandemics.work/need-help
Request Method: GET
Status Code: 404
```

PR #317, opened 2020-05-12
Connect client to backend through NGINX; expose Swagger API docs

### LINKED ISSUE #337: Expose Swagger API docs
We should be able to access the docs at `/api/documentation`

mannykary:
### PR Checklist:

This hooks up the client container to the backend container in Fargate. I had to add an `nginx.conf` in the client container, and proxy requests to `/api` to the backend container.

I also exposed the Fastify Swagger docs at `/api/documentation`.

* Add `nginx.conf` file
* Set Swagger doc route to `/api/documentation`
* Geolocation endpoint fixes - change to get, add default location if not set, remove authentication in the backend

Closes #337, #338, and #278

### DIFF
--- a/backend/lib/endpoints/geo.js
+++ b/backend/lib/endpoints/geo.js
@@ -7,7 +7,7 @@ const { getCountrySchema } = require("./schema/geo");
 async function routes(app) {
   app.get(
     "/country",
-    { preValidation: [app.authenticate], schema: getCountrySchema },
+    { schema: getCountrySchema },
     async (req) => {
       const { latitude, longitude } = req.query;
       return findByLatLong(latitude, longitude);
--- a/client/nginx.conf
+++ b/client/nginx.conf
@@ -18,7 +18,7 @@ http {
 
     location /api {
       # Reject requests with unsupported HTTP method
-      if ($request_method !~ ^(GET|POST|HEAD|OPTIONS|PUT|DELETE)$) {
+      if ($request_method !~ ^(GET|POST|HEAD|OPTIONS|PUT|PATCH|DELETE)$) {
         return 405;
       }
 
--- a/client/nginx.conf
+++ b/client/nginx.conf
@@ -13,8 +13,9 @@ http {
 
   server {
     listen 80;
+    root /usr/share/nginx/html;
+    include /etc/nginx/mime.types;
 
-    # Nginx will reject anything not matching /api
     location /api {
       # Reject requests with unsupported HTTP method
       if ($request_method !~ ^(GET|POST|HEAD|OPTIONS|PUT|DELETE)$) {
@@ -23,7 +24,7 @@ http {
 
       # Only requests matching the whitelist expectations will
       # get sent to the application server
-      proxy_pass http://localhost:3000;
+      proxy_pass http://127.0.0.1:8000;
       proxy_http_version 1.1;
       proxy_set_header Upgrade $http_upgrade;
       proxy_set_header Connection 'upgrade';
@@ -32,4 +33,5 @@ http {
       proxy_cache_bypass $http_upgrade;
     }
   }
+
 }
--- a/client/nginx.conf
+++ b/client/nginx.conf
@@ -20,6 +20,10 @@ http {
       try_files $uri $uri/ /index.html;
       include /etc/nginx/mime.types;
     }
+    error_page 500 502 503 504 /50x.html;
+    location = /50x.html {
+      root  /usr/share/nginx/html;
+    }
 
     location /api {
       # Reject requests with unsupported HTTP method
@@ -38,5 +42,4 @@ http {
       proxy_cache_bypass $http_upgrade;
     }
   }
-
 }
--- a/client/nginx.conf
+++ b/client/nginx.conf
@@ -13,8 +13,13 @@ http {
 
   server {
     listen 80;
-    root /usr/share/nginx/html;
-    include /etc/nginx/mime.types;
+
+    location / {
+      root /usr/share/nginx/html;
+      index index.html index.html;
+      try_files $uri $uri/ /index.html;
+      include /etc/nginx/mime.types;
+    }
 
     location /api {
       # Reject requests with unsupported HTTP method
--- a/client/nginx.conf
+++ b/client/nginx.conf
@@ -1,0 +1,35 @@
+# TODO move to separate sidecar container
+
+events {
+  worker_connections 768;
+}
+
+http {
+  # Nginx will handle gzip compression of responses from the app server
+  gzip on;
+  gzip_proxied any;
+  gzip_types text/plain application/json;
+  gzip_min_length 1000;
+
+  server {
+    listen 80;
+
+    # Nginx will reject anything not matching /api
+    location /api {
+      # Reject requests with unsupported HTTP method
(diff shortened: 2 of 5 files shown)


mannykary:
No idea why the bot didn't post the build URL here, but here it is: https://featurefix-backend-dockerfile.fightpandemics.xyz

Swagger docs: https://featurefix-backend-dockerfile.fightpandemics.xyz/api/documentation

hkellaway:
while i can’t review at a code level - i did take a look at the sample of what this produces, and that’s awesome!

robinv85:
Hey @mannykary thanks for helping out, seems like the API docs are working but we just need to add Nginx config to redirect the requests to React for client-side routing so we can refresh pages and share links, see for example https://featurefix-backend-dockerfile.fightpandemics.xyz/need-help
But we could do that later in another PR

backend/lib/endpoints/geo.js
```diff
@@ -7,7 +7,7 @@ const { getCountrySchema } = require("./schema/geo");
 async function routes(app) {
   app.get(
     "/country",
-    { preValidation: [app.authenticate], schema: getCountrySchema },
+    { schema: getCountrySchema },
```
robinv85, line 10: why remove this auth check? well not sure how this API call is used at the moment, we might need to add some additional security checks later

client/nginx.conf
```diff
@@ -0,0 +1,37 @@
...
+  gzip_min_length 1000;
+
+  server {
+    listen 80;
+    root /usr/share/nginx/html;
+    include /etc/nginx/mime.types;
+
+    location /api {
+      # Reject requests with unsupported HTTP method
+      if ($request_method !~ ^(GET|POST|HEAD|OPTIONS|PUT|DELETE)$) {
```
robinv85, line 21: I think `PATCH` should also be supported with our current API, not sure if that will still be kept with data model v2 updates @Naraujo13 did you also consider PUT vs PATCH for all the data model updates in the API?

client/src/pages/Medical.js
```diff
@@ -67,9 +67,15 @@ const Medical = (props) => {
   };
 
   const fetchGeoData = () => {
-    const geolocation = getGeoLocation();
+    // Defaulting to San Francisco
+    // TODO fix geolocation
```
robinv85, line 71: ah ok so I guess you're just testing here if it works with the geo API, doesn't it work anymore with HTML5 geo coords?

mannykary: Yeah so I kept getting an empty geolocation, so I defaulted it to SF lat/long to test that the call to the backend works. I also removed the authentication check on the geo endpoint for this reason, and I don't believe that it should be there since unauthenticated users can get to this page.

client/nginx.conf
mannykary: Good call, I just copied/pasted this from an AWS sample code. I can add `PATCH` here

backend/lib/endpoints/geo.js
mannykary: The only place I see this being called is when going through the "Need Help" flow to get to the page where it provides an emergency number, so I'm not sure if authentication is necessary to simply get the geolocation from a set of lat/long coordinates.

mannykary:
@robinv85 I can probably address that routing issue in this PR. I think we would need to do something like this: https://levelup.gitconnected.com/dockerizing-a-react-application-using-nginx-and-react-router-43154cc8e58c

```
server {
  listen 80;
  location / {
    root   /usr/share/nginx/html;
    index  index.html index.htm;
    try_files $uri $uri/ /index.html;
  }
  ...
```

client/src/pages/Medical.js
mannykary: I guess I should really try using a different location than the city I'm in just in case it was just a fluke that it didn't work... 🤔

backend/lib/endpoints/geo.js
robinv85: yeah you're right I think for now this is mostly used by unauthenticated users
we could replace it by external services later and we can also consider it for the create profile page and similar to enter the location data, might only get geo coordinates and automatically fill some other fields to make sure it's correctly set

client/nginx.conf
robinv85: yeah I guess `PATCH` is a lesser used HTTP method

client/src/pages/Medical.js
robinv85: depends on the type of error response? was the API just completley unreachable or did you get an error response because of the input?

robinv85:
@mannykary ok yeah sure if you think you can fix that at once that's great! 
I wonder if we also need additional config for static files? (probably /public folder) even though it seems there isn't much in there at the moment

client/src/pages/Medical.js
mannykary: The api call to our geoservice endpoint failed when the geolocation object was empty, since passing in the latitude and  longitude query parameters was required. I don't know why the geolocation object was empty. So I think the HTML5 geolocation failed, so we should fix that in another ticket.

mannykary: The api call to our geoservice endpoint failed when the geolocation object was empty, since passing in the latitude and  longitude query parameters was required. I don't know why the geolocation object was empty. So I think the HTML5 geolocation failed, so we should fix that in another ticket.

robinv85: Ah ok, first version that I made it was working so probably regression, I think it was only making the request when the user had given permission to get the location with html5 geo api

mannykary:
@robinv85 @anniegiang @hkellaway Re-requesting review as I made a few more changes since you all approved:
* Add `PATCH` as permitted HTTP verb
* Handle React routing so navigating directly to React route, or refreshing the page works as expected: https://featurefix-backend-dockerfile.fightpandemics.xyz/need-help
* Use Alpine version of NGINX base Docker image to slim down the container size
* Show nicer 50x pages. We might want to customize the page though since it's still a rather generic nginx error page, but that can be part of another ticket.

Also @robinv85, regarding your comment about adding additional config for static files in the `/public` folder, looks like this isn't necessary: https://featurefix-backend-dockerfile.fightpandemics.xyz/favicon.ico. Looks like everything in the `/public` folder ends up at the root.

client/nginx.conf
```diff
@@ -0,0 +1,45 @@
...
+  gzip_proxied any;
+  gzip_types text/plain application/json;
+  gzip_min_length 1000;
+
+  server {
+    listen 80;
+
+    location / {
+      root /usr/share/nginx/html;
+      index index.html index.html;
```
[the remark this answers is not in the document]
mannykary: Haha, I don't know how this happened. I'm pretty sure it used to be `index.htm` for the second one. In any case, I'll just remove the second one.

robinv85:
Yeah @mannykary btw about the static assets, seems like most of them are places inside  `/src/assets` and imported in the jsx files so probably handles by Webpack loaders, we might look into that later to optimize performance.

client/nginx.conf
Naraujo13: > I think `PATCH` should also be supported with our current API, not sure if that will still be kept with data model v2 updates @Naraujo13 did you also consider PUT vs PATCH for all the data model updates in the API?

Yeah, the API is using `PATCH` at the API updates. It was already using `PATCH` before and I kept that while updating the data models.

PR #329, opened 2020-05-12
added email-validation to step 3 on NeedsHelp page

omarzoghayyer:

Added Email Validation file which extends from common.js. This will allow to fire toast notifications once clicked on submit with empty placeholder. 

Added Toast extent-ion which extends from antd-mobile.

### DIFF
--- a/client/src/pages/NeedHelp.js
+++ b/client/src/pages/NeedHelp.js
@@ -1,6 +1,7 @@
 import React, { useState } from "react";
 import { withRouter } from "react-router-dom";
-
+import {Toast } from "antd-mobile";
+import { validateEmail } from "../utils/common.js";
 import SubmitButton from "components/Button/SubmitButton";
 import { asyncGetGeoLocation } from "utils/geolocation";
 import {
@@ -72,14 +73,18 @@ const Step2 = (props) => {
       </AnswerButton>
     </WizardStep>
   );
-};
-
+}; 
 const Step3 = (props) => {
   const [email, setEmail] = useState("");
   const onChange = (evt) => setEmail(evt);
-  const onSubmit = () => {
+  const onSubmit = (evt) => {
+    evt.preventDefault();
+    if (!validateEmail(email)) {
+      Toast.fail("Invalid email address!", 2.0);
+      return;
+    }
     props.update("email", email);
-  };
+  }
   return (
     <WizardStep>
       <WizardProgress className="text-primary">


robinv85:
Hey @omarzoghayyer can you replace the "placeholder text" of the PR template with your own text? 
It's a bit confusing if we have to read the whole default template before we see your own comment to understand what you have done...

client/src/pages/NeedHelp.js
```diff
@@ -72,14 +73,18 @@ const Step2 = (props) => {
...
-
+}; 
 const Step3 = (props) => {
   const [email, setEmail] = useState("");
   const onChange = (evt) => setEmail(evt);
-  const onSubmit = () => {
+  const onSubmit = (evt) => {
+    evt.preventDefault();
+    if (!validateEmail(email)) {
+      Toast.fail("Invalid email address!", 2.0);
```
robinv85, line 83: I will double check with designers but I think my idea was to have client-side validation handled by form components, like adding an error state
And network requests errors should come back with a pop-up... for example "user already exists" cannot be hanvled by the form itself.

robinv85:
Hey @omarzoghayyer so did you base the decision to use a toaster component on a specific screen in the designs?

I don't think there are designs for error states for all the forms and inputs, so maybe we should add some more documentation or update all the requirements, but I think it would be better not to use the toaster for this. See for example this error state in the MVP design system https://www.figma.com/file/4cV84H1jyMo0rJEAl3HNnJ/FightPandemics-landing-page?node-id=3516%3A22832
There you see error state for "email" input field, with red line and red text...

I would still prefer the toaster to use for other errors like coming from the backend, but I need to verify with UX designers if that's ok, for example "this account already exists" on the signup page is easier to implement for our MVP if we just show a toaster from the Redux action creator on failure, otherwise we need 3 redux actions and update the form state through redux if we want to show it also inline...

omarzoghayyer:
@robinv85 I initially didn't know how to implement it but then I saw it implemented elsewhere in the program, so I decided to follow the same steps. But I understand, it's better to approach it from different design so that we know if the user is already signed up. For that Im not sure how to implement it.

robinv85:
@omarzoghayyer ok yeah good point, I also noticed that, I think on the login page, for example "your password must be at least 8 characters" with a popup... but I think we might have to change that one too. Talking to a designer to determine a more consistent approach for all the error handling, will let you know if I know more.

robinv85:
@omarzoghayyer did you see on Slack and Figma designs and Github tickets that there are new requirements for the validation & error handling now? We will not use the pop-ups anymore, I will probably take a look at it tomorrow for some pages like login & signup page if nobody else did

omarzoghayyer:
@robinv85 I saw it the other day. It looks better. I might need help implementing this type of validation.

manualzuru:
@omarzoghayyer should we put this in New or open new issue?

robinv85:
@omarzoghayyer I'm working on the login / signup page with the react-hook-form library for the form validation and also the server-side validation, but for this need-help page and other parts of the application for 1 single field maybe it's overkill and we can use another  way to validate... but the display of the client-side errors should be inline, with red underline and a red error text... you can check examples in the designs

omarzoghayyer:
@robinv85 I checked the required design on figma. I can try implementing another library, but this might take me  a bit longer than expected.

robinv85:
@omarzoghayyer what do you mean with another library? To validate or to change the appearance of the form inputs?

omarzoghayyer:
@robinv85 I was looking at this one here https://upmostly.com/tutorials/form-validation-using-custom-react-hooks

robinv85:
@omarzoghayyer maybe we can use that approach in some places where we don't have real forms like in the step wizard if we only show 1 input
but for the larger full forms i think it's best to stay consistent everywhere and use react-hook-form instead of "reinventing the wheel" since it already has a lot of the required functionality included

robinv85:
closing this since there is a new PR

PR #333, opened 2020-05-12
#327 eslint config

### LINKED ISSUE #327: Extend ESLint config in frontend
This issue is created to support ESLint configuration for  #199 and #213 since there was some confusion and the pull requests that are currently in review each have their own way to fix this. 

Also see [my comment](https://github.com/FightPandemics/FightPandemics/pull/258#discussion_r422545352) on #258 

Since we use [Create React App](https://create-react-app.dev/), ESLint is already included and their documentation explains [how to extend ESLint config](https://create-react-app.dev/docs/setting-up-your-editor/#experimental-extending-the-eslint-config) and there are more similar settings explained [here](https://create-react-app.dev/docs/advanced-configuration/)

Also keeping in mind that:
- we have a `.env` file but it's not in version control so if the setting is put there it will have to be updated by everyone or they won't have the settings (and we need to document that) or they won't get the same linting rules
- it's possible to set it in the shell, maybe with [cross-env](https://www.npmjs.com/package/cross-env) if it should also be able to run the scripts locally
- the linting rules should also be picked up by the editor and run correctly with the husky pre-commit hook

robinv85:

Enable ESLint extension
Closes #327 

* Added `eslintrc.json` which extends the `react-app` presets to add or override rules, just added one rule for example `no-var` to warn not to use the var keyword
* Set the `EXTEND_ESLINT` environment variable to enable ESLint configuration (see [Create React App](https://create-react-app.dev/docs/setting-up-your-editor/#experimental-extending-the-eslint-config))
* Updated npm scripts to revert back to react-scripts and added separate lint script
* Added lint to pre-commit hook

Tested locally with `cross-env` but doesn't seem to work on Windows, neither setting in `.env` file, but the file does work in Docker and probably on Unix-based OS, and this setting is not needed for enabling the config in editors with ESLint plugin so seems ok to me.

Now I think the other PR's that depend on this can just update the config file and possibly add plugins if needed


robinv85:
@TenNga and @aelnahas can you guys check this and see if it will work for the other 2 tickets?

aelnahas:
just need to make small adjustments on my end but otherwise this is in line with what I had

PR #340, opened 2020-05-13
[295] Create a reusable heading component 

snowiesuet:
### PR Checklist:

We had a few different headings component and all of them have different styles. We should standardize them based on the design styleguide (Design Systems MVP).

- Replaced `CustomH1`, `Title/Title`, `Typography/Title` components with `Heading `and `TextLabel`

- There should be no reason to modify the style of the fonts as they are currently following the Design System MVP Typography.

- Props that `Heading` accept are color and level (1, 2, 3, 4)

- Props that `TextLabel` accept color, font-size (should be taken from theme.typography.font), font-weight and block (`TextLabel` is a span)


client/src/components/Typography/Heading.js
```diff
@@ -0,0 +1,36 @@
...
+
+const Heading = styled(Title)`
+  &.ant-typography {
+    font-weight: bold;
+    color: ${(props) => props.color};
+    ${theme.typography.heading.font}
+  }
+
+  &.h1 {
+    font-size: 4.2rem;
```
anniegiang, line 17: For all the font-sizes let's use them from `theme.js`

silenceway:
Thinking in the accessibility, Heading should return a header tag (h1, h2, h3, etc) instead of a class with the type.

anniegiang:
> Thinking in the accessibility, Heading should return a header tag (h1, h2, h3, etc) instead of a class with the type. @anniegiang: What do you think?

Good point. Does the `Title` component return a class instead of a heading element? If so, we may need to rethink this.

snowiesuet:
> Thinking in the accessibility, Heading should return a header tag (h1, h2, h3, etc) instead of a class with the type. @anniegiang: What do you think?

They are returning in h1, h2, h3, h4 but I couldn't find a way to edit the default font-size directly, so I added the class.

anniegiang:
> > Thinking in the accessibility, Heading should return a header tag (h1, h2, h3, etc) instead of a class with the type. @anniegiang: What do you think?
> 
> They are returning in h1, h2, h3, h4 but I couldn't find a way to edit the default font-size directly, so I added the class.

Overriding ant styles can be tricky, most of the time the selector needs to be very specific in order to override. But I think the current implementation works as well.

PR #341, opened 2020-05-13
Changed font-size static values in Heading to be from theme.js

snowiesuet:
### Quick Description of the PR
Changed the static values for `Heading` (#340 ) to use values from `theme.js`

PR #344, opened 2020-05-14
#341 - Change Heading's static values to variables from theme.js

snowiesuet:
Sorry for the multiple pull requests, there was a merge issue!

### DIFF
--- a/client/src/components/Feed/FilterBox.js
+++ b/client/src/components/Feed/FilterBox.js
@@ -42,7 +42,7 @@ const FilterBox = () => {
   return (
     <FilterBoxWrapper className="filter-box">
       <TextLabel
-        block={"true"}
+        block="true"
         color={DARK_GRAY}
         size={theme.typography.size.medium}
       >
--- a/client/src/components/Feed/FilterBox.js
+++ b/client/src/components/Feed/FilterBox.js
@@ -42,7 +42,7 @@ const FilterBox = () => {
   return (
     <FilterBoxWrapper className="filter-box">
       <TextLabel
-        block={true}
+        block={"true"}
         color={DARK_GRAY}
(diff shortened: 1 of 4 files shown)


client/src/components/Feed/FilterBox.js
```diff
@@ -42,7 +42,7 @@ const FilterBox = () => {
   return (
     <FilterBoxWrapper className="filter-box">
       <TextLabel
-        block={true}
+        block={"true"}
```
anniegiang, line 45: Let's remove the curly brackets, `block="true"` is fine

client/src/components/Typography/Heading.js
```diff
@@ -14,22 +14,22 @@ const Heading = styled(Title)`
   }
 
   &.h1 {
-    font-size: 4.2rem;
+    font-size: ${theme.typography.heading.one};
```
anniegiang, line 17: Let's destructuring `heading` like this:

`const { heading } = theme.typography`

and then replace wherever it's use, for example like this `heading.one`

client/src/pages/Profile.js
```diff
@@ -174,11 +174,15 @@ const Profile = (props) => {
       <div style={userInfoStyle}>
         {popover(props)}
         <ProfilePic noPic={true} initials={getInitials(firstName, lastName)} />
-        <TextLabel weight="500" block={true} size={theme.typography.size.large}>
+        <TextLabel
+          weight="500"
+          block={"true"}
```
anniegiang, line 179: Let's remove the curly brackets as well

client/src/components/Typography/Heading.js
snowiesuet: ah thanks! i was wondering how we can shorten it

Issue #343, opened 2020-05-14
[Mobile] Landing/Home Screen - description disappears when screen is rotating 


galeza:
env. https://staging.fightpandemics.work/

Steps:
1. Open https://staging.fightpandemics.work/ on your mobile
2. Rotate screen
3. When screen is rotated description : Pandemics will continue to happen. We help communities prepare and respond - is not visible

![Screenshot_2020-05-14-15-12-19-657_com android chrome](https://user-images.githubusercontent.com/21987083/81938884-dae22000-95f5-11ea-884d-171ef203404e.jpg)
![Screenshot_2020-05-14-15-12-25-699_com android chrome](https://user-images.githubusercontent.com/21987083/81938892-df0e3d80-95f5-11ea-8a86-0095181eb686.jpg)
Expected results: description is visible

More info:
Android 9 , Chrome

snowiesuet:
#345 related

PR #349, opened 2020-05-15
fixed disappearing of description when rotate

TenNga:
### PR Checklist:

Fix disappearing of description when screen rotate (mobile) issue https://github.com/FightPandemics/FightPandemics/issues/343

### **Issue**:
image

### **Fixed :**
image

* Changed media query and also margin to match mockup design

### DIFF
--- a/client/src/pages/Home.js
+++ b/client/src/pages/Home.js
@@ -17,7 +17,7 @@ const FlexChild = styled.div`
 
 const StyledIntro = styled.div` 
 
-  @media screen and ${mq.phone.narrow.max} {
+  @media screen and ${mq.phone.wide.max} {
     margin-top: 4rem;
   }
 
@@ -46,7 +46,7 @@ const MainContainer = styled.div`
     margin-left: -2.5rem;
   }
 
-  @media only screen and ${mq.phone.narrow.max} {
+  @media only screen and ${mq.phone.wide.max} {
     display: block;
   }
 `;
@@ -59,7 +59,7 @@ const StyledWelcome = styled.h1`
   font-style: normal;
 
   @media only screen and ${mq.phone.wide.max} {
-    font-size: ${theme.typography.size.large};
+    font-size: ${theme.typography.size.xlarge};
     margin: 2.5rem auto;
     font-weight: bold;
     text-align: center;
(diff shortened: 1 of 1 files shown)


client/src/pages/Home.js
```diff
@@ -77,13 +79,13 @@ const StyledStrapline = styled(StyledWelcome)`
   line-height: 7rem;
```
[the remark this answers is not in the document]
TenNga: I'm using the largest heading size, we need to add an extra size in theme

client/src/pages/Home.js
```diff
@@ -51,15 +51,14 @@ const MainContainer = styled.div`
...
-const StyledWelcome = styled.h2`
-  font-size: ${theme.typography.heading.three};
+const StyledWelcome = styled.h1`
+
   text-align: left;
   margin: 0 0 7rem 0;
   font-family: ${theme.typography.font.family.display}, sans-serif;
   font-style: normal;
 
   @media only screen and ${mq.phone.wide.max} {
```
anniegiang, line 61: For the mobile screen let's keep the font styles for the title and p tags like the designs:

current
Screen Shot 2020-05-14 at 7 02 24 PM

figma
Screen Shot 2020-05-14 at 7 04 37 PM

client/src/pages/Home.js
```diff
@@ -15,12 +15,13 @@ const FlexChild = styled.div`
   margin-bottom: 2rem;
 `;
 
-const StyledIntro = styled.div`
+const StyledIntro = styled.div` 
+
   @media screen and ${mq.phone.narrow.max} {
```
joshmorel, line 20: Hi @TenNga I'll add my review as I'm trying to help with getting PR's merged. I'd echo @anniegiang comments on consistent styling. To add the margin between the description and links should be consistent(ish) w phone.narrow.
```suggestion
  @media screen and ${mq.phone.wide.max} {
```
If you think `4rem` is too much as it pushes the first button too much down the screen then at least `2rem`. But I don't think that's an issue keeping them the same is probably best.

client/src/pages/Home.js
TenNga: DONE.

client/src/pages/Home.js
TenNga: I used phone wide cause I need to show a mobile version when the screen is under 767px. and the phone narrow doesn't have option for 767px.

TenNga: Thank you for your review @joshmorel. I totally agree with you and now I will change narrow to wide right away

joshmorel:
lgtm.

Related but potentially separate issue the `phone.wide.max` of 767px is actually too small for larger phone (Pixel XL, iPhone X*), but increasing that would mean we're overlapping with `tablet.narrow.min` of `768px` which could lead to some issues for tablet. 

For MVP and given we are developing native apps, I don't think we need to worry about tablets much. However, if we do want to match all phones to Mobile design and all Tablets to desktop design regardless of rotation, we'd need to include `orientation` in the media query style rules.

PR #373, opened 2020-05-16
Add options for unauthenticated user at /posts

Naraujo13:
### PR Checklist:

* [X] Have you looked if there might be other open [Pull Requests](https://github.com/FightPandemics/FightPandemics/pulls?q=is%3Apr+is%3Aopen+sort%3Aupdated-desc) for the same update/change?

Allow unauthenticated users to access the `/posts` endpoint to access the feed

- removed auth pre hook
- added logic for unauthenticated users, removing visibility filters and changing sorting from location to most recent

* To help understand them.

### DIFF
--- a/backend/lib/endpoints/posts.js
+++ b/backend/lib/endpoints/posts.js
@@ -35,7 +35,7 @@ async function routes(app) {
     async (req) => {
       const { userId } = req.query;
       let user, userErr;
-      if (userId != null) {
+      if (userId) {
         [userErr, user] = await app.to(User.findById(userId));
         if (userErr) {
           throw app.httpErrors.notFound();
--- a/backend/lib/endpoints/posts.js
+++ b/backend/lib/endpoints/posts.js
@@ -30,21 +30,25 @@ async function routes(app) {
   app.get(
     "/",
     {
-      preValidation: [app.authenticate],
       schema: getPostsSchema,
     },
     async (req) => {
       const { userId } = req.query;
-      const [userErr, user] = await app.to(User.findById(userId));
-      if (userErr) {
-        throw app.httpErrors.notFound();
+      let user, userErr;
+      if (userId != null) {
+        [userErr, user] = await app.to(User.findById(userId));
+        if (userErr) {
+          throw app.httpErrors.notFound();
+        }
       }
 
       // Base filters - expiration and visibility
       /* eslint-disable sort-keys */
       const filters = [
         { $or: [{ expireAt: null }, { expireAt: { $gt: new Date() } }] },
-        {
+      ];
+      if (user){
+        filters.push({
           $or: [
             { visibility: "worldwide" },
             {
@@ -63,8 +67,8 @@ async function routes(app) {
               "author.location.city": user.location.city,
             },
           ],
-        },
-      ];
+        })
+      }
       /* eslint-enable sort-keys */
 
       // Additional filters
@@ -74,7 +78,6 @@ async function routes(app) {
       }
 
       // Unlogged user limitation for post content size
-      // TODO: how to check for logged/unlogged user?
       /* eslint-disable sort-keys */
       const contentProjection = user
         ? "$content"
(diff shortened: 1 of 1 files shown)


backend/lib/endpoints/posts.js
```diff
@@ -30,21 +30,25 @@ async function routes(app) {
...
-      preValidation: [app.authenticate],
       schema: getPostsSchema,
     },
     async (req) => {
       const { userId } = req.query;
-      const [userErr, user] = await app.to(User.findById(userId));
-      if (userErr) {
-        throw app.httpErrors.notFound();
+      let user, userErr;
+      if (userId != null) {
```
joshmorel, line 38: Thanks @Naraujo13. This looks like it resolves something I was going to ask about so happy to get it merged. I can't really test the logged-in flow because we need the user V2 with location etc which you're working on.

Only nit-picky suggestion for now I would make is remove `!= null` as seeing this I think we're expecting to maybe see a `null` when really it will be `undefined` or a string (`objectId`). 
```suggestion
      if (userId) {
```

Two suggestions for improvement related to this - for **post-MVP** refactoring backlog - would be to either update `app.authenticate` with `optional` arg, or add `app.optionallyAuthenticate` middleware for such a case instead of passing userId as query string arg. Additionally, also do the fetching of the user in the middleware and add to `req` so we're accessing `req.user` in all the endpoints using such middleware.

robinv85: Yeah good suggestion @joshmorel 
That will work here, just have to be careful in some cases where the `if ( something )` could be any truthy value but it would also fail for anything falsy such as `0 ` and `false` and `""`

backend/lib/endpoints/posts.js
```diff
@@ -30,21 +30,25 @@ async function routes(app) {
   app.get(
     "/",
     {
-      preValidation: [app.authenticate],
       schema: getPostsSchema,
     },
     async (req) => {
       const { userId } = req.query;
```
robinv85, line 36: what is the point of `userId` in the query param? isn't this to get the posts posted by a specific user?
I thought posts was always a public route, but seems like here this is different for some filters, specifically location.
Putting id in query is not secure if you put another user's id? But I might not understand what's happening here.
I think @joshmorel his suggestion to add `optional` param to the `authenticate` hook is actually a great suggestion, or maybe `app.optionalAuthenticate` as alternative or something similar, so we can add `userId` on the request if available

Naraujo13: The posts are returned sorted by distance to the user at the `/posts` endpoint.
For that, if the user is authenticated, we get the user's location to serve as a reference point (to calculate distance and etc).

And I agree that an "optional" param or another hook would be the best way here. 
This endpoint is always public, but the behavior changes based on whether the user is authenticated or not (if he is, sort by his location, otherwise it users other sorting criteria).

Naraujo13: And about not being safe, yeah, ideally that would be handled by the optional auth hook, but even with it as query param, it is not a big problem here since the id is only used here to decide there reference point for sorting the posts (putting another user id will just change posts order).

robinv85: ah ok ... well not sure if we want to allow anyone to paste anyone else's user id in there and see what it returns? :) maybe i'm thinking too far but you never know how someone could abuse this kind of things...

Naraujo13: Yeah, ideally we probably don't want that, but until we have an optional auth hook that may be the best way (and with no major side effects). 

So I suggest leaving it like this until we have the hook and then we switch.

joshmorel: lgtm (for now). I think adding a new ticket for optional auth hook is a good idea. I'll do so.

Naraujo13: Please do so and link to this so we can remember do update it here

PR #378, opened 2020-05-17
Data model/comments v2

### LINKED ISSUE #205: Change Comment data model to v2
Change the Comment mongoose model to [v2 data model version](https://github.com/FightPandemics/DataModel) and adapt the comments endpoints to reflect those changes.

Naraujo13:

Update Comment data model to v2, while adjusting post endpoints.
Add `/posts/:postId/comments` get endpoint with pagination

Closes #205 

Refactor Endpoints:

- `POST /posts/:postId/comments`
- `PATCH /posts/:postId/:commentId`
- `DELETE /posts/:postId/:commentId`
- `PUT /posts/:postId/:commentId/:userId` (like)
- `DELETE /posts/:postId/:commentId/:userId` (unlike)

Added Endpoints:
- `GET /posts/:postId/comments` - gets the comments for that post with pagination

All comment endpoints require auth and comments returned are returned in a tree structure (nested up to two levels), with comments sorted by creation date.


backend/lib/models/Comment.js
```diff
@@ -1,54 +1,53 @@
-const { model, Schema } = require("mongoose");
+const { Schema, model, ObjectId } = require("mongoose");
+const { schema: authorSchema } = require("./Author");
```
joshmorel, line 2: Hi @Naraujo13 why are we importing the `Author.schema` if we aren't using it (same as in model/Post.js actually)? Should it be `author: authorSchema` instead of `author: Object` in the comment/post schema definition?

Naraujo13: Good catch, I've done this at posts also.
The validation for nested documents (author is nested inside comment in this case, and the location is another nested level at author) was giving some problems so I've thought of leaving then as plain objects and adding the authors and locations schema references in a later PR. This way we can have endpoints available to start the integrations with frontend and speedup MVP (while they integrate we can work on these improvements).

joshmorel: Sounds good to me. I'll take a more holistic look tomorrow.

backend/lib/endpoints/posts.js
```diff
@@ -461,22 +564,22 @@ async function routes(app) {
...
-      const {
-        params: { commentId, postId },
-        userId,
-      } = req;
-      const updatedComment = await Comment.findOneAndUpdate(
-        { _id: commentId, likes: userId, postId },
-        { $inc: { likesCount: -1 }, $pull: { likes: userId } },
-        { new: true },
+      const { commentId, userId } = req.params;
+
```
robinv85, line 568: first this needs to check if `req.params.userId === req.userId` or you could remove someone else's like

backend/lib/endpoints/posts.js
```diff
@@ -418,41 +504,58 @@ async function routes(app) {
...
-      const { commentId, postId, userId } = req.params;
-      if (userId !== req.userId) {
-        return new httpErrors.Forbidden();
-      }
-      const updatedComment = await Comment.findOneAndUpdate(
-        { _id: commentId, likes: { $ne: userId }, postId },
-        { $inc: { likesCount: 1 }, $push: { likes: userId } },
-        { new: true },
+      const { commentId, userId } = req.params;
+      const [updateErr, updatedComment] = await app.to(
```
robinv85, line 545: first this needs to check if `req.params.userId === req.userId` or you could add someone else's like

backend/lib/endpoints/posts.js
```diff
@@ -374,42 +374,128 @@ async function routes(app) {
...
-      const { comment } = body;
-      const { commentId, postId } = params;
-      const updatedComment = await Comment.findOneAndUpdate(
-        { _id: commentId, authorId: userId, postId },
-        { comment },
-        { new: true },
-      );
-      if (!updatedComment) {
-        return new httpErrors.BadRequest();
+      const { content, userId } = req.body;
```
robinv85, line 479: `userId` comes from JWT here and not the request body (also needs to update the schema)
this is handled by the `authenticate` hook which adds this on `req.userId`

backend/lib/endpoints/posts.js
```diff
@@ -418,41 +504,58 @@ async function routes(app) {
...
+        throw app.httpErrors.internalServerError();
       }
-      return { deletedCount, success: true };
+
+      const {
+        deletedCount: deletedNestedCount,
+        ok: deleteNestedOk,
+      } = await Comment.deleteMany({ parentId: commentId });
+      if (deleteNestedOk !== 1) {
+        app.log.error("failed removing nested comments for deleted comment", {
```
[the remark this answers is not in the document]
Naraujo13: You mean deleting the comment and failing for the nested?
In that case, we would have dangling comments, the same for deleting post and failing for comments.

The comments would be left dangling their parents here but wouldn't be shown/returned/counted (since we always start at base level and go up, they would never be reached).

Ideally, we would wrap this in a transaction, this way if deleting nested fails we can always abort it and revert changes, but that requires setting up the replica set.

backend/lib/endpoints/posts.js
Naraujo13: Okay, thanks on that. 
I've updated here and it probably needs to be changed on `/posts` endpoint too

backend/lib/endpoints/posts.js
```diff
@@ -374,42 +374,122 @@ async function routes(app) {
...
+        throw app.httpErrors.internalServerError();
+      }
+
+      reply.code(201);
+      return comment;
     },
   );
 
-  app.put(
+  app.patch(
```
anniegiang, line 470: Looking at the designs, I don't think there are options to edit or delete a comment ? I've asked the product team to clarify about this.

Naraujo13: Good point, from what I've seen from the talk with the design team it wasn't expected to have  `update/delete` operations, but I see no harm in having those ready

backend/lib/endpoints/posts.js
```diff
@@ -374,42 +374,122 @@ async function routes(app) {
...
+  );
+
+  app.post(
+    "/:postId/comments",
+    { preValidation: [app.authenticate], schema: createCommentSchema },
+    async (req, reply) => {
+      const userId = req.userId;
+      const { parentId } = req.body;
+      const { postId } = req.params;
+      const { body: commentProps } = req;
```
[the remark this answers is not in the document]
robinv85:
```` 
const {  body: commentProps, params, userId } = req;
```
;)

backend/lib/endpoints/posts.js
```diff
@@ -374,42 +374,121 @@ async function routes(app) {
...
       const { postId } = params;
-      if (parentId) {
-        const parentPost = await Post.findById(parentId);
-        if (!parentPost || parentPost.postId !== postId) {
-          return new httpErrors.BadRequest();
-        }
+
+      const [userErr, user] = await app.to(User.findById(userId));
+      if (userErr) {
+        req.log.error("Failed retrieving user", { userErr });
```
joshmorel, line 435: Sorry for the delay in reviewing but wanted to get caught up in posts/users/orgs first. Anyways these scripts like `const [userErr, user] = await app.to(User.findById(userId));`, If the query is legit and nothing else errors out but the user doesn't exist for whatever reason - both `userErr` and `user` are `null`. Should we also include `if ( userErr || user === null)` instead of just `if (userErr)`?

robinv85: @joshmorel since you've added a separate PR to fix these issues can we approve this now?

backend/lib/endpoints/posts.js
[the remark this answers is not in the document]
robinv85: ok we have talked about this PR in the meeting and thought it's probably better to get this merged ASAP and integrate with frontend so it can be tested
@Naraujo13 do you think we should make a new issue for these transactions?

PR #379, opened 2020-05-17
Adjust post endpoints to reflect changes at user model

Naraujo13:
Changed author name composition with first and last to single name attribute to reflect user model changes

### PR Checklist:

* [X] Have you looked if there might be other open [Pull Requests](https://github.com/FightPandemics/FightPandemics/pulls?q=is%3Apr+is%3Aopen+sort%3Aupdated-desc) for the same update/change?

Change author name composition with first and last to single name attribute to reflect user model changes

* Change author name composition
* Fix edge case at `GET /posts/:postId`

### DIFF
--- a/backend/lib/endpoints/posts.js
+++ b/backend/lib/endpoints/posts.js
@@ -167,7 +167,7 @@ async function routes(app) {
       postProps.author = {
         id: user.id,
         location: user.location,
-        name: `${user.firstName} ${user.lastName}`,
+        name: user.name,
         type: user.type,
       };
 
@@ -242,7 +242,7 @@ async function routes(app) {
         throw app.httpErrors.internalServerError();
       }
 
-      const { comments = [], numComments = 0 } = commentQuery;
+      const { comments = [], numComments = 0 } = commentQuery[0];
 
       return {
         comments,


joshmorel:
Hi @Naraujo13 . This looks fine but for my understanding is this just to match current state User model or will we be having single `name` in v2 for both organization and individual user?

Naraujo13:
> Hi @Naraujo13 . This looks fine but for my understanding is this just to match current state User model or will we be having single `name` in v2 for both organization and individual user?

@joshmorel Yes, it is to match current user state. 
Initially, V2 was projected with two separate fields, but recently we started discussing if that was really needed.

joshmorel:
> > Hi @Naraujo13 . This looks fine but for my understanding is this just to match current state User model or will we be having single `name` in v2 for both organization and individual user?
> 
> @joshmorel Yes, it is to match current user state.
> Initially, V2 was projected with two separate fields, but recently we started discussing if that was really needed.

Probably would be simpler overall and less confusing for some non-English users so I support this. The benefits of storing first & last name are minimal I suspect.

backend/lib/endpoints/posts.js
```diff
@@ -242,7 +242,7 @@ async function routes(app) {
         throw app.httpErrors.internalServerError();
       }
 
-      const { comments = [], numComments = 0 } = commentQuery;
+      const { comments = [], numComments = 0 } = commentQuery[0];
```
joshmorel, line 245: I did approve but don't really understand the reason for this specific change. It's hard for me to confirm if this is right until all the v2 data model changes come in.

Naraujo13: The `groupBy` stage at the aggregation pipeline with a `null` group criteria will group all records into only one element (this way it also counts the comments).

So for this query, the results will always be in the first position as a single object, with the array of comments as an attribute, this way we would always be looking at the first element here.

Naraujo13:
> > > Hi @Naraujo13 . This looks fine but for my understanding is this just to match current state User model or will we be having single `name` in v2 for both organization and individual user?
> > 
> > 
> > @joshmorel Yes, it is to match current user state.
> > Initially, V2 was projected with two separate fields, but recently we started discussing if that was really needed.
> 
> Probably would be simpler overall and less confusing for some non-English users so I support this. The benefits of storing first & last name are minimal I suspect.

Currently, the main defender of first + last name is the marketing team who probably wants do differentiate between them to use at campaigns and etc, but I think a single field might be the better option.

Issue #179, opened 2020-04-26
Connect API with the Feed for displaying Posts


anniegiang:
The`/feed` page displays posts with the following features:

- [ ] On initial render, the Feed component makes a `GET` request to `/api/posts/` to fetch posts to display.
- [ ] On subsequent renders, the Feed component makes a `GET`  request to `<tbd_endpoint>` to fetch **filtered posts** based on the selected filters.
  - [ ] The feed should have endless scrolling, with lazy loading of 10 posts.
  - [ ] Posts are sorted by distance from the users.
  - [ ] Posts authored as "Sourced by FightPandemics" are always appended at the end.
  - [ ] A not logged-in user sees at most 25 words of a post's description.
  - [ ] A logged-in user sees the post's full description. 

More reference:
https://www.notion.so/fightpandemics/Feed-653658508c4649869787c48f3e4eee13

Contact @anniegiang for questions about how the frontend of the Feed was built.

PR #380, opened 2020-05-17
[179] Add api calls to feed page.

silenceway:
### PR Checklist:

https://github.com/FightPandemics/FightPandemics/issues/179
Add API post calls to Feed page.

Modified posts reducer and added endpoint call.

Note: These changes will not work if https://github.com/FightPandemics/FightPandemics/pull/373 is not merged before.

### DIFF
--- a/client/src/components/TextAvatar/index.js
+++ b/client/src/components/TextAvatar/index.js
@@ -1,0 +1,22 @@
+import styled from "styled-components";
+import { Avatar } from "antd";
+
+import { ROYAL_BLUE, SELAGO } from "constants/colors";
+
+const TextAvatar = styled(Avatar)`
+  background:  ${SELAGO};
+  border: 1px solid ${ROYAL_BLUE};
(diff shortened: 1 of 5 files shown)


client/src/components/TextAvatar/index.js
```diff
@@ -0,0 +1,22 @@
+import styled from "styled-components";
+import { Avatar } from "antd";
+
+import { ROYAL_BLUE, SELAGO } from "constants/colors";
+
+const TextAvatar = styled(Avatar)`
```
anniegiang, line 6: Nice! This reminds me - the avatars are going to have name initials if there is no photo. I remember researching a third-party library that generates avatars that have initials. We can implement this later.

silenceway: This component receives initials from the user name if the user has no photo. Current posts service doesn't return this field so I need to add initials avatar.

PR #389, opened 2020-05-18
improved error handling for geolocation api

Chimaobi-Johnson:

Added improved error handling for geolocation api 

### Before submitting this pull request:

3. The name of this branch is: **`feature/<branch_name>`**.


client/src/components/NearestHospital/HealthFacilities.js
```diff
@@ -151,28 +154,70 @@ const NearestHealthFacilities = props => {
...
+                 successCallback,
+                 errorCallback_lowAccuracy,
+                 {maximumAge:600000, timeout:30000, enableHighAccuracy: false}
+               );
+
+              return;
+            }
+            setGettingLocation(false);
+              let errorMsg;
+              if (error.code === 1) {
```
robinv85, line 181: good to see there are error codes, didn't know these were added to the backend
we could use this as a map from code / messages or use a switch case as well to return the right message... but here it's ok for now with the if / else because there are only a few codes

client/src/components/NearestHospital/HealthFacilities.js
```diff
@@ -151,28 +154,70 @@ const NearestHealthFacilities = props => {
...
+
+    const successCallback = position => {
+      setGettingLocation(false);
+      setCoords(position.coords)
+    }
+
+    navigator.geolocation.getCurrentPosition(
+      successCallback,
+      errorCallback_highAccuracy,
+      {maximumAge:600000, timeout:25000, enableHighAccuracy: true}
```
robinv85, line 214: might be better to extract these values to constants instead of using "magic numbers"

PR #397, opened 2020-05-18
Fixed unselect/select options

omarzoghayyer:

When clicking on options, its either you select all options or you select none of these. It was missing the '!none' next the checked under AnswerCheckbox.
Screen Shot 2020-05-18 at 1 02 54 PM

### DIFF
--- a/client/src/pages/SymptomsCheck.js
+++ b/client/src/pages/SymptomsCheck.js
@@ -1,5 +1,4 @@
-import React, { useState } from "react";
-
+import React, { useState} from "react";
 import {
   AnswerButton,
   AnswerCheckbox,
@@ -44,7 +43,6 @@ const Welcome = (props) => {
     </WizardStep>
   );
 };
-
 const Step1 = (props) => {
   const onSelectAnswer = (answer) => {
     props.update("age", answer);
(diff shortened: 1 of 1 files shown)


client/src/pages/SymptomsCheck.js
```diff
@@ -167,7 +165,7 @@ const Step3 = (props) => {
           key={i}
           text={answer}
           onSelect={() => toggleAnswer(answer)}
-          checked={checked}
```
snowiesuet, line 170: The issue now is if I select this option, I cannot select any other options unless I deselect it. I think we should be able to select other options but uncheck "None of those".

omarzoghayyer: @snowiesuet i'm not sure what you mean. Currently, you can select all options and when you select 'None of These', it should unselect all.

snowiesuet: Yes, after you select "None of those", you should be able to select other options which should deselect "None of those". But currently selecting "None of those" disables other options, I'm not sure if this is a bug or by design?

omarzoghayyer: I implemented it based on step 2. Im not sure why they initially started it this way.

snowiesuet: Ah, I see. I guess this was already there, sorry about the confusion!

Issue #207, opened 2020-04-29
Change Organization data model to v2


Naraujo13:
Change the Organization mongoose model to [v2 data model version](https://github.com/FightPandemics/DataModel) and adapt the organization's endpoints to reflect those changes.

PR #399, opened 2020-05-19
Feature/create org profile (mobile)

alexjmcm:
### What does this pull request aim to achieve?

_Please be concise and link any related issue(s):_
Create create organization profile page: https://github.com/FightPandemics/FightPandemics/issues/69

### Before submitting this pull request:

_Please do **NOT** submit this PR if you have not done the following:_

1. I have checked that no one else is working on similar changes.
2. I have read the latest [**README**](https://github.com/FightPandemics/FightPandemics/blob/master/README.md) and understand the development workflow.
3. The name of this branch is: **`feature/<branch_name>`**.
4. This branch is rebased/merged with the **latest master**.
5. There are no merge conflicts.
6. There are no console warnings and errors.
7. On the right hand side of this PR, I have linked the appropriate issue(s) under "Issues" and added a project progress under "Projects".

Issue #393, opened 2020-05-18
Add geocoding (lat/lon from address)


joshmorel:
Current Behaviour:

The geo-service only does reverse geocoding (lat/lon to address).

Expected Behaviour:

Ability to send address and get back latitude/longitude coordinate pair. This is required for load of posts sourced by FightPandemics (to assign coordinates from City, State, Country entered by data team) but likely other purposes as well.

PriyalPatney:
I am working on this.

joshmorel:
@Naraujo13 Do you think we still need this on the backend (or the geo-service at all) since all geocoding will be done on the front-end with Google Maps api before sending to the backend?

We should decide before @PriyalPatney gets too far with this one way or the other.

I can come up with some other solution for the import from airtable.

Naraujo13:
> @Naraujo13 Do you think we still need this on the backend (or the geo-service at all) since all geocoding will be done on the front-end with Google Maps api before sending to the backend?
> 
> We should decide before @PriyalPatney gets too far with this one way or the other.
> 
> I can come up with some other solution for the import from airtable. 
Isn't the frontend currently using the reverse geocode endpoint of the python service to do that?

joshmorel:
I'm personally not sure the best choice. Current we are using https://github.com/thampiman/reverse-geocoder to get country from lat/lon for a V1.1 feature. It's for offline use but hasn't been updated in awhile so how accurate is it for finer details? We should bring this discussion to Slack engineering.

joshmorel:
Closing as created new issue with better outlined requirements #469

PR #401, opened 2020-05-19
Data model/users v2

### LINKED ISSUE #206: Change User data model to v2
Change the User mongoose model to [v2 data model version](https://github.com/FightPandemics/DataModel) and adapt the user endpoints to reflect those changes.

joaofnds:
Update User data model to v2

Closes #206 and #207

- Remove old models
- Add User, IndividualUser and OrganizationUser
- Update endpoints to use the correct model (IndividualUser for /users and OrganizationUser for /organizations)
- Update endpoint schemas


backend/lib/endpoints/schema/organizations.js
```diff
@@ -26,22 +26,25 @@ const getOrganizationsSchema = {
```
Naraujo13, line 26: Why update the `updateOrganizationSchema` and not the `createOrganizationSchema`?

backend/lib/endpoints/schema/location.js
```diff
@@ -0,0 +1,16 @@
+const S = require("fluent-schema");
+const { strictSchema } = require("./utils");
+
+const { LOCATION_TYPES } = require("../../models/Location");
+
+const locationSchema = strictSchema()
```
Naraujo13, line 6: Missing `state` and `zip`

backend/lib/models/User.js
```diff
@@ -1,33 +1,50 @@
...
+    { "author.authorId": this._id },
+    { $set: { "author.location": location } },
+  );
+
+  return location;
+}
+
+userSchema.path("location", {
+  set: updateAuthorLocationReference,
+});
```
Naraujo13, line 37: Actually, we don't need to keep the location ref synced.
If someone updates their location, the post/comments shouldn't move with them, so we don't need this callback here.

backend/lib/models/IndividualUser.js
```diff
@@ -0,0 +1,79 @@
+const { Schema } = require("mongoose");
+const { model: User } = require("./User");
+const { updateAuthorName: updatePostAuthorName } = require("./Post");
+const { updateAuthorName: updateCommentAuthorName } = require("./Comment");
```
Naraujo13, line 4: Where are this methods define?
They are not at `Post` and `Comment` at master and from what I understood you didn't add them at this PR.

backend/.eslintrc
```diff
@@ -9,5 +9,13 @@
     "global-require": "off",
     "no-underscore-dangle": "off",
     "sort-keys": ["error", "asc"]
-  }
+  },
+  "overrides": [
+    {
+      "files": ["lib/models/**/*.js"],
```
robinv85, line 15: ah ok might be a good idea, wouldn't it still be better if the schema fields are ordered? 
But just the indexes aren't or do they have to be sorted on the fields also? Otherwise could be more clear to have an inline comment and only at the part where it matters...

backend/lib/endpoints/schema/location.js
robinv85: and how about required values? Can we have something like `oneOf`` at least one is required?

backend/lib/models/IndividualUser.js
joaofnds: Originally added on [this commit](https://github.com/FightPandemics/FightPandemics/pull/401/commits/e39b358d1ad91fca7af3735f6df53a046747e345) but I think the merge deleted them 🤔

backend/lib/endpoints/schema/location.js
joaofnds: As this is a shared "component" of a schema, I decided not to make any of them required. Can we compose those schema definitions? If not maybe I should make all of them required

joaofnds: @Naraujo13 this is missing on the model also '-'

backend/.eslintrc
joaofnds: Decided to put it there because someone might "autofix" something in the model and affect the  database indexing. ESlint become that tool "that" you run "that" command to make it happy. Someone could be writing an index, tried to commit but ESlint complaints, so they run `npm run lint` to fix it and them commit with the wrong index order.

That's what I had in mind when decided to put that config in there, I can change it back if you want!

joaofnds:
> Some small adjustments, but overall I've liked the way you used the setters to keep the data synced.
> 
> Two bigger questions about things I've missed:
> 
> 1. Why the schema for the organization creation endpoint was not updated?
> 2. I've seen you didn't change anything at the endpoints themselves (just at model and the endpoints schema). That was not necessary?

I tried to keep changes to a minimum and just "swap" models. I plan to add missing fields and schemas on another PR

backend/lib/endpoints/schema/organizations.js
joaofnds: Considering [this comment](https://github.com/FightPandemics/FightPandemics/pull/401#issuecomment-630917180), should I add it in this PR?

backend/.eslintrc
joaofnds: What do you think of it? Should I change it back?

backend/lib/endpoints/schema/location.js
joaofnds: Can we do this on another pull request? Location is a shared schema and I'm trying to merge users on this one. What do you think?

joshmorel:
> > Some small adjustments, but overall I've liked the way you used the setters to keep the data synced.
> > Two bigger questions about things I've missed:
> > 
> > 1. Why the schema for the organization creation endpoint was not updated?
> > 2. I've seen you didn't change anything at the endpoints themselves (just at model and the endpoints schema). That was not necessary?
> 
> I tried to keep changes to a minimum and just "swap" models. I plan to add missing fields and schemas in another PR

@joaofnds You removed the location sync for good or just as part of this PR? With the "Sourced by FightPandemics" posts I've been setting author.type & author.location custom for each post depending on airtable fields (as required for post filtering - more importantly location). In my script I will create one org per type for "Sourced by FightPandemics" to avoid any kind of unintentional update to the post.author.type but I just want to make sure I don't need to worry about location.

joaofnds:
> You removed the location sync for good or just as part of this PR? With the "Sourced by FightPandemics" posts I've been setting author.type & author.location custom for each post depending on airtable fields (as required for post filtering - more importantly location). In my script I will create one org per type for "Sourced by FightPandemics" to avoid any kind of unintentional update to the post.author.type but I just want to make sure I don't need to worry about location.

@joshmorel Location isn't meant to be synced between author and his posts, so I think it's all good with your script

backend/lib/endpoints/schema/location.js
Naraujo13: The missing fields were added in the PR #403, they were really missing in the original model. 

About the required fields, I think this might be tricky here. I think what @robinv85 meant with the `oneOf` was that there is two options for the user to input the location:

- Only with coordinates, with no address - this way we can reverse geocode then to get full address
- Only the address (from gmaps list), with no coordinates - this way we geocode the coordinates to get full address 

This is to always keep coordinates and address consistent without needing to add extra checks in the backend.
So I think he meant a `oneOf` between those two options. 

Anyway, I wonder if this would be done by the backend or before sending the request to the API because if done before, the location would always be sent with all fields.

backend/lib/models/Comment.js
```diff
@@ -50,5 +50,23 @@ CommentSchema.index({
 
 const Comment = model("Comment", CommentSchema);
 
-exports.model = Comment;
-exports.schema = CommentSchema;
+function updateAuthorName(authorID, newAuthorName) {
+  return Comment.where(
+    { "author.authorId": authorID },
```
Naraujo13, line 55: Author doesn't have `author` prefix anymore. 
Should be `author.id` and `author.name` at the updates here.

Naraujo13: Probably should be careful to not conflict with #378 too

backend/lib/models/Post.js
```diff
@@ -159,10 +158,26 @@ postSchema.index({ "author.id": 1, createdAt: -1 });
...
 postSchema.index({ likes: 1 });
 
-/* eslint-enable */
-
 // -- Model
 const Post = model("Post", postSchema);
 
-exports.model = Post;
-exports.schema = postSchema;
+function updateAuthorName(authorID, newAuthorName) {
```
Naraujo13, line 164: Same here as in comments, shouldn't use `author` prefix on attributes anymore

joshmorel:
>     Only with coordinates, with no address - this way we can reverse geocode then to get full address
>     Only the address (from gmaps list), with no coordinates - this way we geocode the coordinates to get full address
> 
> This is to always keep coordinates and address consistent without needing to add extra checks in the backend.
> So I think he meant a oneOf between those two options.
> 
> Anyway, I wonder if this would be done by the backend or before sending the request to the API because if done before, the location would always be sent with all fields.
> 

@Naraujo13 Do you think it's important we clarify how this flow will work as a project pretty soon (given desire to launch MVP)?

Right now I see, in for example in this organization profile form - https://github.com/FightPandemics/FightPandemics/pull/399/files# - it's just a free text location (no city/state split out or lat/lon) that is posting to the back-end. It's called "location" but I guess it maps to "address" in our model.
image

So we'd need to geocode the lat/lon from that location (work in progress?) and then reverse_geocode (we already have the function in geo-service) to get city/state/country (see below example output).

But if we're supposed to be doing it a different way we should all get on the same page so there's not such a disconnect between front-end and back-end work.

@PriyalPatney did start work on geocoding #393 I believe so if this is the right direction maybe we're good. But we should be clear and all on the same page. 


reverse_geocode for coords [10, 10]
```
{
    "data": {
        "admin1": "Bauchi",
        "admin2": "",
        "cc": "NG",
        "lat": "10.11161",
        "lon": "9.82604",
        "name": "Yamrat"
    },
    "result": "sucess"
}```

File: backend/lib/models/User.js [hunk 8]
```diff
@@ -1,33 +1,31 @@
...
+const { Schema, model } = require("mongoose");
+const { schema: locationSchema } = require("./Location");
+const { isValidEmail } = require("../utils");
 
-const UserSchema = new Schema(
+const userSchema = new Schema(
   {
-    _id: {
+    about: { maxLength: 100, trim: true, type: String },
+    authId: { required: true, type: String },
```
joshmorel, line 8: @joaofnds Is this supposed to be the auth0 id? If so shouldn't it only be on IndividualUser?

joshmorel:
So based on @Naraujo13 comment in slack https://fightpandemics.slack.com/archives/C010UL6GDAS/p1590338351000400?thread_ts=1589384526.008100&cid=C010UL6GDAS the front-end should always be sending location details.

a) Detect location: "The first would use browser geolocation, and the then use Google maps to reverse geocode."
b) Enter address: "The second would use gmaps to search for addresses with the user imputed address and them geocode."

So all of the location elements (except neighborhood, do we still need this?) should be sent unless for Google Maps API only sometimes returns certain elements, for example, zip/postal_code?

backend/lib/models/User.js
joaofnds: @joshmorel Not according to the [data model](https://github.com/FightPandemics/DataModel/blob/master/v2/scripts/JSON_Schema/applicationDB/users.json#L23), but maybe we should revisit that! What do you think @Naraujo13? 

Having the `authID` in the User model would allow the organization and its owner to log in independently

joshmorel: @joaofnds  But where is the `authId` coming from on organization creation? https://app.zeplin.io/project/5ebb0c7bc826984e0abb929b/screen/5ebf499e21d04848a62753dc

We're only creating an Auth0 account based on email/password at user sign up time no?

Is my understanding correct, that you sign up as an individual user. You can then post as an individual user or create an organization and post as that, where the `ownerId` associates the org with the signed-in individual (if we want multiple users per org - that would be a post-MVP feature).

joaofnds: Yeah, you're right, I'll remove it! We can add it later if needed anyway 😄

Great catch btw!

joshmorel:
The location stuff - anything that isn't perfect we'll resolve with #469 . I'll take a look tomorrow about everything else but hopefully we can merge this soon.

backend/lib/endpoints/schema/organizations.js
```diff
@@ -1,15 +1,39 @@
 const S = require("fluent-schema");
 const { strictSchema } = require("./utils");
+const { locationSchema } = require("./location");
+
+const organization = {
+  about: S.string(),
+  email: S.string(),
```
robinv85, line 7: do we have email verification for an organization?
Or should we only use an email from a user?
Currently the email verification for user accounts is completely handled by Auth0 and we should be careful with saving any email address to our own database, I would prefer to let Auth0 handle it for now.

backend/lib/endpoints/schema/organizations.js
```diff
@@ -26,22 +50,17 @@ const getOrganizationsSchema = {
...
-    .prop("linkedinUrl", S.string())
-    .prop("location", S.string())
-    .prop("name", S.string())
-    .prop("needs", S.array().maxItems(10).items(S.string()))
-    .prop("ownerId", S.string())
-    .prop("twitterUrl", S.string())
-    .prop("type", S.string())
-    .prop("website", S.string()),
+    .prop("about", organization.about)
+    .prop("email", organization.email)
```
robinv85, line 54: same thing, can you just update an email address? 
Or maybe it can only be updated when it's verified

backend/lib/models/Post.js
```diff
@@ -64,8 +64,8 @@ const postSchema = new Schema(
 );
 
 // -- Indexes
-/* eslint-disable */
 // Indexes for filtered feed
+/* eslint-disable sort-keys */
```
robinv85, line 68: ah ok I thought this was in the lint config now, but this is more clear probably so everyone sees that this part is an exception to the rule

backend/lib/models/User.js
```diff
@@ -1,33 +1,32 @@
...
+const { Schema, model } = require("mongoose");
+const { schema: locationSchema } = require("./Location");
+const { isValidEmail } = require("../utils");
 
-const UserSchema = new Schema(
+const userSchema = new Schema(
   {
-    _id: {
+    about: { maxLength: 100, trim: true, type: String },
+    email: {
```
robinv85, line 8: yea so I'm wondering about emails, if we should save them in our own database
and if we should allow updating through the API
We currently get email also from Auth0 users, also with social media accounts, so I would prefer to use those, and we could save them from the Auth0 user instead of through the API, or at least has to be checked that it's the same anyway.

backend/package.json
```diff
@@ -28,6 +28,7 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
+    "ajv": "^6.12.2",
```
robinv85, line 31: ajv is included in fastify already, i think it was previously removed because it was in there to override the configuration but it's not needed as a separate dependency (might cause conflicting versions as well?)

backend/lib/models/IndividualUser.js
```diff
@@ -23,6 +23,10 @@ function updateAuthorLastName(lastName) {
 
 const individualUserSchema = new Schema(
   {
+    authId: {
```
joshmorel, line 26: @joaofnds I actually have more questions about this one. Is this supposed to be like `auth0|5ecd...` from Auth0? @robinv85 do we need this at all since it's the app metadata mongo id on the Auth0 side that is linking these. Also, I don't see where this would come from in `/users.js` `post("/"...`. We get _id: userId from plugins/auth and everything else from POST body.

Once we clarify this and the questions about email storage I think I'm good to merge.

robinv85: Ah yeah I didn't notice this.
Either we have to auto generate the primary id for organizations
Or we have to create a user account for an organization with an email
but, can it be the same as an individual user? In that case we cannot have 2 user accounts
unless we have a separate dtabase in Auth0 for organizations as well
but we cannot login with email / password and match it to both a individual and an organization user anyway
So we always login as individual, which can be the admin / owner of an organization...
and we don't really have organizations in Auth0, unless in a separate databse? 
Maybe we need to double check to coordinate product requirements with technical possibilities and limitations here.

robinv85: oh maybe i have misunderstood, I though this was for the organization
but if it's only for the individual user then I think it's the same as the mongo_id
this is generated by Auth0 as a `BSON.ObjectId` so we can use it as the `_id` primary key in the users database, we could also store the Auth0 id as well if we need to call the Auth0 API from our backend, so maybe it's this?

robinv85: and this also comes from the JWT in that case which is in `req.user.sub` (subject of the JWT, but containing the Auth0 id which depends on the login method)

joshmorel: If anything it would be the `req.user.sub`.

I'm just not sure why we need to store it on the model. But if we do we need to set it somewhere like, for example in users.js app.post:

`const userData = {
...req.body,
_id: req.userId,
authId: req.user.sub
}`

joshmorel:
@joaofnds @robinv85  I asked about email in the product channel. If the resolution isn't clear by tomorrow mid-day for this PR to get it done and move forward:

1. `authId`: get rid of it for now, we can add it later if we need it
1. `email`: make it optional in both org endpoint and model for now

backend/lib/models/IndividualUser.js
robinv85: yeah I think it's best that we save that also, maybe it's not used at the moment, but what if we need to call the Auth0 API with a user id that logged in, I guess we could use the access token, but usually we use the server token and the access token might need to be changed if it's not for use in the Auth0 API directly, unless we limit scope to read only.

Naraujo13:
> @joaofnds @robinv85 I asked about email in the product channel. If the resolution isn't clear by tomorrow mid-day (EST) for this PR to get it done and move forward:
> 
> 1. `authId`: get rid of it for now, we can add it later if we need it
> 2. `email`: make it optional in both org endpoint and model for now

@joshmorel  so we're keeping `authId`? 
I agree with @robinv85 on this, we might not need it now, but it may be useful later as a FK and doesn't add much complexity.

About the `email`, the org is from a single user, kinda like an additional profile for him. Therefore, a user needs to log into his personal account to them access the org. 
The `email` field at org is the contact email displayed at its profile, it is not used to authenticate or to fill posts email. With that, I agree that the organization email should be optional.

About not storing the email on our side, I think storing it is not a big deal and will be really useful for BI/Analytics reports where we need to have user contact info. Those will be routine for mkt actions and removing email from ou database would make us depend ou Auth0 for that, so I think we should keep them at our side too.

backend/lib/models/User.js
Naraujo13: About not storing the email on our side, I think storing it is not a big deal and will be really useful for BI/Analytics reports where we need to have user contact info. Those will be routine at  mkt actions and removing email from ou database would make us depend on Auth0 for that, so I think we should keep them at our side too.

I think restricting the update through the API and linking it to the social email is okay as long as we keep it at our side synced to use them at metabase and other BI reports.

joshmorel:
> @joshmorel so we're keeping `authId`?
Yep let's keep it. I'll comment on email in slack.

backend/package.json
joaofnds: The first time I tried to run the server node could not find ajv 🤷‍♂️

backend/lib/endpoints/users.js
```diff
@@ -50,6 +50,7 @@ async function routes(app) {
       const userData = {
         ...req.body,
         _id: req.userId,
+        authId: req.user.sub,
```
joshmorel, line 53: @joaofnds looks like email storage is a yes. For org all looks good - it comes through request body & schema.  For individual user we can do:

```suggestion
        authId,
        email,
```

Both can be destructured from `user` on line 46:

`const { email, email_verified: emailVerified, sub: authId  } = user;` 

After that I have no more suggestions and am happy to get this merged!.

robinv85: If we follow the same behavior as the front-end, maybe it's better to only allow creating user if `email_verified === true` and otherwise return `403 FORBIDDEN` response.
In the frontend, we will check this and redirect to email verification page anyway, so probably better if API follows same behavior, will also be needed for mobile app I guess.

joshmorel: @robinv85 doesn't this meet the requirement or am I missing something?

https://github.com/FightPandemics/FightPandemics/blob/9892a582e9f7817ecb0f9a7f28bb38218e3d6ef6/backend/lib/endpoints/users.js#L47

```
      if (!emailVerified) {
        throw app.httpErrors.forbidden("Email address not verified");
      }
```

robinv85: yes that seems good indeed!

robinv85:
> Maybe something I forgot to ask and relating to data model, @Naraujo13 @joaofnds
> Do these 2 fields also need to be saved in the db? When creating user profile they need to be checked:
> 
> ```
> By signing up, I agree to the Privacy Policy
> By signing up, I agree to the Terms and Conditions
> ```

ok lgtm, going to approve this, just these 2 fields, maybe going to ask legal we can still add it later

Naraujo13:
> > Maybe something I forgot to ask and relating to data model, @Naraujo13 @joaofnds
> > Do these 2 fields also need to be saved in the db? When creating user profile they need to be checked:
> > 
> > ```
> > By signing up, I agree to the Privacy Policy
> > By signing up, I agree to the Terms and Conditions
> > ```
> 
> ok lgtm, going to approve this, just these 2 fields, maybe going to ask legal we can still add it later

I forgot to answer that before, sorry!

About these two fields, a user is required to check them to be register right? Otherwise he is blocked from following on.

If that is the case, every user would need to check those to register, making those two fields to always be true...
Data-wise, that makes them semantically useless, making having them the same as not having them since all records would have the same value haha

But if this is required by legal we can add them just to be sure

robinv85:
> 
> 
> > > Maybe something I forgot to ask and relating to data model, @Naraujo13 @joaofnds
> > > Do these 2 fields also need to be saved in the db? When creating user profile they need to be checked:
> > > ```
> > > By signing up, I agree to the Privacy Policy
> > > By signing up, I agree to the Terms and Conditions
> > > ```
> > 
> > 
> > ok lgtm, going to approve this, just these 2 fields, maybe going to ask legal we can still add it later
> 
> I forgot to answer that before, sorry!
> 
> About these two fields, a user is required to check them to be register right? Otherwise he is blocked from following on.
> 
> If that is the case, every user would need to check those to register, making those two fields to always be true...
> Data-wise, that makes them semantically useless, making having them the same as not having them since all records would always have the same value haha
> 
> But if this is required by legal we can add them just to be sure

Yeah I get your point, was also thinking similarly... I just remember I had to do this in the past once for some database to add a field to make sure a "EULA" license was accepted, probably also for legal reasons, but I can't remember exactly the circumstances. Also, in theory what if someone adds a user through the API directly? :)

backend/lib/models/User.js
robinv85: yep if email for org is not really used at the moment it's not a problem, we can check it later

Naraujo13:
> > > > Maybe something I forgot to ask and relating to data model, @Naraujo13 @joaofnds
> > > > Do these 2 fields also need to be saved in the db? When creating user profile they need to be checked:
> > > > ```
> > > > By signing up, I agree to the Privacy Policy
> > > > By signing up, I agree to the Terms and Conditions
> > > > ```
> > > 
> > > 
> > > ok lgtm, going to approve this, just these 2 fields, maybe going to ask legal we can still add it later
> > 
> > 
> > I forgot to answer that before, sorry!
> > About these two fields, a user is required to check them to be register right? Otherwise he is blocked from following on.
> > If that is the case, every user would need to check those to register, making those two fields to always be true...
> > Data-wise, that makes them semantically useless, making having them the same as not having them since all records would always have the same value haha
> > But if this is required by legal we can add them just to be sure
> 
> Yeah I get your point, was also thinking similarly... I just remember I had to do this in the past once for some database to add a field to make sure a "EULA" license was accepted, probably also for legal reasons, but I can't remember exactly the circumstances. Also, in theory what if someone adds a user through the API directly? :)

I was talking to @joaofnds about this and about possible future updates on terms, EULA and etc. and he suggested to store the accepted version instead of a boolean. This would aggregate more than just the bool, since we could track users that didn't accept the new version and etc..

 lthough a lot of platforms just use that banner with the 'by continuing to use the platform you agree with the new terms' and etc for updates

Issue #180, opened 2020-04-26
Connect API for Liking and Unliking a Post


anniegiang:
- Logged in
  - [ ] Make a `PUT` request to `/api/:postId/likes/:userId` if a user HAS NOT liked a post.
  - [ ] Make a `DELETE` request to `/api/:postId/likes/:userId` if a user HAS liked a post.
- Not logged in
  - Redirect to login/sign up

More reference:
https://www.notion.so/fightpandemics/Feed-653658508c4649869787c48f3e4eee13

Contact @anniegiang for questions about how the frontend of the Feed was built.

Naraujo13:
@limaneto Backend endpoints merged at #284

PR #402, opened 2020-05-19
[180] Post like behavior.

silenceway:

https://github.com/FightPandemics/FightPandemics/issues/180:
It allows a registered user to like a post.

* [x] I have checked that no one else is working on similar changes.
* [x] I have read the latest [**README**](https://github.com/FightPandemics/FightPandemics/blob/master/README.md) and understand the development workflow.
* [x] The name of this branch is: **`feature/<branch_name>`** (need to have cloned the repo not forked).
* [x] This branch is rebased/merged with the **latest master**.
* [x] There are no merge conflicts.
* [x] There are no console warnings and errors.
* [x] On the right hand side of this PR, I have linked the appropriate issue(s) under "Issues" and added a project progress under "Projects".


client/src/pages/Feed.js
```diff
@@ -258,6 +260,41 @@ const Feed = () => {
...
+      const endPoint = `/api/posts/${postId}/likes/${user && user.userId}`;
+      const post = posts.posts.filter((post) => post._id === postId);
+
+      if (user) {
+        if (!post[0].liked) {
+          axios.put(endPoint)
+            .then((response) => {
+              postsDispatch({ type: SET_LIKE, postId, count: response.data.likesCount });
+            })
+            .catch(error => {
```
anniegiang, line 279: Should we store the errors in Redux? That way the Post can access errors and display them. What do you think?

client/src/pages/Feed.js
```diff
@@ -258,6 +260,41 @@ const Feed = () => {
     dispatchAction(TOGGLE_STATE, "showFilters");
   }
 
+  const handlePostLike = (postId) => {
```
anniegiang, line 263: Let's use async-await syntax for this function to be consistent with the code-base.

client/src/pages/Feed.js
```diff
@@ -258,6 +260,41 @@ const Feed = () => {
...
+    const { history, isAuthenticated, user } = props;
+
+    /* added here because userId not working */
+    sessionStorage.removeItem("likePost");
+
+    if (isAuthenticated) {
+      const endPoint = `/api/posts/${postId}/likes/${user && user.userId}`;
+      const post = posts.posts.filter((post) => post._id === postId);
+
+      if (user) {
```
anniegiang, line 273: Why keep this as a separate check from line 269? On 269 why not ` if (isAuthenticated && user)` ?

client/src/pages/Feed.js
```diff
@@ -258,6 +260,41 @@ const Feed = () => {
...
 
+  const handlePostLike = (postId) => {
+    const { history, isAuthenticated, user } = props;
+
+    /* added here because userId not working */
+    sessionStorage.removeItem("likePost");
+
+    if (isAuthenticated) {
+      const endPoint = `/api/posts/${postId}/likes/${user && user.userId}`;
+      const post = posts.posts.filter((post) => post._id === postId);
```
anniegiang, line 271: Maybe we could pass `post` as an argument instead of doing an extra search here?

client/src/pages/Feed.js
```diff
@@ -258,6 +260,41 @@ const Feed = () => {
...
+
+      if (user) {
+        if (!post[0].liked) {
+          axios.put(endPoint)
+            .then((response) => {
+              postsDispatch({ type: SET_LIKE, postId, count: response.data.likesCount });
```
anniegiang, line 277: Can we add this line after the `if-else` statements since we need to update state either way? Just thought this would be DRYer.

client/src/pages/Feed.js
```diff
@@ -293,7 +330,17 @@ const Feed = () => {
...
         postsDispatch({ type: ERROR_POSTS });
       })
-  }, []);
+  }, [ props ]);
+
+  useEffect(() => {
+    if (posts.status === SET_POSTS && posts.posts.length) {
+      const likePost = sessionStorage.getItem("likePost");
+
+      if (likePost) {
```
anniegiang, line 339: Can you clarify why we are calling `handlePostLike ` here?

client/src/hooks/reducers/feedReducers.js
```diff
@@ -51,6 +52,14 @@ export const postsReducer = (state = postsState, action) => {
       return { ...state, status: SET_POSTS, posts: action.posts };
     case ERROR_POSTS:
       return { ...state, status: ERROR_POSTS, posts: [] };
+    case SET_LIKE:
+      return { ...state,
+        posts: state.posts.map((post) => (
```
anniegiang, line 57: I think there's a more efficient to update the liked state than to loop through every post. Perhaps we could create a reducer that handles actions for a single post?

On the same note, should we create a separate `postReducer` to handle actions like this? In the future we need to handle things like commenting, liking comments, error handling, etc. So I think adding all the state logic here for a single post may be bloating the code and logic.

robinv85: adding logic in reducers makes it easy to test input / output from functions (like redux docs recommend as well)
post reducer sounds nice... i'm a bit confused at the moment how we are using redux and hooks with useReducer, maybe we should get rid of redux if we are only using it for the authentication & user ... 
to make it more efficient, maybe the state should be normalized and changed to have `posts: { id: value }` as an object like a map with keys and values

client/src/pages/Feed.js
silenceway: The current user object is not been created after login, if I do (isAuthenticated && user) it will always throw false and creates an infinite loop.

client/src/pages/Feed.js
silenceway: The PostSocial component is not receiving all the post object, I will do the change.

client/src/pages/Feed.js
silenceway: If a guest user clicks like, it goes to the login page and when logged user comebacks to feed page, like is saved.

client/src/hooks/reducers/feedReducers.js
silenceway: Will the backend be changed to reflect `id: values` or do I need to do it in the frontend?
If we use a postReducer, how it will be filled? Do all results have a reducer or It will be filled when the user interacts with a post? Which information will store postsRedurer and which postReducer? If postReducer is updated, it needs to update postsReducer post' information too?

client/src/pages/Feed.js
silenceway: And I asked about how to obtain userId information and I didn't receive any answer about this field.

client/src/pages/Feed.js
anniegiang: Ah ok. But why not handle this logic in `PostSocial`? I think it makes more sense to handle this behavior in that component than in the Feed. What do you think?

client/src/hooks/reducers/feedReducers.js
robinv85: The common best practice to normalize the state is to change arrays with objects into an object with key/value and keep a separate list with only the id's in memory and also use that in other places where you need to present the data in a different way... there is extensive documentation about this on the redux docs https://redux.js.org/recipes/structuring-reducers/normalizing-state-shape and also some related blog posts ... I think this is probably a good idea to also do this with the useReducer hooks

robinv85:
````
? Currently, the posts service doesn't return if a post was liked by a user, it's missing it, how do we handle this?
````
Yes good question, I also thought about that when looking at the data model, this was also my idea but since for our use case posts are mostly "read-heavy" the data model is designed for that, and I think we can get that info from the API, maybe check with @Naraujo13  because there might be a change needed, for example if authenticated user requests the list of posts, add an extra boolean if their user id is in the "likes" list

Comment (silenceway) [reply client/src/hooks/reducers/feedReducers.js:57]: > The common best practice to normalize the state is to change arrays with objects into an object with key/value and keep a separate list with only the id's in memory and also use that in other places where you need to present the data in a different way... 

Ok. I will do the change in SET_POSTS dispatch after receive the service response.

Comment (Naraujo13) [reply client/src/hooks/reducers/feedReducers.js:57]: @robinv85 I can add that feature at the `/posts` endpoint.
Just adding a boolean that checks if the user liked the post or not should be easy. 

@silenceway could you open an issue specifying these changes? Also, review if you need any more additional info from the posts' backend.

Comment (robinv85) [reply client/src/hooks/reducers/feedReducers.js:57]: yeah ok wasn't sure if this was already implemented, seems like we forgot about that, but this is also a good reason why we should use the authenticated hook in the API for this request to get the user id, but then optional...
I was actually wondering if this was also allowed by the OpenAPI standard because I haven't seen it yet, and not sure if this will show in our auto-generated API documentation, but seems like it should be possible https://github.com/OAI/OpenAPI-Specification/issues/14#issuecomment-297457320

File: client/src/pages/Feed.js (hunk 2 above)
Comment (silenceway) [reply client/src/pages/Feed.js:263]: Done.

File: client/src/pages/Feed.js (hunk 4 above)
Comment (silenceway) [reply client/src/pages/Feed.js]: Done.

File: client/src/pages/Feed.js (hunk 5 above)
Comment (silenceway) [reply client/src/pages/Feed.js]: Done.

File: client/src/pages/Feed.js (hunk 6 above)
Comment (silenceway) [reply client/src/pages/Feed.js]: Done.

File: client/src/actions/authActions.js [hunk 8]
```diff
@@ -2,8 +2,8 @@ import axios from "axios";
 import { Toast } from "antd-mobile";
 
 import { GET_ERRORS } from "./types";
-import { AUTH_LOGIN, AUTH_SIGNUP } from "constants/action-types";
-import { setAuthToken, getAuthToken } from "utils/auth-token";
```
robinv85: Yeah so this wasn't added because the local storage was a bit of a temporary thing when first developing, currently it wasn't really specified in the requirements and I think out of scope here to store the JWT in local storage after every login.

client/src/hooks/reducers/feedReducers.js
silenceway: Updated the reducer.

client/src/actions/authActions.js
silenceway: Lines commented.

anniegiang:
Is this good to merge?

silenceway:
> Is this good to merge?

Yes

PR #407, opened 2020-05-19
Feature/[406]Changed default text from in my zip code to in my city

omarzoghayyer:

Expected State:
replaced "In my City"/"City" in createPostSettings.js instead of "In my Zip Code"/"Zip Code"


snowiesuet:
Pretty straightforward, just want to check with @joshmorel if this is what he needs.

Issue #194, opened 2020-04-29
Implement seeding scripts to create data for testing/development


mushketyk:
Implement a seeding script that should create projects and organizations.

We can use Mongo Seeding project for this: https://github.com/pkosiec/mongo-seeding

mushketyk:
Duplicate of #163

joshmorel:
@mushketyk - on further consideration I don't think this is a duplicate of #163. #163 is about importing data sourced by the data collection team into the production database mongo both one-time and as part of incremental load. It could be used for development as well but separating this makes sense I think due to the different purpose.

This would be for seeding databases for all collections (not just userd/posts) with dummy data for development & testing.

I did start my work on the other ticket with mongo-seeding and was able to get it to work fine - until I needed to asynchronously pull data from airtable. I've decided to switch to mongodb native driver (eventually mongoose) for that but for this mongo-seeding should work.

indirabhowmick:
@joshmorel - Should I assign the ticket to you?

joshmorel:
@indirabhowmick I assigned myself. This isn't required for MVP but developer productivity. it is different from #408 as it would be "dummy" data for dev/test of more than just posts - users, comments etc. I will work on it after #408 is done because post imports from Airtable might be enough for now for devs working on MVP.

indirabhowmick:
@joshmorel - Thanks so much for the detailed explanation

PR #408, opened 2020-05-19
create a script to import posts from airtable

### LINKED ISSUE #163: Create a script that parses and imports posts data into database
Currently we have data from Airtable that we want to migrate over to MongoDB.

We want a script that parses the data from Airtable and structures the data to fit our current models. The script should also insert the parsed data into our MongoDB database.

joshmorel:
Resolves #163 

I'd like some feedback on this WIP but I won't merge until #401 is merged and I've updated this for the V2 user model. One other major TODO is geocoding coordinates, but this can wait and be included with #393 work.

Apologizes for the long comment but the understanding of the role Airtable will play post MVP is a bit murky so I felt I should err on the long side in explaining the rational for my approach.

**Production Usage**:
We'll need to import data at launch plus a few times afterwards as the data collection team may source more data or fix existing records. However, eventually all data management should be done through our app - by users as well as admins for governance - so at a certain point we _shouldn't need this_.

There is only one other thing we we're getting from Airtable - country emergency codes on the front-end - and I believe this would be a one time simple import to be done once we figure out how we're modeling and managing that data (post-MVP).

**Dev/QA Usage**:
This can also be used be dev/QA so they have a sample of posts to work with. There's one argument to the script to limit records to fetch so they don't need to load all ~2500. I could add a library like `commander` or `yargs` as a means to improve code/documentation if that makes sense.

I'm not sure if  #194 is still being worked on but this would probably be a better option generally - so we're seeding in dev/test based on local files instead of airtable api.

**Code Organization**:
Because I'm assuming this will be the only integration with Airtable, I decided to keep it simple and not add generic Airtable utility function. Everything to process data is in the main script (`backend/scripts/import-posts/index.js`) plus a function for mapping data to from the Airtable record to our mongoose schema (`backend/scripts/import-posts/map-airtable-data.js`). This could be improved on later if my assumption is wrong.

I did consider coding this as an api/endpoint instead of a script, however, the later seemed a better choice to just get something working. If we need to have, for example, a button that an admin could press to import data this changed, but we'd need not only authenticate but ensure they are authorized as an admin (I assume admin-specific functionality is something we'll need eventually?).

**Performance Considerations**:
For 2500 records, it takes ~8s to fetch from Airtable and 4s to load locally via mongoose. I am not sure how network latency will affect the mongoose loading in production (Edit: on further thought maybe quite a bit as each query is await-ed so a few minutes at least. I think `Promise.all` with map would improve script execution time to an extent, however, if anyone knows if mongoose bulk operations would work for this use case let me know and I'll look more into it.). If we need to run more than a few times post-launch we can add a filter to only fetch airtable records based on last modified date.

Airtable does not limit or charge monthly API requests. There is only a rate limit of 5 requests/s - not a problem here.

**Data Consistency Considerations**:
The upsert is based on an airtable record id that will not change if sorted/updated/filtered. Only by duplicating the Airtable base and using the copied base would we have a problem. I will make sure data team understands this for post go-live. Anything deleted in Airtable will be orphaned so these would need to be deleted manually.


backend/scripts/import-posts/map-airtable-data.js
```diff
@@ -0,0 +1,76 @@
...
+    share_with: [visibility] = ["worldwide"],
+    state,
+  } = post.fields;
+  /* eslint-enable prefer-const */
+
+  if (authorType === "Others") {
+    authorType = "Other";
+  } else if (authorType === "Non-Profit") {
+    authorType = "Non-profit";
+  }
```
Naraujo13, line 39: These small data adaptations, maybe we should discuss with Andrés and data collection team to use the same format as we expect it in the DB from now on and just do a "replace all" to change the old ones.
What do you think?

backend/scripts/import-posts/map-airtable-data.js
```diff
@@ -0,0 +1,76 @@
...
+  objective = objective.toLowerCase();
+  visibility = visibility === "Global" ? "worldwide" : visibility.toLowerCase();
```
Naraujo13, line 41: Most of these `.toLowerCase()` will be done at the model automatically (check if the attribute at `Post` model has `lowercase: true` for that). The same from `trim`.

backend/scripts/import-posts/map-airtable-data.js
```diff
@@ -0,0 +1,76 @@
...
+  if (country) [author.location.country] = country;
+  if (state) [author.location.state] = state;
+  if (city) [author.location.city] = city;
+  // TODO: geocode country/state/city
+  author.location.coordinates = [0, 0];
+
+  const externalLinks = { website: website.trim() };
+  if (appStore && appStore !== "NA") externalLinks.appStore = appStore.trim();
+  if (playStore && playStore !== "NA")
+    externalLinks.playStore = playStore.trim();
```
Naraujo13, line 61: Same for the `trim` as for the `toLowerCase` before, check if it is done at the model, if it is not and should always be done we can add there for consistency.

backend/scripts/import-posts/index.js
```diff
@@ -0,0 +1,97 @@
...
+        : (existingPosts += 1);
+    } catch (err) {
+      failedPosts += 1;
+      console.error(
+        `${record.id} failed to import with fields: ${JSON.stringify(
+          record._rawJson.fields
+        )}. Error message: ${err.message}`
+      );
+    }
+  }
```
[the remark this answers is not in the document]
joshmorel: What I anticipate is in production, since we're sending the request from AWS to MongoDB Atlas one at a time, network latency could become a significant factor for my current script (while on local docker network it isn't). 

If we're only doing this a handful of times in production I don't want to spend too much time on optimization until we need to, but I'll look into this.

backend/scripts/import-posts/map-airtable-data.js
joshmorel: I asked him. I think it's a very simple change to just change a select option in. Since we're doing the cleaning manually already, such custom code shouldn't really be needed in our script. The exception email, I asked Andres about it and it sounds like it would take a lot of time to through them all again.

backend/scripts/import-posts/map-airtable-data.js
joshmorel: You are right about trim/toLowerCase . I can remove these but we need to add (trim only) for `externalLinks`. I'll create a separate issue.

```
    externalLinks: {
      appStore: String,
      email: String,
      playStore: String,
      website: String,
    },
```

backend/scripts/import-posts/index.js
```diff
@@ -0,0 +1,97 @@
...
+    try {
+      const res = await Post.findOneAndUpdate(
+        { airtableId: record.id },
+        mapAirtableData(record._rawJson, {
+          id: sourcedById,
+          name: SOURCED_BY_FP_USER.name,
+        }),
+        { new: true, upsert: true, useFindAndModify: false }
+      );
+      res.updatedAt.toString() === res.createdAt.toString() // eslint-disable-line no-unused-expressions
```
joaofnds, line 57: Maybe we should hide this logic behind a method in the Post schema. Below is an example of that:
```js
// backend/lib/models/Post.js
postSchema.methods.hasBeenUpdated = function hasBeenUpdated() {
  return this.updatedAt.toString() === this.createdAt.toString()
}
```

joshmorel: Hi @joaofnds I would agree if we we're using this anywhere else. I think of this script as a rough (hopefully) temporary thing to bootstrap our app so adding functions elsewhere just for it doesn't make sense. If I use bulk write as suggested by Nic I probably will get the update/insert count a different way - http://mongodb.github.io/node-mongodb-native/3.1/api/Collection.html#~BulkWriteOpResult.

I'll try that tomorrow and hopefully I'll have a final PR for approval tomorrow or saturday at the latest.

Naraujo13: About using bulk, if we are gonna use it I would suggest putting to set the  `updateAt` at the same update doc to keep it consistent.

Something like this
```
update_one: {
  $set: {
    fields...
    updateAt: <<currentTime>>
  }
}
```

backend/scripts/import-posts/index.js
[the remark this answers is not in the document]
joshmorel: Got it working and it seems to work nicely. It cut time locally in ~half but in production it should have a significant impact.

backend/scripts/import-posts/map-airtable-data.js
joshmorel: Some was cleaned but I added `validationSync` on each doc and found new issues. Neither `bulkWrite` nor `findOneAndUpdate` will validate based on the schema (although they will transform to lowercase, trim whitespace).  For things that I can fix in the script (i.e. it's obvious what the correct alternate value is) I don't want to force new work on the data collection team. I just need to update for user model v2 and then it should be good.

backend/scripts/import-posts/index.js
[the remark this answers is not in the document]
mannykary: @joshmorel - just wanted to mention this in regards to your comment about latency between AWS and MongoDB, we will have a [VPC peering connection](https://docs.atlas.mongodb.com/security-vpc-peering/) between the two which runs on the global AWS backbone connection (meaning, the traffic stays within the AWS network, and never goes across the internet). But it's always nice to optimize regardless.

joshmorel:
This should work now. Things got a little uglier with organizational users but important thing is it works and as outlined previously should be only necessary for a temporary period.

Without geocoding service I did my own temp thing for distinct locations in Airtable - https://gist.github.com/joshmorel/9cc496fe88fb7c84ed7fa5ffcf7c432c - and saved JSON mapping to project. Again a bit ugly but it works and should be necessary for a temporary period.

Please take a look, try locally with instructions outlined in README.md and let me know. Ask my in slack for Airtable keys if you don't have them (NOT the same as current ones on front-end).

joshmorel:
Closing in favor of #555 so we have review branch

Issue #369, opened 2020-05-16
Client Side Error Handling for Web in User Authentication Flow


sumerjoshi:
This ticket is kind of a catch all - basically going through the code and updating every situation of which there is possibility for a client error.

Current Behavior: Right now, I don't believe there is a current Behavior for error handling client side errors.
Expected Behavior: Using inline error messages to handle client side errors

robinv85:
Checking design inconsistencies, possibly mobile / desktop web variations are different, this design seems wrong with a red "placeholder text" or "input value" instead of a validation error under the input.

robinv85:
Updated the issue description, removed the screenshot to avoid confusion, talked to designers and the conclusion is to stick to one consistent appraoch from the designs (red error text under the input and red underline)

PR #409, opened 2020-05-19
error messaage right under input

TenNga:
Resolve issue https://github.com/FightPandemics/FightPandemics/issues/369
Error message:
image

3. The name of this branch is **`feature/<branch_name>`** (need to have cloned the repo not forked).
7. On the right-hand side of this PR, I have linked the appropriate issue(s) under "Issues" and added project progress under "Projects".


snowiesuet:
I don't think the .json files are meant to be committed?

TenNga:
> I don't think the .json files are meant to be committed?
How can I fix that I don't fine client/report....json

robinv85:
Hi @TenNga I was actually working on this page as well for the error messages and have refactored the login / signup component to use the `react-hook-form` component, on [this branch ](https://github.com/FightPandemics/FightPandemics/tree/feature/auth-errors-frontend) but it's still a work in progres... anyway I think we should use that same library for all the forms to keep it consistent, it also includes error handling

client/src/pages/Login.js
```diff
@@ -200,36 +204,86 @@ const Login = ({ isLoginForm }) => {
     setConfirmPassword(e.target.value);
   };
 
+  const handleErrorMessage = (type) => {
+    if(type === "email"){
+      const email = document.getElementById("emailID");
```
anniegiang, line 209: I see what you're doing here but using DOM selectors/methods to change state is not the "React" way of working with the DOM. This is because React uses the virtual DOM and a diffing algorithm to track changes of the DOM. When you're changing the innerText and the style color on lines 211 and 212 - this will confuse React and potentially cause performance issues. 

I think the "React" way of approaching this is to use booleans within the JSX to toggle classNames/elements to show/hide error messages. 

@robinv85 @snowiesuet What do you think?

client/report.20200519.221358.68.0.001.json
```diff
@@ -0,0 +1,446 @@
+
```
anniegiang, line 1: Not sure what these `.json` files are, don't think we need them?

client/src/pages/Login.js
```diff
@@ -283,6 +338,7 @@ const Login = ({ isLoginForm }) => {
                   style={StyleInput}
                 />
               </InputWrapper>
+              <ErrorMessage id="email-error"></ErrorMessage>
```
anniegiang, line 341: As mentioned previously, one way to approach this is to toggle `ErrorMessage` if there are errors.

client/src/pages/Login.js
robinv85: I think this has to be refactored anyway because I've added the react-hook-form library on my branch for the error messages and that one handles most of the validation logic already ...

robinv85: and that library has a hook `useForm` that returns an `errors` object and different ways to handle validation etc. ... so probably better to merge my branch and then just finish the details with work from this branch for showing the actual validation messages

robinv85:
@TenNga sorry for the inconvenience but my changes are now also in this PR  #415 
I think it's better to use that library and update the client-side validation with that method

@sumerjoshi FYI this might be a good reason to keep the granularity of the tickets limited for closely related tasks, maybe a better way is to open 1 ticket with a check list of items, to avoid situations like this where more people work on exactly the same component for related issues.

robinv85:
@TenNga do you want me to add this to the current version of the form? I will oipen another ticket to make all the forms more consistent afterwards

MohamedEl1:
@TenNga any updates on this ?

robinv85:
@MohamedEl1 @TenNga I'll take this over, it has to be redone with the create-hook-form library client-side validation and error display, I can look at it for login / signup and create profile for now.

client/src/pages/Login.js
robinv85: Also one problem with these form inputs from Ant design is that they don't expose a prop to get the ref, other similar UI libraries do this with `InnerRef` that's also why I had to use the `Control` component to handle some of the input values...

robinv85:
After reviewing this code and checking the requirements, and asking some questions to the designers I decided it's better to close this PR.
The implementation was based on the wrong screenshot that was attached to the original issue, and this view on mobile was inconsistent with the desktop and correct way to show the errors.
So I will work on this on a new branch and open a new PR for this.

PR #414, opened 2020-05-20
trim Posts.externalLinks items

### LINKED ISSUE #413: Always trim whitespace from Post externalLinks
Current Behaviour:

Whitespace is not trimmed from external links with post creation.

Expected Behaviour:

Trim whitespace by adding changing `String` to `{ type: String, trim: true }` for all of these: 

```
    externalLinks: {
      appStore: String,
      email: String,
      playStore: String,
      website: String,
    },
```

joshmorel:
Resolves #413 

As per discussion: https://github.com/FightPandemics/FightPandemics/pull/408#discussion_r427715558

The fluent schema (url/email) would prevent a value with leading/trailing whitespace but for consistency & non-API data loading I've added `trim: true` to the schema.

### DIFF
--- a/backend/lib/models/Post.js
+++ b/backend/lib/models/Post.js
@@ -12,10 +12,10 @@ const postSchema = new Schema(
     },
     expireAt: Date,
     externalLinks: {
-      appStore: String,
-      email: String,
-      playStore: String,
-      website: String,
+      appStore: { trim: true, type: String },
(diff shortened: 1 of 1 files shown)


joshmorel:
From here on I'll make PRs from branches in FightPandemics (instead of joshmorel, as the "Post Review Build URL" won't work) but for this one I don't think it matters.

Issue #372, opened 2020-05-16
[Mobile] Handling errors from server for User Authentication Flow


sumerjoshi:
Current Behavior: In the case of server side error handling the mobile design, there is no current or uniform way to handle how to display client side error messages to the user.
Expected Behavior: For every situation where there is a server side error, use this error message design in the picture below.
<img width="883" alt="Screen Shot 2020-05-16 at 12 49 53 PM" src="https://user-images.githubusercontent.com/593277/82129354-dee97b80-9776-11ea-85f8-063f292d7ed6.png">

joshmorel:
@sumerjoshi  If no one picks it up by 20/05/2020 I will work on it.

I would suggest the PR to resolve this issue also address these other issues - #371 #370 #369

sumerjoshi:
@joshmorel I initially thought this would only be one PR, however, after talking to Robin, there might be multiple other places where this error design might need to be implemented. Just wanted to make sure each ticket was granular enough

robinv85:
@sumerjoshi thanks for opening the tickets, good that there are multiple tickets for this, however it seems a bit confusing, this isn't really a backend ticket, it's rather how the frontend handles errors coming from the backend. 
The actual responses on the backend are separate tasks, and should be implemented already, unless some specific cases like here #225 which needs some more work, and can also be improved by additional frontend validation etc.

sumerjoshi:
@robinv85 sounds good

PR #415, opened 2020-05-20
Feature/auth errors frontend

### LINKED ISSUE #370: [WEB] Handling errors from server for User Authentication Flow
Current Behavior: There is no uniform behavior in dealing with server side messages during the User Authentication Flow
Expected Behavior: In every case where there may be possibility for a server side error, there should should be a red box that appears with the appropriate server side error.
Screen Shot 2020-05-16 at 12 50 07 PM

robinv85:

Fixes #370 and #372 
I've refactored the form to use `react-hook-form` so we don't have to code all the validation etc. manually, and I think we should use this same library consistently for most of the forms in the app.

This means I've also already done part of the work of #409 except for the styling.
The `react-hook-form` has integrated validation and errors, currently disabling the submit button if any input is not valid.

I've disabled the client-side email validation for now, I think that function will need to be changed, we can just pass the regex to the validator directly instead I think, we could also pass the function to the `validate` prop ... but out of scope here so maybe it can also be integrated with #409 

For redux & useReducer hook, I've done most of the form validation with the `useReducer` without Redux, only the `isAuthenticated` is still using redux, including the `emailVerified` and the next step is to add the user in the response & integrate with "create user profile" page. 

I think it might be a good idea to remove redux completely to avoid confusion with `useReducer` actions & reducers, and use the context API instead for the authenticated user data... feedback welcome ofcourse.

### DIFF
--- a/client/src/actions/authActions.js
+++ b/client/src/actions/authActions.js
@@ -1,8 +1,4 @@
-import axios from "axios";
-import { Toast } from "antd-mobile";
-
-import { GET_ERRORS } from "./types";
-import { AUTH_LOGIN, AUTH_SIGNUP } from "constants/action-types";
+import { AUTH_SUCCESS } from "constants/action-types";
 import { getAuthToken } from "utils/auth-token";
 
 // Note: for production apps, both localstorage & cookies contain risks to store user & auth data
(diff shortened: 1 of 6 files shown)


robinv85:
@DanielAlexander84 Yeah good points, I didn't finish everything because @TenNga was working on the client-side validation in #409 

But to follow up with your remarks:

- The formState is managed by react-hook-form and disabled the submit button if `isValid` is false, and this behavior also depends on the configured mode, here "changed" triggers validation, so I suppose I need to double check how the initial form state being empty with required fields is not valid

- The toasts were replaced with the `ErrorAlert` component to show errors coming from the backend, besides schema validation in the API Auth0 also validates the passwords, we need to make sure this is aligned, and I know for sure that Auth0 is configured to have a password strength of min. 8 characters and including special characters. Since the client-side validation is handled in the other ticket I didn't use the form errors yet to actually show them visually for the inputs (since that is out of scope here for this 2 tickets)

robinv85:
@DanielAlexander84 yeah ok, thanks for the feedback, those issues definitely have to be fixed. 
I'm hoping to merge this first and then to add the client-side validation on top if it, and also use this same form component for the user creation and other forms so we have a consistent approach for all the forms. (Do you think that's ok @TenNga ? If you want I can help to update your branch with this one or something like that...)

PR #424, opened 2020-05-21
Add support for Cypress framework and first e2e test using Cypress - Feature/cypresse2e

galeza:

Detailed instruction can be found here: https://www.notion.so/fightpandemics/Instruction-how-to-instal-and-use-Cypress-bc6b0443f7aa4e93bbbd092d1635b7a6

There are no merge conflicts.


cypress/integration/e2eTests/homePage.js
```diff
@@ -0,0 +1,26 @@
...
+
+  getSecondLineText() {
+    return cy.get('h1 ~ p:nth-child(3)');
+  }
+
+  getLogo() {
+    return cy.get('div.am-navbar-left > a > img')
+  }
+  getNeedHelpElement() {
+    return cy.get('div:nth-child(1) > a > div');
```
robinv85, line 18: What's the strategy to find the right DOM elements for tests? 
Some of these look very brittle, depending on the exact HTML structure, e.g. "first child div of link within div"etc... 
Since we use Styled Components we don't have specific class names for the components either (unless we enable this through a Babel Macro / plugin which should be possible) 
Could it be that Cypress docs recommend to use `data-` attributes for this? I remember reading that in testing documentation but not sure if it was Cypress.

minsu3:
Hi, I think this github thread explains why the Post Review Build URL check failed: https://github.com/actions/first-interaction/issues/10 
The issue might be that merging the forked repository to the source repository causes a permission issue.

Instead, we could create a branch on the source repository and merge it from there.

cypress/integration/e2eTests/homePage.spec.js
```diff
@@ -0,0 +1,29 @@
+import HomePage from './homePage';
```
minsu3, line 1: Just curious, what calls homePage.spec.js?

galeza:
> Rebuilt to see if it works, also wondering what you think about the selectors for DOM elements.

Hi @robinv85 ! The problem with check looks the same as in the PR: https://github.com/FightPandemics/FightPandemics/pull/408. Github Fork security issue/feature???: https://github.com/actions/first-interaction/issues/10.

robinv85:
@galeza ah, right, if I understand correctly from reading @mannykary his comment on Slack for another similar issue, this is because the PR is created from your fork and not from a branch on this repo, so that changes the permissions somehow.

cypress/integration/e2eTests/homePage.js
galeza: Hi @robinv85,
Yes you're right. It's recommended to use data- attributes to have test more resilient. Is it possible to enable/add data-* attributes?

galeza:
> @galeza ah, right, if I understand correctly from reading @mannykary his comment on Slack for another similar issue, this is because the PR is created from your fork and not from a branch on this repo, so that changes the permissions somehow.

@robinv85 yes there is a problem with permissions. There are another PRs created from fork like: #408, #402 (also from fork). I sent message to @joshmorel

joshmorel:
Hi @galeza . Yeah I realized that too. If we want to have a review URL we'll need to merge from another branch in this repo not a fork. 

If you don't add "feature/" to the branch name it will skip that check which would have been fine for this one. I think that's only necessary for PRs requiring user testing so I wouldn't worry here about that failing check.

cypress/integration/e2eTests/homePage.js
robinv85: Yeah I'm not a big fan of adding these data attributes everywhere but I guess if QA engineers want to do it it's ok, if we could use the StyledComponents plugin to automatically generate the class-names based on Component / File name I would also like that, maybe even more, then we could used more classes.

PR #426, opened 2020-05-21
Update styling for Create User Profile 

### LINKED ISSUE #176: Update styling for Create User Profile for desktop
- [x] Update current Create User Profile page styling for desktop.

Refer to the latest design in FIGMA under Desktop App MVP.
https://www.figma.com/file/4cV84H1jyMo0rJEAl3HNnJ/FightPandemics-landing-page?node-id=1377%3A8027

- [x]  Designs need to be pixel perfect

T-Allen:

Created a CreateUserProfile page and added styling for desktop and mobile.

- New page CreateUserProfile.js
- Added 'create-profile' route to routes.js
- Added new image asset (.svg) from Figma to 'create-profile-images' folder in 'assets' folder.
- New Location Marker icon.
- Added transparent top-border to BaseInput.js component to prevent page movement on hover. 
- Installed 'typeface-dm-sans' package in client to display DM Sans font. 

Resolves #176


client/src/components/Input/Checkbox.js
```diff
@@ -3,12 +3,12 @@ import styled from "styled-components";
...
     border-color: #646464;
-    width: 2rem;
-    height: 2rem;
+    width: ${(props) => props.size || "2rem"};
+    height: ${(props) => props.size || "2rem"};
   }
   .ant-checkbox-checked .ant-checkbox-inner {
     background-color: ${(props) => props.color || "#6076ef"};
-    border-color: #939393;
+    border-color: ${(props) => props.color || "939393"};
```
silenceway, line 11: Update color to DARK GRAY. Not sure the background color with the constants list.

client/src/pages/CreateUserProfile.js
```diff
@@ -0,0 +1,283 @@
...
+        font-family: ${props => props.font};
+    }
+`
+
+const CheckboxGroup = ({children, description, font, label}) => {
+    return (
+    <CheckboxContainer font={font} label={label} description={description}>
+        <Checkbox 
+        color={theme.colors.royalBlue} 
+        size={theme.typography.size.xxlarge}
```
robinv85, line 101: indentation seems off? I'd think this is auto fixed by ESLint & Prettier?

client/src/pages/CreateUserProfile.js
```diff
@@ -0,0 +1,283 @@
...
+    }
+    font-size: ${theme.typography.size.small};
+`;
+
+const Submit = styled(SubmitButton)`
+    font-weight: 500;
+`
+const CreateProfile = () => {
+useEffect(() => {
+    let header = document.querySelector('.header');
```
robinv85, line 220: why do you need query selectors and using JS to update the styles? 
probably better to avoid this, and use styled components? 
if the changes are for parent components, probably this needs to be configured with the "template" on the route

robinv85:
Hey @T-Allen thanks for this PR and if it's ok with you I'll finish this with a layout template and adding the react-hook-form so I can continue on the onboarding flow...

client/src/components/Input/Checkbox.js
T-Allen: Yeah, I went with the exact color from Zeplin which wasn't in our theme. Not sure if it's worth it to add this exact color to the theme or just go with the colors.DARK_GRAY?

client/src/pages/CreateUserProfile.js
T-Allen: Yes, I agree this is definitely better done with a layout template. It was a dirty way to remove the margin on a parent element and modify the toolbar to appear the same as in the design.

T-Allen:
> Hey @T-Allen thanks for this PR and if it's ok with you I'll finish this with a layout template and adding the react-hook-form so I can continue on the onboarding flow...

Sounds great, thanks for working on this. Let me know if you need anything.

robinv85:
I've added a layout template "navless" and made the styles and components more consistent with the login / signup form, all the forms and layouts should probably be checked to improve the codebase.

I also noticed an issue with the label component which contained a `p` element, seemed wrong and it wasn't working on the forms so fixed that. 

I think it should be ready for re-review, I would like to merge this and work on related issues in separate pull requests:
- add the react-hook-form library to finish the integration with the API
- fix the redirection logic so it works correctly in all cases (when you can and can't redirect to certain pages based on what the API returns: if user exists in databse, email confirmed etc.)

T-Allen:
Cool, thanks! I'm noticing a few issues with the styling of the text inputs.
- The border-width appears stuck in the hover state. 
- The labels and icon for the address input are missing, and the overall style is not consistent with the other inputs.
- The inputs move the entire form slightly when hovered.  
image

T-Allen:
Also, I see your comment addressing this, but the heading alignment is centered, while it is left-aligned and somewhat off-set in the design. 
image

T-Allen:
One other small issue I'm seeing, is that the form's container is not filling the viewport height and exposing the default white background.
image

robinv85:
@T-Allen yeah thanks for the feedback, I still have to fix the merge conflicts and finish some of the details, I was trying to make this more consistent with login / signup page but I think we should probably make a ticket to review all of the similar forms and layouts and make the code more reusable and consistent...
I think the design for this header was wrong, all the other similar pages had it centered...
Also the placeholders aren't really clear in the design, some inputs show a placeholder and other inputs are filled so it's not clear if they should have a placeholder besides the label, it looks more like unfinished designs with some dummy text.
I will look into these issues now, got a meeting too in a few minutes with engineering & porduct, will maybe ask to open a ticket to fix the consistency and make sure the designs are a bit more detailed

robinv85:
@T-Allen where do you see this form container issue with viewport height (what browser and resolution?)
I'm going to make another ticket to fix the containers for the whole app, i think we might be able to reuse this layout more, needs to be checked
Also the border-bottom issue, it's still moving 1 pixel on hover but i didn't want to change the base component in this PR, might break other things, so i would also suggest handling this in another ticket for consistent form components.

robinv85:
Going to merge this and finish remaining details in the other related tickets so we can make some progress.

T-Allen:
> @T-Allen where do you see this form container issue with viewport height (what browser and resolution?)
> I'm going to make another ticket to fix the containers for the whole app, i think we might be able to reuse this layout more, needs to be checked
> Also the border-bottom issue, it's still moving 1 pixel on hover but i didn't want to change the base component in this PR, might break other things, so i would also suggest handling this in another ticket for consistent form components.

Sounds good. I noticed the container issue in Chrome. My display resolution is 1920x1080.

vikjoshi:
@Chimaobi-Johnson , FYI for Create Org Profile

Issue #355, opened 2020-05-15
Symptom Checker- Desktop- Multi-choice selection logic


vikjoshi:
For Question "If so, have you traveled to an area..." and "Accordingly to what you know...": 
Expected Behavior: User can select multiple options like in earlier questions 
Actual Behavior: User can only select one option

snowiesuet:
Not sure if this is a bug or intended. Needs clarification from product team.

vikjoshi:
@snowiesuet , these are the two questions I'm referring to- it looks like from the Design that they should be multi-choice selections but they're single choice. 
Staging: 
![image](https://user-images.githubusercontent.com/41554004/82157950-816a3300-9839-11ea-95b7-5828397812e8.png)
![image](https://user-images.githubusercontent.com/41554004/82157960-8dee8b80-9839-11ea-8b31-ad0dc8769652.png)


Figma: 
![image](https://user-images.githubusercontent.com/41554004/82157974-a5c60f80-9839-11ea-94b9-e0a9bfdf74fc.png)

snowiesuet:
@vikjoshi  Ah I see now, I was looking at the mobile version and all of them don't have checkboxes by default. Apparently only desktop version has the checkboxes.

![image](https://user-images.githubusercontent.com/3907305/82160391-fc940f00-9862-11ea-9c2e-f4374e36b179.png)

snowiesuet:
Waiting for #374 to be merged to work on this

robinv85:
Should we confirm this with designers if this is correct? It's a bit confusing to me. At first we had checkboxes / radio buttons, then we removed it... now it seems only desktop?

vikjoshi:
@robinv85 @snowiesuet , I checked the requirements doc and for both desktop and mobile, those should be multi-selection with checkboxes/radio buttons so the desktop Figma has it correct. I asked the Design team to update the design so the mobile will match.

vikjoshi:
Just to further clarify, the selection logic for these questions is similar to Question 2 and 3 in the Symptom checker where the user can select multiple options OR select "None of the above/I'm not sure/they've had no exposure" which unselects any other selected option

snowiesuet:
@vikjoshi thank you for the clarification. Linking this to #357 so we can refactor this component.

robinv85:
Alright, @vikjoshi maybe talk to @agmaal1 who is also looking at the StepWizard components to make them more reusable, so we avoid merge conflicts etc.

[...]

PR #435, opened 2020-05-21
#355 Change single option answer to multiple options for question 6 

snowiesuet:
#355 This should change the logic for question 6 from single option answer to multiple options. 
I've also edited some of the styles that don't look right in Symptoms Checker.

This component is getting too large and we **definitely** need to refactor in the future.


client/src/pages/SymptomsCheck.js
```diff
@@ -200,10 +203,21 @@ const AnswerButton = ({ children, onSelect }) => {
...
     <AnswerStyles onClick={onSelect} className={checked && "selected"}>
-      <span>
-        {text}
-        {!!content && <h6>{content}</h6>}
-      </span>
+      <div>
+        <TextLabel size={theme.typography.size.mediun} block="true">
+          {text}
+        </TextLabel>
+        {!!content && (
```
robinv85, line 210: don't think you need `!!` here? 
I guess if you just omit it, it will also turn  `expression && ...` into a Boolean (falsy / truthy)

client/src/pages/SymptomsCheck.js
```diff
@@ -248,7 +262,7 @@ const Welcome = (props) => {
     </WizardStep>
   );
 };
-
+//all steps require refactor
```
robinv85, line 265: not sure what the purpose if this comment is, as a general rule I think it's best to avoid comments that don't really add value and aren't strictly necessary (prefer self-describing code that doesn't need explanation if possible)

client/src/pages/SymptomsCheck.js
snowiesuet: Good catch! This was already there before and I overlooked it. Removed falsy now.

client/src/pages/SymptomsCheck.js
snowiesuet: Yea, sorry about that. This is more of a reminder for enhancement in the future. Removed!

client/src/pages/SymptomsCheck.js
```diff
@@ -481,10 +495,24 @@ const STEP_6_ANSWERS = [
...
+};
+
 const Step6 = (props) => {
-  const onSelectAnswer = (answer) => {
-    props.update("exposureLast14Days", answer);
-    props.nextStep();
+  const [state, updateState] = useState(STEP_6_STATE);
+  const { answers, none } = state;
+  const toggleAnswer = (answer) => {
+    const updatedAnswers = { ...answers, [answer]: !answers[answer] };
```
anniegiang, line 507: For now it looks good but maybe we can use `useReducer` in the future to handle complex state

Issue #442, opened 2020-05-22
[Mobile] Correct 2 scenarios when trying to save FP to home screen


tiagocostasilva:
Hi Team,
i noticed that when i open FP (https://staging.fightpandemics.work/ ) on my mobile phone and i try i go to this menu and click on "Add to Home screen":
![image](https://user-images.githubusercontent.com/45389561/82709283-22148480-9c78-11ea-871c-7798a9eabc3f.png)

The following appears:
![image](https://user-images.githubusercontent.com/45389561/82709323-3eb0bc80-9c78-11ea-8358-305c1f9da493.png)

Can we change both the text and the icon to our favicon ?

Also, after adding to the home screen and trying to open the app from there, before our landing page i see a blank page with a text on the center saying "Create React App Sample". Difficult for me to capture this, but easily replicate following the steps here.
Can you also check it and correct it please ?

Thanks,
Tiago

stavares843:
title fixed here

https://github.com/FightPandemics/FightPandemics/pull/443


> Also, after adding to the home screen and trying to open the app from there, before our landing page i see a blank page with a text on the center saying "Create React App Sample". Difficult for me to capture this, but easily replicate following the steps here.


instead of "Create React App Sample" what should it say?

tiagocostasilva:
Waiting on some opinions from the community here: https://fightpandemics.slack.com/archives/C010QR5D4KY/p1590183606401800

stavares843:
@TiagoaTavares any more suggestions there? should we just say ''Loading''?

tiagocostasilva:
@tiagocostasilva  you mean? :D Just left a comment on the channel (same post) to get help from our rockstar designers

tiagocostasilva:
Pending designer task to start : https://www.notion.so/fightpandemics/Loading-FP-logo-screen-before-the-app-1d444471f971472b9ca8a0ef41ef3c61

stavares843:
edited

tiagocostasilva:
@stavares843 the designs are on notion: https://www.notion.so/fightpandemics/Loading-FP-logo-screen-before-the-app-1d444471f971472b9ca8a0ef41ef3c61
Is there any use case where you can see this is useful ?

stavares843:
still missing fixing the logo

tiagocostasilva:
@stavares843 any update on this? :)

stavares843:
still missing the logo

tiagocostasilva:
[...]

PR #443, opened 2020-05-22
fix title

stavares843:

fix title

https://github.com/FightPandemics/FightPandemics/issues/442

### DIFF
--- a/client/public/manifest.json
+++ b/client/public/manifest.json
@@ -1,5 +1,5 @@
 {
-  "short_name": "React App",
+  "short_name": "FightPandemics",
   "name": "Create React App Sample",
   "icons": [
     {
--- a/client/public/manifest.json
+++ b/client/public/manifest.json
@@ -1,6 +1,6 @@
 {
   "short_name": "FightPandemics",
-  "name": "Create React App Sample",
+  "name": "Loading",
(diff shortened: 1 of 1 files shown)


client/public/manifest.json
```diff
@@ -1,5 +1,5 @@
 {
-  "short_name": "React App",
+  "short_name": "FightPandemics",
   "name": "Create React App Sample",
```
anniegiang, line 3: Let's say "FightPandemics" here too

stavares843: https://github.com/FightPandemics/FightPandemics/issues/442#issuecomment-632919293

anniegiang: Do you think we should merge for now? Or should we wait for more inputs?

stavares843: will check if there's any more suggestions

robinv85, line 3: and "name" ?

stavares843: https://github.com/FightPandemics/FightPandemics/issues/442#issuecomment-632919293

stavares843: updated

stavares843: updated

PR #445, opened 2020-05-23
Add photo and liked to /posts endpoint and update userId usage count

### LINKED ISSUE #437: Add missing fields to post endpoint.
Please add to the GET /api/post these fields:
* liked (if logged in user liked the post).
* user avatar if the user has a photo and the full name to create a text avatar if the user doesn't have a picture.

Naraujo13:
# What does this pull request aim to achieve? 

Add `photo` field at `Author` subdocument stored at `Post`.
Add boolean field `liked` to `/posts` endpoint`, representing if the authenticated user liked that post or not.

Closes #437 

- Add `photo` field at `Author`
- Add `liked` field at `/posts` endpoints
- Update userId usage to use auth generated req.userId on all endpoints
- Update all ObjectId comparisons to use equals()


backend/lib/endpoints/posts.js
```diff
@@ -276,12 +276,12 @@ async function routes(app) {
       schema: deletePostSchema,
     },
     async (req) => {
-      const { userId } = req.body;
+      const { userId } = req;
       const { postId } = req.params;
       const [findErr, post] = await app.to(Post.findById(postId));
       if (findErr) {
         throw app.httpErrors.notFound();
```
robinv85, line 283: maybe out of scope here but wonder if the `findErr` should be logged instead of just returning  "not found" (I think I've mentioned this before on another PR, this occurs multiple times in similar checks)

joshmorel: Also out of scope but this logic (not just here but across the app) seems a bit incomplete.

```
      const [findErr, post] = await app.to(Post.findById(postId));
      if (findErr) {
        throw app.httpErrors.notFound();
```
If I pass an invalid ObjectId there is an error - like posts/1

BUT if I pass a valid ObjectId that doesn't exist both are null so I get, for example:

image

should it be `if (findErr || post === null)` ?

Naraujo13: Good point. 
What you say makes sense @joshmorel
Did you open an issue for that?

robinv85:
Ok I guess according to new rules we can merge it when second review, but we should make sure if it's approved it was a good review.
Also, please fix merge conflicts @Naraujo13

PR #446, opened 2020-05-23
Add support for Cypress framework and first e2e test using Cypress - to the cypresse2e branch

galeza:
The same as #424 
Branch is cypress, see if that 'create link' check will not be triggered
Detailed instruction can be found here: https://www.notion.so/fightpandemics/Instruction-how-to-instal-and-use-Cypress-bc6b0443f7aa4e93bbbd092d1635b7a6
There are no merge conflicts.


galeza:
Post Review Build URL was triggered.

PR #447, opened 2020-05-23
Add support for Cypress framework and first e2e test using Cypress - from cypresse2e to cypresse2e branch

galeza:
same changes as #424
From fork cypresse2e to central cypresse2e2

Detailed instruction can be found here: https://www.notion.so/fightpandemics/Instruction-how-to-instal-and-use-Cypress-bc6b0443f7aa4e93bbbd092d1635b7a6
There are no merge conflicts.


galeza:
Still error from fork?

PR #460, opened 2020-05-24
Add Sentry error notifier

### LINKED ISSUE #387: Set up Sentry error reporting
Set up Sentry projects for each environment: review, staging, and production. 

* Update backend code to log errors to Sentry
* Add Sentry API keys to AWS SSM Parameter Store
* Update `main.tf` Terraform to pass in Sentry environment variables

mannykary:

Adding Sentry error notifier so that we can get alerted of Sentry errors in staging/production.

Summary of changes:
- Added Sentry npm package
- Added Fastify error handler plugin
- Updated Terraform/Github workflows to pass in environment variables needed for Sentry

Close #387.

### DIFF
--- a/backend/lib/plugins/error-handler.js
+++ b/backend/lib/plugins/error-handler.js
@@ -32,6 +32,7 @@ function errorHandler(app, config, done) {
     Sentry.init({
       dsn: config.url,
       environment: config.environment,
+      release: config.release,
     });
     errorHandler = errorNotifierHandler;
   }
--- a/backend/lib/plugins/error-handler.js
+++ b/backend/lib/plugins/error-handler.js
@@ -1,0 +1,42 @@
+const fp = require("fastify-plugin");
+const Sentry = require("@sentry/node");
+
+// Default Fastify error handler
+// https://github.com/fastify/fastify/blob/5bc406d4c6677131237f85b6ae69fe9e34d4b8dd/lib/context.js#L29
+function defaultErrorHandler(error, request, reply) {
+  if (reply.statusCode >= 500) {
+    reply.log.error(
+      { req: request, res: reply, err: error },
+      error && error.message,
+    );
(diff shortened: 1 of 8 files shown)


backend/lib/plugins/error-handler.js
```diff
@@ -0,0 +1,43 @@
+const fp = require("fastify-plugin");
+const Sentry = require("@sentry/node");
+
+// Default Fastify error handler
+// https://github.com/fastify/fastify/blob/5bc406d4c6677131237f85b6ae69fe9e34d4b8dd/lib/context.js#L29
+function defaultErrorHandler(error, request, reply) {
```
robinv85, line 6: isn't this also conflicting with the `fastify-sensible` plugin which also includes an override for the default error handler?

backend/lib/plugins/error-handler.js
```diff
@@ -0,0 +1,43 @@
...
+  if (reply.statusCode >= 500) {
+    reply.log.error(
+      { req: request, res: reply, err: error },
+      error && error.message,
+    );
+  } else if (reply.statusCode >= 400) {
+    reply.log.info({ res: reply, err: error }, error && error.message);
+  }
+  reply.send(error);
```
robinv85, line 15: this will put back the original default handler which was overridden by `fastify-sensible` and will not return all internal error server error info to the client which is something we probably want to avoid

backend/lib/plugins/error-handler.js
mannykary: I didn't realize that `fastify-sensible` has an override for the default error handler. i will fix this to use their error handler.

backend/lib/plugins/error-handler.js
mannykary: I've changed this to the fastify-sensible error handler.

PR #500, opened 2020-05-27
Fix bug in error handling

DanielAlexander84:
I was having an issue with docker a proxy request error while trying to send requests to the backend from the client. After talking with @mannykary he discovered that there was a bug in the new error plugin he installed. This fixes that bug.

### DIFF
--- a/backend/lib/plugins/error-handler.js
+++ b/backend/lib/plugins/error-handler.js
@@ -5,6 +5,8 @@ const Sentry = require("@sentry/node");
 // https://github.com/fastify/fastify-sensible/blob/5eec059b2a77579f75bafc91e12f60fb90b6dab5/index.js#L46
 function defaultErrorHandler(error, request, reply) {
   if (
+    reply &&
+    reply.raw &&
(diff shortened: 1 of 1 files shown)


backend/lib/plugins/error-handler.js
```diff
@@ -5,6 +5,8 @@ const Sentry = require("@sentry/node");
 // https://github.com/fastify/fastify-sensible/blob/5eec059b2a77579f75bafc91e12f60fb90b6dab5/index.js#L46
 function defaultErrorHandler(error, request, reply) {
   if (
+    reply &&
+    reply.raw &&
     reply.raw.statusCode === 500 &&
```
robinv85, line 10: what you could also do to make this shorter, not sure if it works in Node 12: `reply?.raw?.statusCode` but I think this currently only works in the frontend

mannykary: It only works in Node 14: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

PR #511, opened 2020-05-28
Fixed search bar border and replaced navigation icon to match the style on figma

omarzoghayyer:
…e on figma

- [x] Fixed share my-location icon and replaced it to match figma styling 

- [x] replaced the search bar text from city, neighborhood, to zipcode, city, state"

- [x] changed the font size of share my location to small from medium.
TODO

- [x] Change the search bar icon inside the search bar to match figma as shown below. 
Screen Shot 2020-05-28 at 2 31 22 PM
 


snowiesuet:
Are waiting for the new search icon from design? @omarzoghayyer

omarzoghayyer:
@snowiesuet Im actually stuck at changing that icon. Its not allowing me through prefix.

snowiesuet:
Let's come back to it in the future if there is no easy fix for now. @anniegiang @robinv85 what do you guys think?

anniegiang:
> @snowiesuet Im actually stuck at changing that icon. Its not allowing me through prefix.

Hmm that's strange since it seems work as the doc suggests. Can you should me the code for it? But I agree we should probably keep the existing icon and move on if it's taking a while

client/src/components/Input/LocationSearch.js
```diff
@@ -26,15 +31,15 @@ const LocationSearch = () => {
       <WhiteSpace />
       <StyledSearchBar
```
anniegiang, line 32: `prefix={<SvgIcon />}` doesn't work?

Btw, `SvgIcon` is a component we're using for SVG icons. You can check out the implementation or reach out to @nadinejuraschek who created it

robinv85:
> 
> 
> Let's come back to it in the future if there is no easy fix for now. @anniegiang @robinv85 what do you guys think?

well... it's kind of the subject of the PR so probably best to try to fix it, like Annie suggested for example a solution, I didn't have time yet to check but there is probably some kind of solution possible.

client/src/components/Input/LocationSearch.js
omarzoghayyer: That didn't work.

omarzoghayyer: This is the actual folder that they used here. It doesn't look like we can change the icon. https://mobile.ant.design/components/search-bar/#components-search-bar-demo-basic

omarzoghayyer:
@robinv85 @anniegiang I was able to change it to this status by using Input instead of SearchBar.
It seems like whoever used the search bar got it from here [Link] which doesn't indicate change of icon. But when I used this [Link]. the input field allowed me to add an icon.

What are your suggestions?


(https://mobile.ant.design/components/search-bar/#components-search-bar-demo-basic)

(https://mobile.ant.design/components/search-bar/#components-search-bar-demo-basic
)
Screen Shot 2020-05-29 at 3 30 53 PM

anniegiang:
> @robinv85 @anniegiang I was able to change it to this status by using Input instead of SearchBar.
> It seems like whoever used the search bar got it from here [Link] which doesn't indicate change of icon. But when I used this [Link]. the input field allowed me to add an icon.
> 
> What are your suggestions?
> 
> (https://mobile.ant.design/components/search-bar/#components-search-bar-demo-basic)
> 
> (https://mobile.ant.design/components/search-bar/#components-search-bar-demo-basic
> )
> **This is how it looks now with different field.**
> Screen Shot 2020-05-29 at 3 39 43 PM

It looks like the `SearchBar` component is for mobile screens so it probably has difference usage than `Input`. It looks good now so I think we can keep the current changes.

omarzoghayyer:
@anniegiang then its ready to be merged.

PR #517, opened 2020-05-28
Feature/461 create user form + API integration

### LINKED ISSUE #461: Integrate API with create-user-profile form
Finish the form & hook up with the API to create the user profile after registration.
Redirect to this page after a user has created an account (and verified email if needed) but hasn't entered the required user information yet to create a user in the database.
When user created successfully it should also show the next screen.

The form should have client-side validation and display error messages coming from the server as well.

The location component is quite complex and there is some discussion about this so can be considered out of scope to completely finish this, can be done in a separate ticket.

robinv85:

Closes #461 
Adds API integration to create a user, with current version of API 
(some fields are missing in API, made those form components hidden for now, and the location input will also be handled in a separate issue because it's quite complex).


galeza:
Hi @robinv85 

I did four tests. 
Test1: correct email and password
1. I received email after Signup 
2. I confirmed email
3. Then in the application - I was correctly redirected from Login screen to the Create User Profile screen.

Test2: Incorrect email: xyz@gmail.commmmmmmmmmmmmmmmmmmmm, correct password on the SignUp screen
1. **Should we have validation for email string length in this PR?** 
2. Redirected to the Please verify your email screen
3. **What should happen on the Please verify your email screen when I click FightPandemics logo?** Currently logo is disabled.

Test 3 Correct email, weak password 
1. Correct validation: Weak password. 

Test4 User exists in DB
1. correct validation - User already exists

robinv85:
Hi @galeza thanks for the help! 

To answer your question about email validation, this validation should be done for signup & login, but this PR is for "create user form" so the email is currently read-only for create-user and will automatically be used from the signup / social media authentication step.

There aren't very clear requirements for the validation at the moment, for e-mails we have a valdate function but I think it uses a regex only. For password it can set minimum characters only on the client-side I think. This can definitely be improved but is out of scope here.

Also, "Please verify your email" screen is out of scope of this PR.

joshmorel:
@galeza let's add another issue for email format validation on signup. This should be simple - just an addition to fluent-schema to use email format (which covers length & format).

@robinv85 I went through the "happy path" with both email & social authentication. Email worked, with social (Google) I get this - no email and email/create profile disabled.

I tried to go through the code to figure out what was going on but I haven't been in the front-end much so I couldn't do so easily.

image

I'll take a deeper dive later.

robinv85:
Ah @joshmorel good point, I actually forgot to take the email from the social accounts as well! Will fix that now

robinv85:
@joshmorel ok I've fixed the social login + profile creation as well, had to return it from backend oauth endpoint

robinv85:
Ok thanks for the feedback... only some details I wasn't sure like 
- form compoents: I used BaseInput and not TextInput or FormInput but I'm not sure where else those are used so I made some changes and (temporarily) put some styles in the constants/formStyles and maybe it can be refactored later to the other components for many of the forms
- the redux & useReducers dispatches are a bit confusing at the moment, probably best to clean that up too

But both of these tickets have separate issues to do later, so I will merge this now.

PR #526, opened 2020-05-29
Fix IndividualUser first and last name setters

### LINKED ISSUE #512: Don't (re)set value in IndividualUser setters
@joaofnds missed this when doing PR review but noticing now when troubleshooting the import script for the new user/org models.

We should have it like `OrganizationUser` just return the value don't set e.g. `this.firstName = firstName` within the setter. Otherwise the setter is recalled (again & again...). Mongoose somehow prevents a loop from running endlessly but it's still called many times so to be avoided.

joaofnds:

Fix Individualuser first and last name setters. Closes #512 

### DIFF
--- a/backend/lib/models/IndividualUser.js
+++ b/backend/lib/models/IndividualUser.js
@@ -5,18 +5,25 @@ const { updateAuthorName: updateCommentAuthorName } = require("./Comment");
 
 const INDIVIDUAL_USER_TYPES = ["individual"];
 
-function updateAuthorFirstName(firstName) {
-  this.firstName = firstName;
+function fullName(firstName, lastName) {
+  return `${firstName} ${lastName}`;
+}
 
-  this.updateAuthorNameReferences();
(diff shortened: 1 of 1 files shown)


joshmorel:
Hi @joaofnds thanks this addresses the issue I opened but I've noticed something else. I doesn't seem updating the references works through SchemaType setters because the update is asynchronous while the setters are synchronous. I change org `'name: 'Sourced by FightPandemics - Community'` to `'name: Sourced by FightPandemics (from Community)`, the `updateAuthorName` function does execute but the post.author.name isn't actually updated.

I think this would be easier to troubleshoot, however, once we have all the endpoints working - create/update post as organization; update individual user. So I can open the issue but let's tackle before launch but after the preceding are done.

What's your suggestion @Naraujo13?

As a resolution, maybe we need to use post-save middleware instead? https://mongoosejs.com/docs/middleware.html

joshmorel:
I'm going to close this as we're solving this whole syncing thing a different way

PR #559, opened 2020-06-03
Added email-validation to needs help page.

omarzoghayyer:

Email validation to needs help page. Still unfinished. 

### DIFF
--- a/client/src/pages/NeedHelp.js
+++ b/client/src/pages/NeedHelp.js
@@ -1,11 +1,8 @@
 import React, { useState, useEffect } from "react";
 import { withRouter } from "react-router-dom";
-import { Link } from "react-router-dom";
-import { Input } from "antd";
-import { theme } from "constants/theme";
-import styled from "styled-components";
+
 import SubmitButton from "components/Button/SubmitButton";
-import { asyncGetGeoLocation } from "utils/geolocation";
+import InputError from "components/Input/InputError";
 import {
   AnswerButton,
   StyledWizard,
@@ -19,7 +16,8 @@ import {
   WizardFormWrapper,
   WizardFormGroup,
 } from "components/StepWizard";
-const { white, lightGray } = theme.colors;
+import { asyncGetGeoLocation } from "utils/geolocation";
+import { validateEmail } from "utils/validators";
 
 const INITIAL_STATE = {
   answers: [],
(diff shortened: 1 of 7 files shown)


omarzoghayyer:
@robinv85 @anniegiang Users won't be able to pass by and submit empty email, but I have an 2 issues with this branch.

1. - [ ] It doesn't allow me to type anything.

2. - [ ] It gives me errors as soon as you start typing in the email field.

Screen Shot 2020-06-02 at 8 53 15 PM

client/src/pages/NeedHelp.js
```diff
@@ -108,6 +131,15 @@ const Step3 = (props) => {
     </WizardStep>
   );
 };
+function validate(values) {
+  let errors = {};
+  if (!values.email) {
+    errors.email = 'Email address is required';
+  } else if (!/\S+@\S+\.\S+/.test(values.email)) {
```
BMariscal, line 138: There's an existing email validator utility function in `client/src/utils/validators.js` that you can use here. I have a PR [here](https://github.com/FightPandemics/FightPandemics/pull/560/files#diff-c22aa1881b72d5dada9129049a0a36ed) that's going to make some changes to it.

omarzoghayyer: Thats great, but we still need to customize the error message for these emails.

BMariscal: The utility function returns a Boolean, so instead of implementing your own regex here, call the utility function to validate the email and raise the error string you already have. Implementing your own regex here to validate emails duplicates existing functionality.

robinv85: Good points @BMariscal 
Seems like the regex comes from Auth0 which is good because we want this consistent with backend, API validation and frontend everywhere https://github.com/auth0/lock/blob/4aaefa4a9ed00423e665e0743a7b0a952c6a267d/src/field/email.js#L7 

The react-hook-form library also has support for regular expressions to validate fields but maybe it's better to use the function everywhere since I see now it also adds lowercasing which I'm not sure would be handled by that library.

robinv85:
@omarzoghayyer is this PR replacing the older one #329 ?

omarzoghayyer:
@robinv85  we can either use any one of them. The previous validation method is cool. This method is as you requested before not a "pop up notification" Let me know which one would you like us to go forward with so I can make the changes.

client/src/pages/NeedHelp.js
```diff
@@ -108,6 +145,15 @@ const Step3 = (props) => {
     </WizardStep>
   );
 };
+function validate(values) {
+  let errors = {};
+  if (!values.email) {
+    errors.email = 'Email address is required';
```
anniegiang, line 151: Need to use `setErrors` to update state

client/src/pages/NeedHelp.js
```diff
@@ -108,6 +145,15 @@ const Step3 = (props) => {
     </WizardStep>
   );
 };
+function validate(values) {
```
anniegiang, line 148: Let's use variable declaration for function statements to stay consistent

client/src/pages/NeedHelp.js
anniegiang: Haven't worked with regex much but I ran into an [post](https://www.loggly.com/blog/regexes-the-bad-better-best/) that talks about how regex can have poor performance. Maybe it's not a worry for us?

robinv85:
@omarzoghayyer yes if the other one is with the pop up then it should probably be closed

robinv85:
@omarzoghayyer another question, so this is a problem here that you can't type fast?

omarzoghayyer:
@robinv85 the email validation on offer help isn't a pop up. We can change it to this method as well. 
& yeah you cant type fast. its like 1 letter per 1 second.

client/src/pages/NeedHelp.js
```diff
@@ -74,11 +78,36 @@ const Step2 = (props) => {
...
+  const onChange = (event) => {
+    event.persist();
+    setValues(values => ({
+      ...values,
+      [event.target.name]: event.target.value
+    }));
+  };
+
+  const onSubmit = (event) => {
+    if (event.preventDefault());
```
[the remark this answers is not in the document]
robinv85: Yeah never seen this before, not sure if that returns something

client/src/pages/NeedHelp.js
robinv85: yeah that's also a good point, from my experience regexes can also get very complex and it's best not to try to do too much with them... 
as far as performance, it also depends how it's used, if the regex is only called to validate once, not too bad, but if it fires for every key stroke etc. then you probably have to optimize that too, e.g. with a debounce

PR #560, opened 2020-06-03
[537] Email validation fix - validate email top-level domains on sign-up

### LINKED ISSUE #537: [Desktop][Mobile] SignUp | email validation
env: https://staging.fightpandemics.work/

Steps: 
1. Open Sign Up screen
2. In the email input: xyz@gmail.commmmmmmm(X number of letters)
3. Enter correct password and repeat password.
4. Press Sign Up button

No validation for incorrect email

BMariscal:
## Previously:
Our email validation did not check whether the email's top-level domain was valid. 
As described in #537, any top-level domain of length >= 2 was valid. 

## Now:
The user can submit the sign-up form if their email contains a valid top-level domain registered with [ICANN](https://www.icann.org/resources/pages/tlds-2012-02-25-en).

The `tlds` package contains an up-to-date list of valid top-level domains scraped from ICANN.
 You can check out the full list of top-level domains [here](https://github.com/stephenmathieson/node-tlds/blob/master/index.js). 

Closes #537 
##

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -63,9 +63,6 @@
     "lint-staged": "^10.2.0",
     "prettier": "^2.0.5"
   },
-  "optionalDependencies": {
-    "fsevents": "*"
-  },
   "husky": {
     "hooks": {
       "pre-commit": "lint-staged"
--- a/client/package.json
+++ b/client/package.json
@@ -28,6 +28,7 @@
     "snyk": "^1.319.2",
     "styled-components": "^5.0.1",
     "super-tiny-icons": "^0.3.2",
+    "tlds": "^1.207.0",
     "typeface-poppins": "^0.0.72",
     "typeface-work-sans": "^0.0.72"
   },
@@ -61,6 +62,9 @@
     "husky": "^4.2.5",
     "lint-staged": "^10.2.0",
     "prettier": "^2.0.5"
+  },
+  "optionalDependencies": {
+    "fsevents": "*"
   },
   "husky": {
     "hooks": {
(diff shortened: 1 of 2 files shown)


robinv85:
@BMariscal @OmarIbrahim991 
I'm also working on client-side validation, currently for login, signup and create-user forms
These forms are using https://react-hook-form.com/ 
To show a email validation message I can currently just do this:
`return validateEmail(email) || "Invalid email" ` 
But it could also be interesting to return specific errors... well anyway just wondering what you think about this? I know @omarzoghayyer was looking at custom error messages, I think that's also useful for places where we don't need the full form library functionality

BMariscal:
@robinv85 I checked out `react-hook-form` and unfortunately it's not validating top-level domains D: Seems like something it should be validating. I'll open an issue in https://github.com/react-hook-form/react-hook-form.  

wow, I just checked a couple of website's sign-up forms, and neither validates top-level domains!! Hmmm.

robinv85:
@BMariscal alright, I guess for now just showing a generic "invalid email" is ok.

Otherwise you could return specific error messages from the `validateEmail` function
React-hook-form uses the string messages to display them in the `errors` object
[So I added this here ](https://github.com/FightPandemics/FightPandemics/commit/6db9cb2e32b66ec2c2693ad78327772fab9d6f85#diff-9f8a9582a00efc037203fb7737fe80fdR304)

So that's why I wanted to check, because in other places I think @omarzoghayyer worked on email validation as well without that library and he might not expect string messages from the validate function.

Issue #520, opened 2020-05-29
[WEB] Scroll bar in a post


ppagarawal:
While creating a new post, a new dialog box appears to input the data from the users. This dialog box has a scroll bar as well. The scroll bar is not required and must not be on the screen.

![image](https://user-images.githubusercontent.com/6276256/83230599-a6e72d00-a1a7-11ea-9d8c-d949b32e9c8d.png)

![image](https://user-images.githubusercontent.com/6276256/83230758-f9c0e480-a1a7-11ea-803b-115e1b236cb5.png)

Please let me know in case of questions.

CeesWang:
add overflow-y to the container to hide the scrollbar

stavares843:
whats the browser? IE?

stavares843:
https://github.com/FightPandemics/FightPandemics/pull/580/

PR #580, opened 2020-06-03
hide scroll bar

stavares843:

https://github.com/FightPandemics/FightPandemics/issues/520

### DIFF
--- a/client/src/components/CreatePost/CreatePost.js
+++ b/client/src/components/CreatePost/CreatePost.js
@@ -105,6 +105,7 @@ const Wrapper = ({ onCancel, visible, children }) => {
       footer={null}
       visible={visible && currentStep !== 3}
       destroyOnClose={true}
+      overflow-y: hidden;
       closeIcon={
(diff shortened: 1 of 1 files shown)


client/src/components/CreatePost/CreatePost.js
```diff
@@ -105,6 +105,7 @@ const Wrapper = ({ onCancel, visible, children }) => {
       footer={null}
       visible={visible && currentStep !== 3}
       destroyOnClose={true}
+      overflow-y: hidden;
```
robinv85, line 108: ok can't believe I didn't see this but this is actually CSS code on a react element (JSX) attribute... so it needs to go into the styles, currently making the frontend build fail

stavares843: https://github.com/FightPandemics/FightPandemics/pull/597

robinv85:
@stavares843 it's ok I got a fix i will push it in a minute just building to see if it works

PR #582, opened 2020-06-03
Client-side validation for login/signup and create user profile

### LINKED ISSUE #369: Client Side Error Handling for Web in User Authentication Flow

robinv85:
Closes #369 
add client-side validation for login/signup and create user profile pages
integrated with the react-hook-form errors, currently hard coded validation messages, they can be replaced by translation strings later

### DIFF
--- a/client/src/components/Input/BaseInput.js
+++ b/client/src/components/Input/BaseInput.js
@@ -4,9 +4,10 @@ import { theme } from "constants/theme";
 const { darkGray, darkerGray, primary, red, royalBlue } = theme.colors;
 
 const BaseInput = styled.input`
-  border: unset;
+  border: none;
   border-bottom: ${(props) =>
     props.disabled ? "1px solid " + darkGray : "1px solid " + primary};
+  box-shadow: none;
   color: ${(props) => (props.disabled ? darkGray : darkerGray)};
   transition: 150ms border;
 
--- a/client/src/components/Input/BaseInput.js
+++ b/client/src/components/Input/BaseInput.js
@@ -1,7 +1,7 @@
 import styled from "styled-components";
 import { theme } from "constants/theme";
 
-const { darkGray, darkerGray, primary, royalBlue } = theme.colors;
+const { darkGray, darkerGray, primary, red, royalBlue } = theme.colors;
 
 const BaseInput = styled.input`
   border: unset;
@@ -10,6 +10,10 @@ const BaseInput = styled.input`
   color: ${(props) => (props.disabled ? darkGray : darkerGray)};
   transition: 150ms border;
 
+  &.has-error {
+    border-bottom: 1px solid ${red};
+    color: ${red};
+  }
   &:focus,
   &:hover,
   &:active {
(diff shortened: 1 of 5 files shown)


galeza:
Hi @robinv85 

I tested SignUp/SignIn scenario:
1. Empty email
2. Empty password
3. Empty confirmation password
4. Too short password
5. Passwords don’t match
6. Password length
Question: We have a validation for too short password. Should we set max? 
Currently very long password are accepted

robinv85:
@galeza good question
I didn't spend time for more advanced validation, for the password, depending on the configuration on Auth0, we can also add more rules like "min. required special characters" for password strength... but I don't know how much max. character length is.

galeza:
Hi @robinv85 

OWASP recommends max 64 characters (https://cheatsheetseries.owasp.org/cheatsheets/Authentication_Cheat_Sheet.html)

robinv85:
@galeza yes well Auth0 uses some of OWASP's recommendations but they have different settings
And our client-side validation, and API validation should match the one that is used on their side https://auth0.com/docs/connections/database/password-strength so I configured it as "good" which means 

> "at least 8 characters including at least 3 of the following 4 types of characters: a lower-case letter, an upper-case letter, a number, a special character (such as !@#$%^&*)."

but I just didn't want to spend too much time on these details yet, most important here is the general approach for client-side form validations I guess

galeza:
Hi @robinv85 
128 max is ok.
One more question: we accept this email as valid
- "><script>alert(1);</script>"@example.org
 Security -> (XSS attacks)

BMariscal:
> 
> 
> Since there are several error messages would it be cleaner to organize them in one central object and access errors from there?

Like an Enums file? I think that's a great idea @anniegiang

anniegiang:
> > Since there are several error messages would it be cleaner to organize them in one central object and access errors from there?
> 
> Like an Enums file? I think that's a great idea @anniegiang

@BMariscal Yea, something like storing the errors as constants

robinv85:
@BMariscal nope, I need to remove that console.log
@anniegiang I would prefer to keep it like this right now and later update it with translations, I see in other places people have added objects but I don't see an advantage here because it should all contain the name of the field (was discussed on slack in engineering channel) and if we add translations later it can just use function calls with translation strings and some replacement variables similar to `$t("validation.required", "email")`

robinv85:
@galeza for the XSS vulnerability, the form input only sends this to the backend, which should be sanitized (and which is out of scope here) but React should be safe unless explicitly `dangerouslySetInnerHTML` is used https://reactjs.org/docs/introducing-jsx.html#jsx-prevents-injection-attacks

robinv85:
@galeza I've added more detailed password validation for strong passwords, showing a message that it needs 3 of the 4 types of characters...
For the XSS I don't think it's a problem for now, if you think it is we can still open a ticket for it.

PR #587, opened 2020-06-04
Add Papertrail, and allow log level to be configurable

### LINKED ISSUE #571: Make log level configurable/Set up Papertrail
The backend uses the built-in logger from Fastify which is Pino and by default uses "info" level.
It should be possible to change this configuration, so you can change, for example for debugging the application locally. 
It could also be nice to use a logfile locally for better debugging, both are configurable in fastify which is documented [here](https://github.com/fastify/fastify/blob/master/docs/Logging.md).
On deployed environments we also don't want to log everything so it's best to only log WARN and ERROR messages so they can be monitored and if needed handled appropriately.

mannykary:

This PR adds logging to Papertrail, and also allows for the log level to be configurable using the `LOGGER_LEVEL` environment variable. By default, it will be set to `info`.

Resolves #571

### DIFF
--- a/backend/lib/index.js
+++ b/backend/lib/index.js
@@ -3,6 +3,7 @@
 const Ajv = require("ajv");
 const cors = require("cors");
 const fastify = require("fastify");
+const logStream = require("./logger");
 
 const auth = require("./endpoints/auth");
 const feedback = require("./endpoints/feedback");
@@ -14,7 +15,10 @@ const version = require("./endpoints/version");
 
 module.exports = function createApp(config) {
   const app = fastify({
-    logger: true,
+    logger: {
+      level: config.logger.level,
+      stream: logStream(config.logger),
+    },
   });
   const ajv = new Ajv({
     allErrors: true,
--- a/backend/lib/index.js
+++ b/backend/lib/index.js
@@ -14,12 +14,13 @@ const users = require("./endpoints/users");
 const version = require("./endpoints/version");
 
 module.exports = function createApp(config) {
-  const app = fastify({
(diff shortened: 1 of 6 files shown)


backend/lib/models/Comment.js
```diff
@@ -65,7 +65,7 @@ function updateAuthorType(authorID, newAuthorType) {
 
 module.exports = {
   model: Comment,
-  schema: CommentSchema,
+  schema: commentSchema,
```
mannykary, line 68: Note that this commit is actually part of this [PR](https://github.com/FightPandemics/FightPandemics/pull/586). I need to include this commit so that the backend service is able to run.

backend/lib/index.js
```diff
@@ -14,7 +15,10 @@ const version = require("./endpoints/version");
 
 module.exports = function createApp(config) {
   const app = fastify({
-    logger: true,
+    logger: {
+      level: config.logger.level,
+      stream: logStream(config.logger),
```
robinv85, line 20: maybe this should only create a stream if we want a log stream in config?
Because now it seems like if you want to only configure the log level, it will always create a log stream
I already saw a stream write error in Sentry logs, might be related to this?

mannykary: This [error](https://sentry.io/organizations/fightpandemics/issues/1708382915/?project=5251967&query=is%3Aunresolved)? This was me testing locally; I forgot to remove the SENTRY_DSN environment variable from my .env so it sent to Sentry.

I'll change this to conditionally add the stream.

PR #592, opened 2020-06-04
fix: throw & log errors if  mongo id is not found in JWT

### LINKED ISSUE #572: Throw error when Auth0 integration with user mongo_id isn't working
If a user signs up and there is no mongo_id set in the JWT access token coming from Auth0, currently it falls back to the default and generates a unique _id in mongo, but this way the integration with Auth0 isn't working so the backend should return a 500 error instead of creating the user.

robinv85:

fixes #572 
throw & log errors if no mongo id found in JWT, to prevent creating users if the app or Auth0 is not correctly configured

### DIFF
--- a/backend/lib/endpoints/auth.js
+++ b/backend/lib/endpoints/auth.js
@@ -112,6 +112,9 @@ async function routes(app) {
       const { email_verified: emailVerified } = auth0User;
       const { payload } = app.jwt.decode(token);
       const userId = payload[authConfig.jwtMongoIdKey];
+      if (!userId) {
+        throw new Error("no mongo_id found in JWT");
+      }
       const dbUser = await User.findById(userId);
       let user = null;
       if (dbUser) {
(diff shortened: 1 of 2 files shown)


backend/lib/endpoints/auth.js
```diff
@@ -112,6 +112,9 @@ async function routes(app) {
       const { email_verified: emailVerified } = auth0User;
       const { payload } = app.jwt.decode(token);
       const userId = payload[authConfig.jwtMongoIdKey];
+      if (!userId) {
+        throw new Error("no mongo_id found in JWT");
```
anniegiang, line 116: maybe we can access the error for this from ` app.httpErrors`?

robinv85: that's what I firs thad, but if you notice that the whole block is wrapped in `try/catch` it will get caught below and then log the error and return 500...
I also mentioend this in slack backend channel, unlike other loggers the Pino logger doesn't seem to log the errors if they are passed as the second argument so we were doing that wrong

PR #593, opened 2020-06-04
Fix AUTH_APP_URL

mannykary:

Fix AUTH_APP_URL to be the same for all review builds; this is necessary for Auth0 namespacing custom claims

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -71,7 +71,7 @@ module "main" {
     },
     {
       name  = "AUTH_APP_URL"
-      value = local.auth_app_url[var.fp_context]}
+      value = local.auth_app_url[var.fp_context]
     },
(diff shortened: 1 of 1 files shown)


main.tf
```diff
@@ -71,7 +71,7 @@ module "main" {
     },
     {
       name  = "AUTH_APP_URL"
-      value = "https://${var.env_name}.${local.domain[var.fp_context]}"
+      value = local.auth_app_url[var.fp_context]}
```
joshmorel, line 74:
```suggestion
      value = local.auth_app_url[var.fp_context]
```
I think deploy_review failed because of the extract `}`.

Otherwise looks good
Did deploy_build

Issue #565, opened 2020-06-03
[Desktop][Mobile] Create Post | no validation when tag is missing


galeza:
env: staging

1. Go to the Feed screen
2. Add new post
3. Add title and text
4. Do not add tags
5. Post can be added without tags

@TiagoaTavares is this behaviour correct? Should we add validation: add a tag? 
yes -> required

![Screen Shot 2020-06-03 at 19 18 00](https://user-images.githubusercontent.com/21987083/83668042-a7305f80-a5cf-11ea-8808-1a41794c623f.jpg)

galeza:
@manualzuru confirmed that this validation is required

PR #602, opened 2020-06-05
added logic to disable Post button if there are no tags

idjevm:
Doesn't allow the "Post" button when creating a post if there are no tags

#565 

### DIFF
--- a/client/src/components/CreatePost/Form/AskHelp.js
+++ b/client/src/components/CreatePost/Form/AskHelp.js
@@ -32,7 +32,7 @@ const AskHelp = ({
         <Submit
           primary="true"
           onClick={handleSubmit}
-          disabled={!formData.title || !formData.description}
+          disabled={!formData.title || !formData.description || selectedTags < 1}
         >
           Post
         </Submit>
--- a/client/src/components/CreatePost/Form/OfferHelp.js
+++ b/client/src/components/CreatePost/Form/OfferHelp.js
@@ -32,7 +32,7 @@ const OfferHelp = ({
         <Submit
           primary="true"
           onClick={handleSubmit}
-          disabled={!formData.title || !formData.description}
+          disabled={!formData.title || !formData.description || selectedTags < 1}
         >
           Post
         </Submit>


client/src/components/CreatePost/Form/AskHelp.js
```diff
@@ -32,7 +32,7 @@ const AskHelp = ({
         <Submit
           primary="true"
           onClick={handleSubmit}
-          disabled={!formData.title || !formData.description}
+          disabled={!formData.title || !formData.description || selectedTags < 1}
```
robinv85, line 35: @idjevm  just checking, is `selectedTags` a number?  or do you need to use `.length` on an array or something?

idjevm: It is an empty array if there are no tags. There are a few ways I can check for this though. I am not sure which one is better over the other ones...
I could do any of the following: 
```
selectedTags == 0 // true if empty
selectedTags < 1 // true if empty
seledctedTags.length === 0 // true if empty
selectedTags.length < 1 // true if empty

```

Maybe I should change it to `selectedTags.length === 0`?

robinv85: yes it's not going to work if it's an array and you don't use `.lengh` property
`[] < 1` makes no sense

robinv85: you can even use `if ( !selectedTags.length )` which will be evaluated as `false` if 0

idjevm: I mean `[] < 1` worked but weirdly, I guess it is because, behind the scenes, that empty array becomes a string ? or a string 0? But if you evaluate and empty array with < 1 would return true. 

I think you meant `!selectedTags.length` will be evaluated as `true` if 0... 

I updated it to `selectedTags.length === 0` but I could change it to `!selectedTags.length` as well. Both ways should work well though.

robinv85: yes lol i mean `true` :) but i mean `[].length` if 0 is "falsy" in JS terms... but equals 0 is fine too

Issue #581, opened 2020-06-03
fix border on create a post flow


stavares843:
flow
- go to https://staging.fightpandemics.work/
- do all the flow on need help section
- create a post
- appears like that

<img width="673" alt="Screenshot 2020-06-03 at 23 48 07" src="https://user-images.githubusercontent.com/29093946/83696532-980fd880-a5f4-11ea-946e-ec5878eae575.png">

Chrome

idjevm:
This issue was solved with PR #608

PR #608, opened 2020-06-05
Fix for issue 581 - Setting overflow-y to 'auto'

Joeu:

Fix fot the issue [581](https://github.com/FightPandemics/FightPandemics/issues/581) by setting the overflow-y to auto.

### DIFF
--- a/client/src/components/CreatePost/StyledPostAs.js
+++ b/client/src/components/CreatePost/StyledPostAs.js
@@ -36,7 +36,7 @@ const Container = styled(Modal)`
   .ant-modal-body {
     display: flex;
     max-height: 49.6rem;
-    overflow-y: scroll;
+    overflow-y: auto;
     padding: 3.8rem 4rem 4.7rem 4rem;
     min-height: 25.7rem;
     flex-direction: column;


client/src/components/CreatePost/StyledPostAs.js
```diff
@@ -36,7 +36,7 @@ const Container = styled(Modal)`
   .ant-modal-body {
     display: flex;
     max-height: 49.6rem;
-    overflow-y: scroll;
+    overflow-y: auto;
```
robinv85, line 39: I think this was probably added to prevent scrollbars?
Maybe the property can just be removed? 
Did you also try this with different things like if the content grows and things like that?

Joeu: Yes. I tried with smaller screen sizes and works fine for width larger than 120px with **auto** and also **without** the property.
Should I go with removing the overflow-y?

robinv85: yeah just removing is better if it's not needed, less code is usually better :)

PR #610, opened 2020-06-05
fixed email validation on offer help page

omarzoghayyer:

Added email validation to offer help page using { validateEmail } from "utils/validators";

- [x] 1. I have checked that no one else is working on similar changes.
- [x] 2. I have read the latest [**README**](https://github.com/FightPandemics/FightPandemics/blob/master/README.md) and understand the development workflow.
- [x] 3. The name of this branch is: **`feature/<branch_name>`** (need to have cloned the repo not forked).
- [x] 4. This branch is rebased/merged with the **latest master**.
- [x] 5. There are no merge conflicts.
- [x] 6. There are no console warnings and errors.
- [x] 7. On the right hand side of this PR, I have linked the appropriate issue(s) under "Issues" and added a 

project progress under "Projects".

### DIFF
--- a/client/src/pages/OfferHelp.js
+++ b/client/src/pages/OfferHelp.js
@@ -1,8 +1,9 @@
-import React, { useState } from "react";
+import React, { useState, useEffect } from "react";
+import InputError from "components/Input/InputError";
 import { withRouter, Link } from "react-router-dom";
-import styled from "styled-components";
-import { mq } from "constants/theme";
 import { asyncGetGeoLocation } from "utils/geolocation";
+import { validateEmail } from "utils/validators";
+
 import {
   StyledWizard,
   WizardContainer,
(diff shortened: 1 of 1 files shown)


client/src/pages/OfferHelp.js
```diff
@@ -130,19 +131,23 @@ const Step2 = (props) => {
...
+    setValid(validated);
+  }, [email]);
+
+  const onChange = (event) => {
+    setEmail(event.target.value);
   };
+
   const onSubmit = () => {
-    props.update("email", email);
+    console.log("submit");
```
robinv85, line 148: ah yeah, now i see this one is using `props.update()` that one is passed from the page component below... we could still use that too but I'm not sure how this should work when the wizards is completed

omarzoghayyer: We can just have it removed for now and can fix it when the we have the full picture ready.

robinv85: yeah talked on slack about it and it's not clear what has to be done here, good for now

snowiesuet:
Can we add in an error message when user tries to submit without inputting email? Also on desktop, this doesn't look right: 
image

robinv85:
@snowiesuet the components have not been completely finsihed in terms of styling, i also didn't do that because funcitonality was priority
for the error message with a blank input, the problem is, we don't have clear requirements, i already brought this discussion up a few times on slack. We don't have a backend endpoint to save the data and we are not using it yet, so I think this is out of scope here

snowiesuet:
> @snowiesuet the components have not been completely finsihed in terms of styling, i also didn't do that because funcitonality was priority
> for the error message with a blank input, the problem is, we don't have clear requirements, i already brought this discussion up a few times on slack. We don't have a backend endpoint to save the data and we are not using it yet, so I think this is out of scope here
> I also didn't add the "required" property to the email input, that's all there is to it

Got it!

PR #612, opened 2020-06-06
[513] Clean up thrown and logged errors

### LINKED ISSUE #513: Check for null doc when throwing notFound errors
We have code like this in several places https://github.com/FightPandemics/FightPandemics/blob/f48489ac7503b140d0f5de399a0c6e4b95bba989/backend/lib/endpoints/posts.js#L175

```javascript
const [userErr, user] = await app.to(User.findById(userId));
if (userErr) {
    throw app.httpErrors.notFound();
}
```

But if the doc isn't found `userErr = null` and `user = null`. We should check for null user as well in the condition.

@robinv85 I recall you referencing we should log errors in these kind of scenarios. Should that be included in the resolution to this issue?

BMariscal:
### Error string cleanup:
changed `.log.error()` args' order and added conditions to raise a notFound()/404 error when User entity does not exist. 

Closes #513


backend/lib/endpoints/posts.js
```diff
@@ -399,6 +412,8 @@ async function routes(app) {
       );
       if (updateErr) {
         throw app.httpErrors.notFound();
```
joshmorel, line 414: Thanks @BMariscal for the quick turnaround. Just to note a future improvement - the user find stuff after `authenticate` & `optionalAuthenticate`  - it seems were repeating ourselves a lot I think that would be better handled in the middleware. But this is for post MVP refactoring not now.

A few suggestions:

```suggestion
        req.log.error(updateErr, "Failed unliking post");
        throw app.httpErrors.internalServerError()
```
Make this consistent with the others.

backend/lib/endpoints/posts.js
```diff
@@ -226,7 +234,8 @@ async function routes(app) {
       const { postId } = req.params;
       const [postErr, post] = await app.to(Post.findById(postId));
       if (postErr) {
-        throw app.httpErrors.notFound();
+        req.log.error(postErr, "Failed retrieving post");
+        throw app.httpErrors.internalServerError();
       }
```
joshmorel, line 239:
```suggestion
      } else if (post === null) {
        throw app.httpErrors.notFound();
      }
```

backend/lib/endpoints/posts.js
```diff
@@ -368,6 +379,8 @@ async function routes(app) {
         ),
       );
       if (updateErr) {
+        throw app.httpErrors.internalServerError();
```
joshmorel, line 382:
```suggestion
        req.log.error(updateErr, "Failed liking post");
        throw app.httpErrors.internalServerError();
```

backend/lib/endpoints/posts.js
BMariscal: Good catch! Thanks Josh!

backend/lib/endpoints/posts.js
BMariscal: I'll check if there are any more I'm missing. Thanks Josh!

BMariscal: Update: I couldn't find any other instances of missing log.error messages.

backend/lib/endpoints/posts.js
```diff
@@ -180,6 +185,9 @@ async function routes(app) {
       const { userId, body: postProps } = req;
       const [userErr, user] = await app.to(User.findById(userId));
       if (userErr) {
+        req.log.error(userErr, "Failed retrieving user");
+        throw app.httpErrors.internalServerError();
+      } else if (user === null) {
```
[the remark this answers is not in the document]
BMariscal: makes sense. Thanks!

backend/lib/endpoints/posts.js
```diff
@@ -226,6 +234,9 @@ async function routes(app) {
       const { postId } = req.params;
       const [postErr, post] = await app.to(Post.findById(postId));
       if (postErr) {
+        req.log.error(postErr, "Failed retrieving post");
+        throw app.httpErrors.internalServerError();
+      } else if (post === null) {
```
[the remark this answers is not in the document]
BMariscal: makes sense. thanks!

backend/lib/endpoints/posts.js
```diff
@@ -162,8 +165,10 @@ async function routes(app) {
       );
 
       if (postsErr) {
-        req.log.error("Failed requesting posts", { postsErr });
+        req.log.error(postsErr, "Failed requesting posts");
         throw app.httpErrors.internalServerError();
+      } else if (posts === null) {
+        throw app.httpErrors.notFound();
```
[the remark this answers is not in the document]
BMariscal: I do have a question @robinv85  In what scenario is `posts === null`? I've been using postman to trigger this case, and even an empty db returns an empty array, never null

backend/lib/endpoints/posts.js
```diff
@@ -162,8 +166,10 @@ async function routes(app) {
       );
 
       if (postsErr) {
-        req.log.error("Failed requesting posts", { postsErr });
+        req.log.error(postsErr, "Failed requesting posts");
         throw app.httpErrors.internalServerError();
+      } else if (posts === null) {
+        return []
```
robinv85, line 172: small nitpick, I think missing a semicolon `;` 
I think normally when committing this should be fixed with ESLint automatically, not sure if it works for you?

BMariscal: it's definitely not working for me. I'll double check. Thanks!

backend/lib/endpoints/posts.js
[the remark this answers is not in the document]
robinv85: ah yeah well I don't know if this can happen, I thought you considered that
I wasn't sure how the Mongoose API works in all cases, I think there are 2 cases that should be covered
- error occurs
- collection doesn't exist


So if `posts === null` only occurs in case of error, that was already covered
maybe could put them together like `if ( postErr || !posts ) { throw 500 } `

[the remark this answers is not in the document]
joshmorel: I think we only have null for Mongoose `findById`, `findOne` , not `find` similar to most ORM/ODM libraries I think. So 

```
      } else if (posts === null) {
        return [];
      }
```

Is unnecessary but at the same time not harmful (might just cause confusion later) we can revisit with refactoring but I'm fine to merge.

BMariscal:
Oh no, I wasn't done!

backend/lib/endpoints/posts.js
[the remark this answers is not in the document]
BMariscal: I'll create a ticket to follow up on these changes in the future. Thanks all!

Issue #604, opened 2020-06-05
[Desktop][Mobile] Give Help | Question 3 screen link Skip


galeza:
env: https://staging.fightpandemics.work/offer-help#Step3

Steps:
1. Go to Give Help Question 3 screen
2. Click **Skip** link
3. Open empty https://staging.fightpandemics.work/AirTableCOVID screen

**Should open Feed with the address given as a filter.**

![Screen Shot 2020-06-05 at 11 49 25](https://user-images.githubusercontent.com/21987083/83872688-c693cd80-a732-11ea-90e6-d9df68086f5d.png)

idjevm:
So I am not sure about the filters since they don't seem to be working on my end, however, I can change this to go to the feed instead for now.

robinv85:
Eh 2 questions
isn''t this conflicting with the email validation that was added (merged for need help, offer help is open PR right now #610

idjevm:
@galeza So based of @robinv85 comments I think it would be better to wait till #610 has been merged to figure out the next steps. 

I also think that page is deleted. Not sure what would you suggest @robinv85 as I was looking into how the filters work, but they seem to be broken/not working correctly. So as for now, We should have it go to /feed rather than just to /AirTableCOVID since that page seems to no exist anymore.

pthoma14:
@tiagocostasilva @ppagarawal wanted to include you in this discussion about the Home page feature.

indirabhowmick:
@robinv85 @idjevm @galeza @TiagoaTavares @ppagarawal - If the user clicks on the skip button, it should be redirected to the Feed Page.

idjevm:
Yup, that should be already implemented on staging! Was updated on the last PR merged.

idjevm:
We should keep this issue open as we should update it later than we should skip to the Feed Page, and apply the filter of the location once that gets up and running.

indirabhowmick:
@idjevm - I agree, I was about to mention that in the comments. Thanks for pointing that out.

idjevm:
This issue has been resolved.

PR #613, opened 2020-06-06
fix redirect for give help skip email opt

idjevm:
As of now, this just adds a redirect to /feed when a user wants to offer help chooses to skip providing an email
#604 

Note: We can merge this PR for now but leave the issue open so later we can integrate the filters as well.


galeza:
Issue #604 -> based of @robinv85 and @idjevm  comments I think it would be better to wait till #610 has been merged to figure out the next steps.

robinv85:
Ah I see, I actually removed this link from the other Need Help flow... because I wasn't sure what it was supposed to do in fact.

PR #617, opened 2020-06-06
Feature/dummy location

joshmorel:

These changes allow us to work on stuff requiring geolocation until geolocation is done.

1) Commit: "add temp dummy location to CreateUserProfile" - is temporary, when you create profile (via the front-end; in Postman just copy the same object) it adds New York City location regardless of what is in address
2) Commit: "combine fluent locationSchema" is necessary for this but a change I'll be doing anyways with geolocation feature. 

Instead of for feed when authenticated: 
image

You will see:
image

### DIFF
--- a/client/src/pages/CreateUserProfile.js
+++ b/client/src/pages/CreateUserProfile.js
@@ -235,6 +235,14 @@ const CreateProfile = ({ email, history }) => {
     createUserFormDispatch({ type: CREATE_USER });
     try {
       const { email, ...body } = formData;
+      /* TEMP location until geolocation service id done */
+      body.location = {
+        address: 'New York, NY, USA',
+        coordinates: [ -74.0059728, 40.7127753 ],
+        city: "New York",
+        state: "NY",
+        country: "US"
+      }
       const res = await axios.post("/api/users", body);
 
       dispatch({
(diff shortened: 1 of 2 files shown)


client/src/pages/CreateUserProfile.js
```diff
@@ -235,6 +235,14 @@ const CreateProfile = ({ email, history }) => {
     createUserFormDispatch({ type: CREATE_USER });
     try {
       const { email, ...body } = formData;
+      /* TEMP location until geolocation service id done */
+      body.location = {
+        address: 'New York, NY, USA',
```
robinv85, line 240: ah ok cool didn't know it was like this, so does that mean, `Address` field is what can be displayed? Or is that street address and it still has to be mergd together with whatever other info is in here @joshmorel

joshmorel: Based on what some of the designs and a question in Slack my plan was to equate address to Google formatted_address which could be of varying degrees of granularity but is never just street address (e.g. "44 Bank St").

So for example from this: 
https://app.zeplin.io/project/5ecaed1b2522d9280f1a4792/screen/5ed56900260b9c1e9ddf8b2f

image


The Google address is:
"Neukölln, Berlin, Germany"

And the structured location is:
"country": "DE"
"state": "Berlin",
"city": "Berlin",
"neighborhood": "Neukölln"
"coordinates": [13.4445071, 52.4407709]

Based on autocomplete suggestions you could go more granular. This might be confusing so there's way to limit that I think we could resolve post soft launch.

PR #621, opened 2020-06-06
Feature/user profile

### LINKED ISSUE #50: User Profile - Front End
- [x] frontend user profile

robinv85:
fixes #50
update user profile (was not able to render because of the Feed Context used in Posts and sub components, for now just added a few props to make posts reusable for profile).

styling WIP => needs to be double checked

API integration WIP => also needs some more things like location

opening draft PR just to get early feedback if possible


robinv85:
@idjevm which errors do you mean?
there were a few errors here, biggest problem was the FeedContext and the Posts were just imported into the profile without that context, but i fixed that (quick fix for now)
some other small fixes... I'm working on the API integration now before it can be merged.

robinv85:
@idjevm yep this is fixed now, you can preview on  https://featureuser-profile.fightpandemics.xyz
added the API integration for the user, just need to add some more fields
and then load the posts for that user as well and should be good to merge

vikjoshi:
@vikjoshi tagging myself here so I can follow PR and test once merged

PR #626, opened 2020-06-07
Add production deployment; fix staging/review workflows

### LINKED ISSUE #547: Set up production environment
- [x]  Set up MongoDB cluster (after finalizing with contact at Mongo Atlas)
- [x] Run Terraform to set up infrastructure in AWS

mannykary:

Setting up deployment to production through Github Actions.

I also fixed some other issues as well:
* Pass in the git commit properly - before it wasn't actually getting the commit from the environment variable properly, so I had to create a new step to output the commit hash, and then use that output to pass it to Terraform
* Pass in `aws_region` Terraform variable; this is necessary because the default region is `us-east-1`, but for production we are deploying to `eu-west-1`.

Closes #547

### DIFF
--- a/.github/workflows/production-workflow.yml
+++ b/.github/workflows/production-workflow.yml
@@ -47,13 +47,17 @@ jobs:
           terraform {
             backend "s3" {
               bucket = "fp-production-terraform-state"
-              region = "us-east-1"
+              region = "eu-west-1"
               key = "production.tfstate"
             }
           }
           EOF
       - run: cat backend.tf
       - run: cat main.tf
+      - id: get-git-commit
+        name: Get git commit
+        shell: bash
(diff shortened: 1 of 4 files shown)


.github/workflows/production-workflow.yml
```diff
@@ -0,0 +1,107 @@
...
+
+jobs:
+  build_images:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        app:
+          - backend
+          - client
+          - geo-service
```
robinv85, line 16: yeah i guess by the time we will deploy to production we won't need the python service anymore since @joshmorel is working on that and putting it into node

mannykary: Yeah I know, I was going to ask about removing it but didn't want to break things

PR #631, opened 2020-06-07
[606] Password recovery endpoint: Sends user an email with a change password ticket to update password 

### LINKED ISSUE #606: Recover Password Backend
Create route to recover password and functionality to request through Auth0:

Likely endpoint: https://auth0.com/docs/api/management/v2#!/Tickets/post_password_change

Related front-end issue is #63

BMariscal:
This `api/auth/change-password` endpoint makes a POST to Auth0's `/dbconnections/change_password`
https://auth0.com/docs/api/authentication#change-password to send a  change password email and ticket to the email provided. 

Test curl POST request:

```
curl -X POST \
  http://localhost:3000/api/auth/change-password \
  -H 'accept: application/json, text/plain, */*' \
  -H 'cache-control: no-cache' \
  -H 'content-type: application/json;charset=UTF-8' \
  -H 'postman-token: {POSTMAN TOKEN}' \
  -H 'user-agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_12_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/83.0.4103.61 Safari/537.36' \
  -d '{"email":"{YOUR EMAIL}"}'
```

Response:  
```
{
    "email": "bleh@mail.com",
    "responseMessage": "We've just sent you an email to reset your password.",
    "token": ""
}


```

Password reset flow and email template sent:
https://auth0.com/docs/connections/database/password-change

Closes #606

### DIFF
--- a/backend/lib/components/Auth0.js
+++ b/backend/lib/components/Auth0.js
@@ -113,12 +113,6 @@ const changePassword = async (token, email) => {
     );
     return res.data;
   } catch (err) {
-    console.log(
-      "ERR HERE",
-      err,
-      AUTH_DOMAIN,
-      JSON.stringify(config.auth.clientId),
-    );
     return wrapError(err);
   }
 };
--- a/backend/lib/components/Auth0.js
+++ b/backend/lib/components/Auth0.js
@@ -85,9 +85,36 @@ const getUser = async (token) => {
   }
 };
 
+const getUserByEmail = async (token) => {
+  try {
+    const res = await axios.get(
+      `${AUTH_DOMAIN}/api/v2/users-by-email`,
+      getAuthHeaders(token),
+    );
+    return res.data;
+  } catch (err) {
+    return wrapError(err);
+  }
+};
+
+const changePassword = async (token, payload) => {
+  try {
+    const res = await axios.post(
+      `${AUTH_DOMAIN}/api/v2/tickets/password-change`,
+      payload,
+      getAuthHeaders(token),
+    );
+    return res.data;
+  } catch (err) {
+    return wrapError(err);
+  }
+};
+
 module.exports = {
   authenticate,
   buildOauthUrl,
+  changePassword,
   createUser,
   getUser,
+  getUserByEmail,
 };
--- a/backend/lib/components/Auth0.js
+++ b/backend/lib/components/Auth0.js
@@ -85,10 +85,10 @@ const getUser = async (token) => {
   }
 };
 
-const getUserByEmail = async (token) => {
+const getUserByEmail = async (token, email) => {
   try {
     const res = await axios.get(
-      `${AUTH_DOMAIN}/api/v2/users-by-email`,
+      `${AUTH_DOMAIN}/api/v2/users-by-email?email=${email}`,
       getAuthHeaders(token),
     );
     return res.data;
--- a/backend/lib/components/Auth0.js
+++ b/backend/lib/components/Auth0.js
@@ -85,18 +85,6 @@ const getUser = async (token) => {
   }
 };
 
-const getUserByEmail = async (token, email) => {
-  try {
-    const res = await axios.get(
-      `${AUTH_DOMAIN}/api/v2/users-by-email?email=${email}`,
-      getAuthHeaders(token),
-    );
-    return res.data;
-  } catch (err) {
-    return wrapError(err);
-  }
-};
-
 const changePassword = async (token, email) => {
   const client_id = config.auth.clientId;
   const connection = "Username-Password-Authentication";
@@ -123,5 +111,4 @@ module.exports = {
   changePassword,
   createUser,
   getUser,
-  getUserByEmail,
 };
--- a/backend/lib/components/Auth0.js
+++ b/backend/lib/components/Auth0.js
@@ -97,15 +97,28 @@ const getUserByEmail = async (token, email) => {
   }
 };
 
-const changePassword = async (token, payload) => {
+const changePassword = async (token, email) => {
+  const client_id = config.auth.clientId;
+  const connection = "Username-Password-Authentication";
+  const payload = {
+    client_id,
+    email,
+    connection,
+  };
   try {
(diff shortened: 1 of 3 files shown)


backend/lib/components/Auth0.js
```diff
@@ -85,9 +85,49 @@ const getUser = async (token) => {
   }
 };
 
+const getUserByEmail = async (token, email) => {
```
BMariscal, line 88: I'm no longer using this but is it OK to leave it here for future use? It could be useful.

backend/lib/endpoints/auth.js
```diff
@@ -135,6 +136,31 @@ async function routes(app) {
...
+    async (req) => {
+      const { body, token } = req;
+      const { email } = body;
+
+      try {
+        const responseMessage = await Auth0.changePassword(token, email);
+        req.log.info(
+          `Change password email created successfully for email=${email}`,
+        );
+        return { email, responseMessage, token };
```
BMariscal, line 152: Should the token also be returned here?

backend/lib/components/Auth0.js
robinv85: I prefer to delete unused code (YAGNI) and we still have this commit for reference if we need it later
But most of the time I think we will probably use the user id or token
Also, ESLint might complain?
Besides, this `components/` folder originally started from the first version using Express and Passport middleware... and I don't really like the structure (like with the `wrapError` etc. which was a bit of a workaround that I added later to patch it... ) so I think it's better to keep this file minimal and maybe refactor the backend later.

backend/lib/endpoints/auth.js
robinv85: No the server token cannot be returned to the frontend because this can do everything with the Auth0  management API.
Only thing that can be returned is the access token JWT with authorization scopes for the signup & login, 
But I think in fact we should probably change that for Id Token instead eventually, I didn't add this originally but I think the access token also contains scopes for the Auth0 API and not only for use in our API.

backend/lib/endpoints/auth.js
```diff
@@ -135,6 +136,31 @@ async function routes(app) {
...
+      } catch (err) {
+        if (err.statusCode === 400) {
+          throw app.httpErrors.badRequest(err, `User ${email} does not exist`);
```
robinv85, line 155: what's the difference between "not exist" and "not found" for 404 here? I would think 400 is bad request, not sure if I understand

BMariscal: The idea behind this validation came from the response messages from the `POST	/api/v2/tickets/password-change` API endpoint https://auth0.com/docs/api/management/v2#!/Tickets/post_password_change but the `/dbconnections/change_password` endpoint will actually accept any string, even if it's not a valid email address, and send that string an email. 

I tested the  `/dbconnections/change_password` endpoint and even an email string containing just `@` or `@mail.com` is valid and an email is sent. Granted those email addresses don't exist it's funny that the Auth0 endpoint does zero email validation and just returns "We've just sent you an email to reset your password." regardless of what the string input contains. 

Anyway, this error validation should not be here because the `/dbconnections/change_password` does not raise these errors. I will delete this conditional block. Thanks for pointing this out!

backend/lib/endpoints/auth.js
BMariscal: makes perfect sense. Thanks!

backend/lib/components/Auth0.js
BMariscal: Ooh true, ESlint does complaint if declared but unused. There is a ticket to clean up ESLint messages...there's too much noise atm. I'll take care of that.

backend/lib/endpoints/auth.js
joshmorel: Looks like this is resolved and I'm generally good with the PR with the changes you've made. 

The `/dbconnections/change_password` is intended to also be usable directly from a frontend (unlike `password-change` in Management API, which is for backend servers only) so for security reasons they wouldn't want to reveal if the email exists in the db or not.

Not sure for the reason to not validate the format however. That is something I think we should still do ourselves with `email: S.string().format("email"),`

joshmorel: Actually looking at the other auth.js endpoints they aren't using .format("email"). Maybe an oversight or is there a reason @robinv85 ? If an oversight I'd make that as a separate low priority issue.

robinv85: @joshmorel validation of the API through schema's is definitely an important point that needs improvement
I've been working on adding some more validations for the update user endpoint but it definitely takes some effort to get it right, `fluent-schema` should just be syntactic sugar for JSON schema's, but AJV actually includes some custom formats on top of the JSON schema standards. We also have to make sure the backend and frontend have the same validations, I've jsut tested a regular expression directly from the AJV core and it wasn't even accepted as a valid regex, but it works on the frontend.

robinv85: Also, for the email validation on the frontend we have pretty advanced validations, like the domain extension is actually checked, but that might not work for other valid emails like e.g. an IP address which would probably be accepted by the standard email format of JSON schema, and it would be harder to add this domain extension validation on the backend in the JSON schema's as well.

backend/lib/endpoints/schema/auth.js
```diff
@@ -15,6 +15,11 @@ const signupSchema = {
     .prop("confirmPassword", S.string().required()),
 };
 
+const changePasswordSchema = {
+  body: strictSchema()
+    .prop("email", S.string().required())
```
robinv85, line 20: Yeah so adding the `format("email")` would be a pretty good addition
it follows these standards https://json-schema.org/understanding-json-schema/reference/string.html#email-addresses
obviously also depends on the implementation in AJV which is our validation library

joshmorel: I will create another ticket for this & go ahead and merge this PR.

Issue #647, opened 2020-06-08
[Desktop] Profile | doubled add post sign


galeza:
env: https://staging.fightpandemics.work/profile

1. Sign In
2. Go to Profile
3. For desktop application "plus" sign is shown twice (create post and at the bottom of the page):
![Screen Shot 2020-06-08 at 14 06 53](https://user-images.githubusercontent.com/21987083/84028946-e70d9300-a991-11ea-8e43-2396a180eabb.png)

For mobile is ok:
![Screenshot_2020-06-08-14-08-19-862_com android chrome](https://user-images.githubusercontent.com/21987083/84029148-394eb400-a992-11ea-9cad-8ba6dd77d159.jpg)

PR #648, opened 2020-06-08
removing extra icon from the /profile route

idjevm:
This fixex the small bug that create an extra icon on the /profile page
#647


robinv85:
Yeah I knew this but I didn't bother, wasn't sure if it was supposed to be removed
Did you remove the one on the bottom? I can't see here, might be the wrong one, better to create `feature/` branch even for fix so we can preview

galeza:
Hi @idjevm 
Tested also on your https://ec0f9d449a01.ngrok.io
Screen Shot 2020-06-08 at 17 06 40

Issue #564, opened 2020-06-03
[Mobile] Home Page | incorrect Need Help, Give Help | no corners


galeza:
env: staging

Steps:
1. Open app on mobile phone
2. There are no corners for Need Help, Give Help - please see screenshot

![Screenshot_2020-06-03-19-00-31-592_com android chrome](https://user-images.githubusercontent.com/21987083/83666466-4738b980-a5cd-11ea-84f8-33fc22e9c6b0.jpg)

Joeu:
Fixed with this [PR](https://github.com/FightPandemics/FightPandemics/pull/650)

PR #650, opened 2020-06-08
Fix for issue 564 - No borders in mobile homepage

Joeu:

Fix for [issue 564](https://github.com/FightPandemics/FightPandemics/issues/564)
Using !important to override specificity of the third party css border property in ant button. 
 
### DIFF
--- a/client/src/components/Button/FeedbackButton.js
+++ b/client/src/components/Button/FeedbackButton.js
@@ -1,6 +1,6 @@
 import { Button } from "antd-mobile";
 import styled from "styled-components";
-import { theme, mq } from "constants/theme";
+import { theme } from "constants/theme";
 
 const { black, royalBlue, white } = theme.colors;
 
--- a/client/src/components/Button/FeedbackButton.js
+++ b/client/src/components/Button/FeedbackButton.js
@@ -1,6 +1,6 @@
 import { Button } from "antd-mobile";
 import styled from "styled-components";
-import { theme } from "constants/theme";
+import { theme, mq } from "constants/theme";
 
 const { black, royalBlue, white } = theme.colors;
 
@@ -16,8 +16,8 @@ const FeedbackButton = styled(Button)`
     background-color: ${royalBlue};
     color: ${white};
   }
-  &.am-button-ghost:before {
-    border: none;
+  &.am-button {
+    border: solid ${royalBlue} !important;
   }
 `;
 
--- a/client/src/components/Button/FeedbackButton.js
+++ b/client/src/components/Button/FeedbackButton.js
@@ -5,7 +5,6 @@ import { theme } from "constants/theme";
 const { black, royalBlue, white } = theme.colors;
 
 const FeedbackButton = styled(Button)`
-  border: 0.2rem solid ${royalBlue};
   border-radius: 0.8rem;
   cursor: pointer;
   color: ${black};
@@ -16,8 +15,8 @@ const FeedbackButton = styled(Button)`
     background-color: ${royalBlue};
     color: ${white};
   }
-  &.am-button {
-    border: solid ${royalBlue} !important;
+  html body &.am-button {
+    border: 0.2rem solid ${royalBlue};
   }
 `;
 


client/src/components/Button/FeedbackButton.js
```diff
@@ -16,8 +16,8 @@ const FeedbackButton = styled(Button)`
     background-color: ${royalBlue};
     color: ${white};
   }
-  &.am-button-ghost:before {
-    border: none;
+  &.am-button {
+    border: solid ${royalBlue} !important;
```
idjevm, line 20: @robinv85 any thoughts of using `!important`? I have come across this before, and was strongly suggested not to use it and look for alternatives... 

Also @Joeu I think the design doesn't look good. 
So here it is how it is right now but just with the borders being wrong. 
image

Here it is how it looks with your new change -> 
image

@galeza @Yu-Alexa-Zhao I think this should look like it was but with the corners not being cut off.. Not sure though

galeza: Hi @idjevm @Joeu 

This is from Figma design:
Screen Shot 2020-06-08 at 19 02 42

idjevm: Cool! That means that now it looks the right way. 

@Joeu then for now just remove the mq import and wait to see what @robinv85  has to say on `!important` tag...

Joeu: Thanks, @idjevm . I was testing over the media queries approach and forgot to remove it. Just pushed the change.
About the **!important**, the thing is that without it, the property will be set by antd lib.

client/src/components/Button/FeedbackButton.js
```diff
@@ -1,6 +1,6 @@
 import { Button } from "antd-mobile";
 import styled from "styled-components";
-import { theme } from "constants/theme";
+import { theme, mq } from "constants/theme";
```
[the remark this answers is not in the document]
Joeu: Thanks for this catch, bro!

client/src/components/Button/FeedbackButton.js
Yu-Zhao-UX: Yes I believe the Figma & Zeplin screenshot you have is the correct design

Yu-Alexa-Zhao: Yes I believe the figma/zeplin screenshot is the correct one

silenceway: This rule is only overwriting the rule from line 8. the important can be deleted. And it's only overwriting the border width that can be changed in the same line 8. All this line can be deleted.

Joeu: @silenceway , applying the changes you suggested and using the _Toggle device toolbar_ on Chrome, the issue still remains. I also tested with my mobile.

That's why I needed to use the !important, even knowing that normally it should be avoided.

Screenshot from 2020-06-09 00-46-38

silenceway: If it's happening only on mobile, you can check the rule and try to override without an important. This works:
`  html body &.am-button {
    border: 0.3rem solid ${royalBlue};
  }`

robinv85: Yeah using `!important` is generally not recommended because it can sometimes cause more problems than solve them.
For this particular example it's not really a problem because it's overriding styles from Ant Design.
It's better to learn about the[ CSS specificity ](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity)to understand why using `!important` is not recommended, but that's especially for reusable components, here it's only for the end result on the home page so probably nothing else will extend those styles.

Joeu: > If it's happening only on mobile, you can check the rule and try to override without an important. This works:
> ` html body &.am-button { border: 0.3rem solid ${royalBlue}; }`
> And I think the other rule can be removed. (Rule in line 8).

Thanks for this useful tip, @silenceway . Works like a charm and avoided overriding the specificity.

Joeu:
> Please review.

Can you check it now, @silenceway ? Thanks.

silenceway:
I will merge but please review @idjevm's comment.

galeza:
Hi,

@idjevm I'll ask designers about these borders

manualzuru:
That blue border shouldn't be there. @galeza @idjevm this is low priority now!

Issue #490, opened 2020-05-26
Remove redux and replace with Context API


robinv85:
Currently redux is only used for the user authentication: to store the token and initial response from the server (that can contain some user info like name, email, emailVerified, avatar etc.)

For other places we are using mostly local state with useReducer hooks, which is similar to redux but not in global state and requires less boilerplate code.

To avoid confusion with the actions & reducers from redux, and because it seems overkill for our use case, we could probably remove redux and replace it by the React Context API ... alternatively we could add another state management library if someone has suggestions and thinks this is useful for this use case (e.g. MobX, Recoil...).

But I think to improve performance and to clean up the code, Context API is probably enough. 
Also this ticket would be nice to have for MVP but it not essential, and it should probably wait until the tickets for user auth MVP and forms cleanup are finished.

Issue #462, opened 2020-05-24
Clean up forms & layouts


robinv85:
There are multiple pages with forms and layouts that are very similar
It's best to check the form components and styles so they are optimized for reusability, instead of reinventing the wheel, leveraging libraries such as `react-hook-form` for some consistency.

The layout templates can be set on the route configuration, while some pages seem to have a custom layout without a template, this can be checked to see how it can be improved, maybe it's possible to just use a reusable page component e.g. left/right column that just has an image prop. and can wrap the content on the right panel (for desktop view). Maybe it's better not to use these custom layouts since it makes the RouteWithSubRoutes component currently quite complex.

PR #654, opened 2020-06-09
feat: edit user profile / account

robinv85:

Add edit user profile / edit user account
API endpoint added (first draft version basic implementation)
todo: check URL's input format & sanitization

Will add frontend too before merging


robinv85:
Some work I've been doing also relates to these 2 big issues 

#462 to clean up forms:
I've done some work on these form components, and currently it seems that the components like  `BaseInput` and `InputWrapper` that are used on Login / Signup and Create Profile can probably be merged and replaced by the `FormInput` in which I've added some more functionality here, but needs to be checked for the whole application 

#490 to improve the state management and look at useage of Redux, useReducer hooks and Context API:
Here I've been able to reuse the `userReducer` logic for user profile and use the Context API with a HOC wrap a provider around these pages, but this doesn;t seem sufficient yet, it would be better to integrate this at a global level after login with the user, and it seems like we might be trying to reinvent redux here which makes me think maybe redux isn't so bad after all (for example: it helps to prevent unneeded renders for state updates and also helps to combine mutliple reducers like sstate slices).

galeza:
Hi @robinv85 

I have questions about frontend (let me know, if it's not in the scope of this PR)

1. When I navigate between Profile Information/ Account Information - On Account Information Avatar is not shown, only bottom of the icon is shown:
Screen Shot 2020-06-11 at 18 38 36

2. In zeplin: there is email field (disabled) on the Account Information screen. 

3. When I select edit icon from Profile page:
Edit Account Information has a href="/edit-profile". Shouldn't it have href="/edit-account"?:
Screen Shot 2020-06-11 at 18 46 12

4. Could we move Don't show my address checkbox a little on the left like in the design:
https://app.zeplin.io/project/5ecaed1b2522d9280f1a4792/screen/5ed0333ed0ac739bb73c36df

robinv85:
Hi @galeza 
thanks for already testing, it's not completely ready yet, still draft PR

1. about point 1 

The problem is because if you refresh the page, currently the "session" data is lost, so the avatar with the initials is not properly rendered. How I want to fix this in the first place:
- make sure this page is not accessible when not logged in (will redirect back to login page)
- fix the session storage so you can refresh and keep the data

And then next steps probably we can also fix that Avatar component to make it more robust if no data is passed, and we also have to fix that for image uploads here but i think it's a separate github issue that also needs backend work.

2. I asked about this on slack and seems like the design is not accurate, same for location fields, so i don't think it has to completely match the current design on zeplin, we can add email change later after MVP

3.  didn't notice that, I will check :)

joshmorel:
The idea of adding the url base for social accounts is good but we need to either include the base url anywhere the link is used OR pass the base with the username to the backend (so `https://www.linkedin.com/in/joshmorel/` instead of `joshmorel`) 

image

Right now href="joshmorel" which is no good.

I'd prefer sending the passing the base url when updating as we there there are fewer places we need to worry about it.

robinv85:
@joshmorel yeah I'm working on updating the backend too so it uses the same regex validations, and only stores the username part, it's pretty straightforward to add the URL part of the social networks... I didn't really consider this in scope here but I guess it's not much work to fix it on view profile too

robinv85:
I think this should be ready for review now but there are 2 related problems: 
- location component doesn't always work, but it's quite complex and also used on the create-profile without initial value so maybe it needs to be fixed in a separate pull request
- view profile also might not be working with these social URL's and we also need to add icons and maybe some more finishing touches but naybe better also in separate PR

client/src/pages/EditProfile.js
```diff
@@ -17,59 +18,123 @@ import {
...
-    "Github URL": ["githubURL", githubURL],
-    "Personal Website": ["personalURL", personalURL],
-  };
+function EditProfile() {
+  const { userProfileState, userProfileDispatch } = useContext(UserContext);
+  const { errors, formState, register, handleSubmit } = useForm({
+    mode: "change",
+  });
+  const { error, loading, user } = userProfileState;
+  const { firstName = "", lastName = "", urls = {}, about } = user || {};
```
joshmorel, line 101: The URL object keys are sent with empty strings if not entered in the form:

image

I think before sending we should delete any keys in the `url` object sent to the backend which are like `"facebook": ""`

joshmorel: Sorry I didn't have recent version. it seems like the bakcend is handling this case now.

robinv85:
ok yeah just doing some finishing touches before I merge and we can fix location input on a separate issue

client/src/pages/EditProfile.js
robinv85: yeah i think it should be fixed now, this terrible error message could possibly still return from the backend and use these JSON schema and regex rules but if we see those we can hopefully fix it

PR #659, opened 2020-06-09
fix post patch undefined postProps

### LINKED ISSUE #653: `postProps` undefined in `PATCH /api/posts/:postId`
Seems like I've spotted a bug here

https://github.com/FightPandemics/FightPandemics/blob/master/backend/lib/endpoints/posts.js#L348

```
      if (EXPIRATION_OPTIONS.includes(postProps.expireAt)) {
        body.expireAt = moment().add(1, `${body.expireAt}s`);
```

And makes me wonder why this got merged, should probably be caught by ESLint and preventing to commit...

joshmorel:

Resolves #653 

@robinv85  on your comment

> And makes me wonder why this got merged, should probably be caught by ESLint and preventing to commit...

husky will prevent this but **only** if installed and I think it can be skipped to. We could add to github actions workflows but right now that will create a lot more headaches than it is worth - we can look into this again post MVP. (Create a ticket?)

### DIFF
--- a/backend/lib/endpoints/posts.js
+++ b/backend/lib/endpoints/posts.js
@@ -332,7 +332,7 @@ async function routes(app) {
       schema: updatePostSchema,
     },
     async (req) => {
-      const { userId } = req;
+      const { userId, body: postProps } = req;
       const [err, post] = await app.to(Post.findById(req.params.postId));
       if (err) {
         req.log.error(err, "Failed retrieving post");
--- a/backend/lib/endpoints/posts.js
+++ b/backend/lib/endpoints/posts.js
@@ -332,7 +332,7 @@ async function routes(app) {
       schema: updatePostSchema,
     },
     async (req) => {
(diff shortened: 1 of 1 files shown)


backend/lib/endpoints/posts.js
```diff
@@ -332,7 +332,7 @@ async function routes(app) {
       schema: updatePostSchema,
     },
     async (req) => {
-      const { userId } = req;
+      const { userId, body: postProps } = req;
```
robinv85, line 335: one line is probably not sufficient
in the code below there's also 
`const { body } = req;`
and then both `postProps` and `body` are used so should be one const

joshmorel: Silly me just fixing error without looking at rest of code. :sweat_smile: 

The `{ body: postProps }` thing is only used in POST post. Other places use `{ body } = req` or just `req.body` so we're not exactly consistent. I think `{ body } = req` is good, you?

Issue #551, opened 2020-06-02
Update comments API to allow replies.


silenceway:
It depends on https://github.com/FightPandemics/FightPandemics/pull/378

vikjoshi:
@silenceway , #378 has been merged. Are there any other blockers to being working on this?

adisen:
This is like replying to comments right?

vikjoshi:
@adisen , yes that's right

adisen:
Okay.

Naraujo13:
This was already implement in # as discussed in #672.

robinv85:
Ok I think @silenceway wass working on this in the frontend, might have been a misunderstanding that this wasn't implemented yet...

PR #672, opened 2020-06-10
Feature reply to comment

adisen:
A user can reply to a comment


backend/lib/endpoints/posts.js
```diff
@@ -29,6 +29,9 @@ async function routes(app) {
   const UNLOGGED_POST_SIZE = 120;
   const EXPIRATION_OPTIONS = ["day", "week", "month"];
 
+  // @desc    Get all posts
```
robinv85, line 32: not sure why you added this now, seems out of scope of the pull request and it might cause more merge conflicts with other PR's 
not saying this is bad but maybe it's better to open a separate issue for this and look at the possibilities to improve code documentation instead of just adding it

robinv85: and if we decide to use this we could add everywhere in the API
also maybe we should chedk the integration with the `fastify-oas` plugin because this can help improve maybe also the API documentation that is automatically generated and can be seen with the Swagger UI?

adisen: Sorry. I already created a branch for the documentation. Didn't know it'll cause issues here.

adisen:
> would like to get feedback about the code documentation, what do other engineers think @joshmorel @BMariscal @Naraujo13 @joaofnds
> 
> Also, not sure if this is the best way to use `POST /api/posts/:postId/comments/:commentId/replies` endpoint since this was implemented already before with `POST /api/posts/:postId/comments/` and using `parentId` in the request body, this seems better to have not too complex API routes and offers more flexibility like possible to add more nested levels in the future without adding another level in the route...?

Don't get what you mean here @robinv85

adisen:
I sort of just picked up an issue. That means the issue was not necessary initially then.

robinv85:
@adisen I will try to explain again so it's more clear

so you added this endpoint
`POST /api/posts/:postId/comments/:commentId/replies`

and I thought this functionality already existed before, i also worked on it and I know that @Naraujo13 and @joaofnds have worked on the updated data model and the API but i'm not sure about current atatus, what is working and what still needs to be done

so what I think we had before was more simple, with the same result

`POST /api/posts/:postId/comments/`

- create a comment on a post
- create a reply to a comment, which is also a comment, but includes a parent id in the request body

this way, if you want more than 1 level of nesting you can also use this, but you don't need to have the long URL like `comment/id/replies/id/replies/id...` etc.

adisen:
> @adisen I will try to explain again so it's more clear
> 
> so you added this endpoint
> `POST /api/posts/:postId/comments/:commentId/replies`
> 
> and I thought this functionality already existed before, i also worked on it and I know that @Naraujo13 and @joaofnds have worked on the updated data model and the API but i'm not sure about current atatus, what is working and what still needs to be done
> 
> so what I think we had before was more simple, with the same result
> 
> `POST /api/posts/:postId/comments/`
> 
> * create a comment on a post
> * create a reply to a comment, which is also a comment, but includes a parent id in the request body
> 
> this way, if you want more than 1 level of nesting you can also use this, but you don't need to have the long URL like `comment/id/replies/id/replies/id...` etc.
> 
> @adisen is there a github issue for this? we should make sure that we avoid confusion about what still needs to be done and what is already implemented

Yea there was, I guess we can close this PR?

Naraujo13:
> @adisen I will try to explain again so it's more clear
> 
> so you added this endpoint
> `POST /api/posts/:postId/comments/:commentId/replies`
> 
> and I thought this functionality already existed before, i also worked on it and I know that @Naraujo13 and @joaofnds have worked on the updated data model and the API but i'm not sure about current atatus, what is working and what still needs to be done
> 
> so what I think we had before was more simple, with the same result
> 
> `POST /api/posts/:postId/comments/`
> 
> - create a comment on a post
> - create a reply to a comment, which is also a comment, but includes a parent id in the request body
> 
> this way, if you want more than 1 level of nesting you can also use this, but you don't need to have the long URL like `comment/id/replies/id/replies/id...` etc.
> 
> @adisen is there a github issue for this? we should make sure that we avoid confusion about what still needs to be done and what is already implemented
> 

You're right @robinv85 , the endpoint `POST /posts/:postId/comments/` is used to create comments, nested or not. The nesting is done through the attribute `parentId` and though we currently use only 1 level, we could add more in the future with this structure.

This endpoint is already working since PR #378

robinv85:
@adisen @Naraujo13 is #551 the ticket that is already implemented or not? https://github.com/FightPandemics/FightPandemics/issues/551

Naraujo13:
> @adisen @Naraujo13 is #551 the ticket that is already implemented or not? 

Well that ticket didn't exist when I implemented this feature on #378, but I think it is the same feature.

Issue #561, opened 2020-06-03
Fix home navbar signed in view 


joshmorel:
@robinv85 mentioned this was working awhile ago but not anymore. Can someone get this fixed for both mobile and desktop?

**Mobile**
https://app.zeplin.io/project/5ecaed1b2522d9280f1a4792/screen/5ed56445c6ac97228f5752dd
![image](https://user-images.githubusercontent.com/10546720/83635882-f98f6300-a572-11ea-81c4-1c3c9e7778e3.png)

**Important:** Exclude "Health Facilities" and "Symptom Check" links for now MVP (will add later post-MVP)

**Desktop**:
https://app.zeplin.io/project/5ecaed1b2522d9280f1a4792/screen/5ed034aa6f722bba0e0e6db2
![image](https://user-images.githubusercontent.com/10546720/83636579-3019ad80-a574-11ea-9d3d-2585952b25de.png)

~~Note: There is some inconsistencies with design for Desktop I will follow up with designers in Slack.~~ (addressed use one above)

agmaal1:
Some doubts:
What is the "+add new one button" expected to do? 
What route "feedback" points to?

Issue #656, opened 2020-06-09
[Desktop][Mobile] persist login session after refresh


galeza:
env: https://staging.fightpandemics.work/

Steps:
1. Log in to app
2. Go to Profile page (See you email, account data)
3. Refresh page
4. Login screen is shown. User is logged off. 

Should add "keep me logged in" feature.

robinv85:
@galeza following your description of the issue
is it ok if I rename the title to "persist login session after refresh" or something like that?
then we can look at a more advanced "keep me logged in" feature  later which is not only session after refresh, but can keep a logged in user later, also depending on token expiration time etc.

galeza:
Hi @robinv85 

I'm changing the title of this issue to "[Desktop][Mobile] persist login session after refresh"
Should I add a new issue "[Desktop][Mobile] Add advanced "keep me logged in" feature" or sth like that? Just I want to keep track of this problem?

robinv85:
@galeza maybe we should ask to product / designers if it should be added on the login page

galeza:
@robinv85 
I'll ask product/designers.

galeza:
@indirabhowmick @ppagarawal 

Are we going to implement an advanced "keep me logged in" feature. Some application have checkbox:  keep me logged in?
If yes, I'll create a separate issue for that.

pthoma14:
Hi @galeza I think an advanced "keep me logged in" feature is a great idea!  I personally think for now though we should add it to our Product backlog on Notion to be discussed further. I'm not sure adding it as an issue right now is the best strategy as we're all laser-focused on launching our MVP-Critical features.

indirabhowmick:
@galeza @robinv85 @ppagarawal - I like the idea of having a persist login session after a refresh. Like @ppagarawal mentioned - we can add to the product backlog and prioritize for later release. Thanks so much!

manualzuru:
But let's eitherway keep them logged for softlaunch. For a later release we give them the option.

joshmorel:
@galeza @manualzuru @indirabhowmick @pthoma14 This will be resolved with #732.

With regards to keeping them logged in - I set it for a week expiry for now. We can either ask them to keep them logged in (after browser close) or just do it for them. Because this isn't an app where, for example, you aren't moving funds around like a banking app, the later may be better. GitHub, as one example, does this but but if you want to do something destructive (like deleting a repo) requests your password.

robinv85:
[...]

PR #680, opened 2020-06-10
Feature/fixhomenavbarsignedinview#561

agmaal1:

Fix home navbar signed in view #561 

- added links pointing to the corresponding routes. 
- added logout link and funcitonality both for mobile and desktop.
- display avatar with user initials.
- display username.

I did not added "Health Facilities" and "Symptom Check" as indicated in the issue. I also did not added "+add new one button" and "feedback" since I'm not sure what they are meant to do. 

Screenshot 2020-06-10 at 21 18 09

Screenshot 2020-06-10 at 21 18 31


client/src/utils/auth-token.js
```diff
@@ -21,3 +21,7 @@ export const getAuthToken = () => {
 
   return token;
 };
+
+export const removeAuthToken = () => {
+  localStorage.removeItem(KEY_TOKEN);
```
robinv85, line 26: @agmaal1 looks good except that we shouldn't use local storage for the JWT token by default
this was done in first version, but now there is a github issue about that
might be out of scope here but if we use a storage for the token it should probably be session storage if the user doesn't choose to stay logged in

also see #656 and my comment

Issue #563, opened 2020-06-03
Onboarding flow small UI bugs, and Submit button size issue 


ppagarawal:
Please help fix the "Submit" button size to equalize to the email address field

![image](https://user-images.githubusercontent.com/6276256/83656487-3bed8a00-a5dd-11ea-878e-e74313a97ed3.png)

Comment/ping me for questions.

ppagarawal:
This issue seems to have come up again but with respect to size and alignment. 

![image](https://user-images.githubusercontent.com/6276256/83971139-44a0d180-a8f7-11ea-926c-979135d4e1f6.png)

ppagarawal:
Another image:

![image](https://user-images.githubusercontent.com/6276256/83978383-bcd1bc00-a924-11ea-8158-2839b39628c4.png)

omarzoghayyer:
@BMariscal Im not sure why I was unassigned on this issue. I was working on it since @ppagarawal reopened it, and I already have a PR open for it.

Joeu:
I'm assigning to you then, @omarzoghayyer , since you already have a PR fixing this.

idjevm:
Reopening this issue instead of creating a new one, the problem is still happening, so let's try to get it as it is on the design. 

Need help step 3 design:
https://app.zeplin.io/project/5ecaed1b2522d9280f1a4792/screen/5ed0378f0c4a0f97f183cf26

Offer help step 3 design: 
https://app.zeplin.io/project/5ecaed1b2522d9280f1a4792/screen/5ed03866f134a6b47f0746c0

Right now we match neither of these designs, so we should iterate and try to get it there.

@Joeu  I know you were gonna work on this issue but Omar had started. But since this is a new iteration, can you take a look and see if the design can be matched? I know Omar has already started looking into other issues so it would be good to get another set of eyes on this one.

idjevm:
Also, I would want to know if in the need help we are trying to force the user by not giving it the chance to skip on step 3? or should both the `need help` and `offer help` be equal and have the option to skip (step 3 (email input) )? @manualzuru @galeza @ppagarawal ?

idjevm:
Some more context here: 
- [x] Create a skip button for the need-help page. It should look like the one similar to the one on the link https://staging.fightpandemics.work/offer-help#Step3
- [x] Skip must redirect the user to the feed.

Joeu:
[...]

PR #702, opened 2020-06-12
Fixing issue #563

Joeu:

Fix for issue #563 

### This PR should:
- [x] Add skip button to both Offer Help and Need Help flows, and the skip button should redirect to /feed.
- [x] The size of the submit button should be the same as the container of the enter your email address.
- [x] Submit button should initially be disabled and only become available once a correct email has been put in the field
- [x] All the NEXT buttons should become white when hovered over the same as the Submit button for both Need Help and Offer Help


robinv85:
yeah i dunno code looks ok what's the problem @idjevm ?

client/src/components/StepWizard/WizardSubmit.js
```diff
@@ -1,24 +1,29 @@
...
-border-radius: 4rem;
-margin-right: 7rem;
-margin-left: 7rem;
+  border-radius: 4rem;
+  align-self: center;
   cursor: pointer;
   color: ${white};
   padding: 0 0.1rem;
   background-color: ${royalBlue};
+  width: inherit;
```
robinv85, line 13: is this the same as 100%? or maybe can just remove, not sure, i never use this

robinv85:
@idjevm the last isn't needed I think

>  All the NEXT buttons should become white when hovered over the same as the Submit button for both Need Help and Offer Help

because we discussed this before on slack with designers, and this was not really part of the design
I think we want to change it to a lighter color on hover / click

idjevm:
> @idjevm the last isn't needed I think
> 
> > All the NEXT buttons should become white when hovered over the same as the Submit button for both Need Help and Offer Help
> 
> because we discussed this before on slack with designers, and this was not really part of the design
> I think we want to change it to a lighter color on hover / click

So @ppagarawal perhaps I can get some input here?

ppagarawal:
> > @idjevm the last isn't needed I think
> > > All the NEXT buttons should become white when hovered over the same as the Submit button for both Need Help and Offer Help
> > 
> > 
> > because we discussed this before on slack with designers, and this was not really part of the design
> > I think we want to change it to a lighter color on hover / click
> 
> So @ppagarawal perhaps I can get some input here?

I think that to maintain that consistency we can have the next button change it to a lighter color on hover / click interaction as @robinv85 mentioned. Currently, mouse over hides the text on the button.

image

PR #707, opened 2020-06-12
refactor address input component and integration in form

robinv85:

Just a draft PR, trying to fix the location input, not completely sure about the required functionality (e.g. when typing backspace after value is already set) :

better integration with react-hook-form by passing `validate` function to a custom `register()` and use the `setValue` with the `onLocationChange` callback.

changed internal state in the component because it seems that `inputAddress` was derived from the location, and changes inside the input value, while the `locaiton` object now also has internal state while also triggering the `onLocationChange` callback to update the parent

So instead of `setInputAddress` now the `setLocation` can be used... 

Not entirly sure about this , probably even a better way to make this a custom form component that integrates nicely with the form library

### DIFF
--- a/client/src/components/Input/AddressInput.js
+++ b/client/src/components/Input/AddressInput.js
@@ -39,14 +39,17 @@ const SubLabel = styled.small`
   display: block;
 `;
 
-const AddressInput = ({ location, error, onLocationChange }) => {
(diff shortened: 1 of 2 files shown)


robinv85:
fixed in #721

PR #721, opened 2020-06-15
Feature/improved location input

### LINKED ISSUE #705: Improved location input
Make it work with create & edit individual profile, clean up the react code to better utilize `react-form-hooks` and also on backend schema make it required for create user (optional for update user). On the location object, the coordinates are required but nothing else.

joshmorel:

Resolves #705 

I updated the `LocationInput` (renamed from `AddressInput`) to work in Feed filter as well as create/edit profile. 

I also switched to using antd `Select` instead of input with dropdown menu because a) it's more appropriate and b) I think will be easier to modified work with later. I did need to copy some `BaseInput` styles so there is some duplication now but we can refactor it now.

I did add it in the give help onboarding flow but I believe that whole thing isn't working yet. I can also work on that if no one else is.

A few key considerations:

1. The style will match the profile form design, **NOT the location search bar design** for onboarding & feed filters. We can either address this later, or ideally come up with a more common design post launch.
2. I thought the `react-hook-form` `Controller` would work as a wrapper to properly integrate this input with those forms using the library. I still think it will, but the whole thing was still too confusing and I bailed before going down the rabbit hole. I kind of hate that library lol but I figure after reading through more examples I'll figure it out later. IMHO the workaround is fine until much post MVP refactoring. When edit organization profile is included it will just need to follow usage in `EditAccount.js`
3. A weird thing, when you tab into the `Select` the cursor doesn't show up til you start typing. Not sure the cause but can look into this later.

feed-location-filter


joshmorel:
Thanks. I will merge so we can continue integrating with onboarding flow.

PR #725, opened 2020-06-15
Fix header nav to top of page

### LINKED ISSUE #710: Fix the header navigation bar
The header bar currently is fixed to the top location and is not fixed when the user scrolls down.

As a user I want to have access to the top nav bar which includes FP logo, Feed button, Profile, etc. on the header so that I can access them even when I am browsing the feed by scrolling down.

Current behaviour:
header

Acceptance criteria:
- [ ] The header remains fixed even when the user scrolls down.
- [ ] The side menu is not impacted.
- [ ] No other functionality is impacted

jyclam:
closes #710

Fix header nav to top 

gif

### DIFF
--- a/client/src/templates/layouts/Main.js
+++ b/client/src/templates/layouts/Main.js
@@ -3,6 +3,8 @@ import { theme, mq } from "constants/theme";
 const { offWhite } = theme.colors;
 
 const Main = styled.main`
+  padding: 7rem 0rem 0rem 0rem;
+
   display: flex;
   align-items: center;
   justify-content: center;
--- a/client/src/components/Header.js
+++ b/client/src/components/Header.js
@@ -91,9 +91,16 @@ const activeStyles = {
   color: `${colors.royalBlue}`,
 };
 
+const HeaderWrapper = styled.div`
+  position: fixed;
+  top: 0;
+  z-index: 1;
+  width: 100vw;
+`;
+
 export default ({ onMenuClick, isAuthenticated }) => {
   return (
-    <div className="header">
+    <HeaderWrapper className="header">
       <StyledNavBar
         mode="light"
         leftContent={
@@ -160,6 +167,6 @@ export default ({ onMenuClick, isAuthenticated }) => {
           </div>
         }
       />
-    </div>
+    </HeaderWrapper>
   );
 };


client/src/templates/layouts/Main.js
```diff
@@ -3,7 +3,7 @@ import { theme, mq } from "constants/theme";
 const { offWhite } = theme.colors;
 
 const Main = styled.main`
-  padding: 0 2rem;
+  padding: 7rem 2rem 0rem 2rem;
```
idjevm, line 6: @robinv85 do we need this much padding though? I am actually deleting this padding in another PR. I had a picture that I want to display from left to right completely and this causes to padding always be 20px on the left and right... Can we remove this @robinv85 ?

robinv85: @djevm I don't know, is that to provide padding where the nav manu will come? have no idea, would have to check the code and see the elements in the browser

idjevm: So what happens here is that this padding it is applied to all the main part of the page, except the footer and the navbar. I had some trouble actually displaying an image on width: 100% when I was working on the About Us page cuz this damn 2rem padding on left and right. I believe doing this will provide padding from the navbar with the main page content, and again 2 rem left and 2 rem right. I would much rather not have this since as my example of wanting to display an image full width, there was not an easy way to actually override this since everything would be a child of main. I think if one's wants padding on a specific page, they should add it on their container rather than applying it across the platform.

Joeu: Since we have sections that must be full width, I think we should avoid using this padding in a container that holds all the other components.

idjevm: That's correct. I agree with you. I have already removed this property in another PR, so perhaps we should keep it removed. If padding is needed, then add padding to the header or the component that needs padding?

idjevm: But yeah @robinv85, I guess to better explain it as I didn't do really well on the last one. This setting applies to the main component that holds all the components that are on the page, so as Joeu said, if you had a component that you want to display in full width, then this padding on the main will give you a hard time since it is the parent of that component.

jyclam: Are we okay with keeping just the top and bottom paddings on main? @Joeu @idjevm

Joeu: @jyclam , so far I've just hit the need of change for left and right padding when @idjevm and I discussed about the AboutUs page. So I think it is ok to keep top and bottom paddings, after all we have header and footer across pages.

robinv85: yeah now I understand, seems better not to add this here indeed
so the padding at the top can stay? 
But maybe also only add that inside the layout, not on the full Main component here
in fact the layout templates structure also needs some work i think

jyclam: @robinv85 we should probably do the layout templates rework in another PR. Is there a better place to put this top padding with the current layouts?

idjevm: Let's just remove that padding completely. If there are any issues, let's add padding to the needed container.

client/src/templates/layouts/Main.js
```diff
@@ -3,6 +3,8 @@ import { theme, mq } from "constants/theme";
 const { offWhite } = theme.colors;
 
 const Main = styled.main`
+  padding: 7rem 0rem 0rem 0rem;
```
idjevm, line 6: Please remove this completely.

idjevm: 70 px is way too much padding...

jyclam: That's the current header's height. If I remove it, the content (feed) will be underneath the header

client/src/templates/layouts/Main.js
jyclam: Ah, okay. Just saw this comment. I can go through and see which pages will need to be padded

client/src/templates/layouts/Main.js
idjevm: No it won't happen. I removed it already on a previous PR. So let's just remove that padding property completely and let's merge this. Then we can double check,

jyclam: I've just double checked, with the css changes that added to make the header fixed. The main flex container slides underneath. Should I go and edit the padding on each page that has the header or should we just keep it with this main layout?

screencapture-localhost-3000-feed-2020-06-18-16_00_04

idjevm: Yeah I see what you mean, then let's add some padding to that component specifically.

jyclam: I've just checked, it's probably not a good idea to do it specifically for each page affected as every page with the header is affected. eg.:

- /about-us
- /need-help
- /feed
- /nearest-hospital
- /offer-help
- /symptoms-check
...

Albeit for some of these pages the effect isn't that noticeable, but it's definitely shifted under. I think it makes sense to keep the padding with main until we decide to undertake reworking the layout templates.

client/src/templates/layouts/Main.js
robinv85: I mean, can't we add it inside the `NavigationLayout` component only? 
Just because we could have a "main" element also in another layout without the main navigation so it might have different needs, but I have opened an issue to rework the layout templates  as well

client/src/templates/layouts/Main.js
robinv85: No I think you're right @jyclam 
I didn't want to suggest that, but seems a bit weird that we have a separate component `Main` and that it always gets the padding too, seems to me the padding should be set on the layout
maybe it doesn't even make sense to have the Main as a separate component, I think this was done a few weeks ago but some people have done weird things with the code that just got merged without proper reviews and we still have some technical debt there

idjevm: The strangest thing is that top padding on the main component was 0, so I am confused how now it is required?

jyclam: @idjevm I believe it's because `position:fixed` makes the header into a `positioned` element and a stacking context is also created. `Positioned` elements are displayed in front of `non-positioned` elements.

https://philipwalton.com/articles/what-no-one-told-you-about-z-index/

idjevm: Ahh that makes total sense, lol, let's get this merged then!

PR #727, opened 2020-06-16
Add unique constraint to user's email at mongoose model

### LINKED ISSUE #590: Make Users.email unique
Email would need to be unique across IndividualUsers & OrganizationalUsers.

We need to decide if we let Mongo handle the error OR use this library to handle with a library: https://www.npmjs.com/package/mongoose-unique-validator

Naraujo13:

Add a constraint to guarantee that the user's email is unique.

Closes #590

### DIFF
--- a/backend/package-lock.json
+++ b/backend/package-lock.json
@@ -3410,6 +3410,16 @@
       "resolved": "https://registry.npmjs.org/lodash.clonedeep/-/lodash.clonedeep-4.5.0.tgz",
       "integrity": "sha1-4j8/nE+Pvd6HJSnBBxhXoIblzO8="
     },
+    "lodash.foreach": {
+      "version": "4.5.0",
+      "resolved": "https://registry.npmjs.org/lodash.foreach/-/lodash.foreach-4.5.0.tgz",
+      "integrity": "sha1-Gmo16s5AEoDH8G3d7DUWWrJ+PlM="
+    },
+    "lodash.get": {
+      "version": "4.4.2",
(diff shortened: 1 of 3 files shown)


backend/package-lock.json
```diff
@@ -3410,6 +3410,16 @@
       "resolved": "https://registry.npmjs.org/lodash.clonedeep/-/lodash.clonedeep-4.5.0.tgz",
       "integrity": "sha1-4j8/nE+Pvd6HJSnBBxhXoIblzO8="
     },
+    "lodash.foreach": {
```
idjevm, line 3413: Is this a dependancy of mongose-unique-validator? Or do we not need to update the package-lock.json? 

@robinv85 any thoughts?

joshmorel: You'll see the requirements here: Every package referenced as a `require` by packages we install for direct use in our app will then also be at top level of the package-lock.json.
```
    "mongoose-unique-validator": {
      "version": "2.0.3",
      "resolved": "https://registry.npmjs.org/mongoose-unique-validator/-/mongoose-unique-validator-2.0.3.tgz",
      "integrity": "sha512-3/8pmvAC1acBZS6eWKAWQUiZBlARE1wyWtjga4iQ2wDJeOfRlIKmAvTNHSZXKaAf7RCRUd7wh7as6yWAOrjpQg==",
      "requires": {
        "lodash.foreach": "^4.1.0",
        "lodash.get": "^4.0.2"
      }
    },
```

idjevm: Cool! Missed that part haha Thank you so much!

PR #732, opened 2020-06-16
Feature/logged in cookie

### LINKED ISSUE #656: [Desktop][Mobile] persist login session after refresh

joshmorel:

Persist login by embedding JWT in secure httpOnly cookie. Resolves #656 

This is draft for now, @robinv85 I'm looking for feedback on a few things.

#### Backend

I think is good. 

A few notes:

* The age is set to 7 weeks, I will ask the PMs more about this later.
* There is an Auith0 logout but I don't think we should use it here. This would be more for a "log me out of all devices" feature, no? 
* Don't think we need to send back the jwt token anymore as part of response so should I take that out now? Or no harm so just leave it. 

#### Client

Definitely more things to address here.

* Getting the current user for `initAuth` is I'm pretty sure the right move here. Just there is a brief flash in the header until the user is fetched. If you have a suggestion let me know or leave it til post soft launch?
* Only thing for the above, errors are ignored, which makes sense if there is no cookie. But I'm not sure if/how we should be handling other kinds of errors.
* Logout, I used redux here similar to App component, even though I know were trying to move away from it. What I think however we should do is instead of a page just add it as a method on the header. While it's waiting for the clear cookie response we can use some kind of loading thing to disable other header buttons then redirect to home. If you have another suggestion let me know.
* On handling logout errors if they happen to occur now that the client is dependent on server with httpOnly cookie, I guess it can be addressed with a resolution to above point. An error message?


robinv85:
Hey @joshmorel like I've said on the github issue... I think this might not work well because of the JWT expiration, the expiration time can be configured on Auth0

These token expiration times are short because it's more complex to invalidate a token if something goes wrong and risky to get the token in the wrong hands. 

We also don't have token refresh in our API right now to renew the tokens...

I think we'll have to add that and more security like described here https://security.stackexchange.com/questions/133787/securely-using-jwts-with-csrf-protection-and-refresh-tokens

joshmorel:
Regarding the two main things that seem to be under consideration in the posting:

1. JWT refresh token. Should I implement this now or just align refresh time to JWT expiration of 24 hours? Obviously we'll need it eventually if we want to keep people signed in longer than 24hr but could be a later PR, no?
2. CSFR. The posting is almost 4 years old. Now all [modern browsers support](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite) `SameSite=Strict` so aren't we covered from this perspective for our app in 2020?

If there's anything else I'm missing let me know or we can discuss in `#mvp_authentication`.

Feedback on the f/e questions is also appreciated.

But anyways this is still draft - I found another issue looking into now.

joshmorel:
Another thing @robinv85 will the Native platforms handle the httpOnly cookie similarly or will they need to use authorization header & some other mechanism for persistence?

I adapted my approach from this https://github.com/fastify/fastify-jwt#example-using-cookie . `fastify-jwt` will work now with cookie, but also authorization header if there is no cookie. If we do need to support the later I just need to tweak the function used here to handle both: https://github.com/FightPandemics/FightPandemics/blob/1b56fd10a3bdb118a7d224a87cb168a008dd4b75/backend/lib/utils.js#L16

robinv85:
> 
> 
> Regarding the two main things that seem to be under consideration in the posting:
> 
>     1. JWT refresh token. Should I implement this now or just align refresh time to JWT expiration of 24 hours? Obviously we'll need it eventually if we want to keep people signed in longer than 24hr but could be a later PR, no?
> 
>     2. CSFR. The posting is almost 4 years old. Now all [modern browsers support](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite) `SameSite=Strict` so aren't we covered from this perspective for our app in 2020?
> 
> 
> If there's anything else I'm missing let me know or we can discuss in `#mvp_authentication`.
> 
> Feedback on the f/e questions is also appreciated.
> 
> But anyways this is still draft - I found another issue looking into now.
> 
> **Edit** - looked more into browser support. IE11 doesn't render on our site anyways so it's really only a concern if the user is using a non-upgraded browser. caniuse states UC Android/Opera Mini doesn't support but they are listing old versions, so as far as I can tell the latest versions support it.

Hey @joshmorel yeah seems ok to just use the same expiration time for now.
This fixes mostly the page refresh for now which will help in development and we can work on token refresh and such later. I've been reading a bit on Auth0 as well and there are some different approaches possible as well, also with access tokens / id tokens etc. 

About CSFR, indeed that example was just a quick google result and it's not an ideal solution, there are other ways to handle that and I think the cookie `SameSite=Strict` would probably be sufficient, we don't have to over complicate things either

backend/lib/plugins/auth.js
```diff
@@ -53,6 +86,16 @@ const authPlugin = async (app) => {
     }
   });
 
+  /* eslint-disable func-names */
```
robinv85, line 89: why did you need the ESLint disable? Seems like we already have a lot of anonymous callback functions in other places

client/src/actions/authActions.js
```diff
@@ -1,18 +1,32 @@
...
-    if (token) {
-      dispatch({ type: AUTH_SUCCESS, payload: { token } });
-    }
+// Token stored in httpOnly cookie set/cleared by server
+export const initAuth =  () => {
+  return async (dispatch) => {
+    try {
+      const { data: user } = await axios.get("/api/users/current");
+      dispatch({ type: SET_USER, payload: { user } });
+    } catch {} // TODO: consider are there any errors to handle?
```
robinv85, line 14: well request could always go wrong so might be good to display an error
we don't really have more application-level error display in the UI I think
we could use some kind of toast / popup message which would be easy to implement everywhere
the least you could do is log the error here to the ocnsole

backend/lib/plugins/auth.js
joshmorel: Working on a few front-end improvements now. `func-names` only errors for non-arrow functions as far as I can tell. I need non-arrow functions here for `this`.

client/src/actions/authActions.js
joshmorel: There is an error with no cookie (logged out) - 401 Unauthorized - but this is expected here so I don't want to do anything about. But maybe handle other errors a certain way? Maybe on the backend if no cookie we return success with {user: null} because it isn't an actual error so it's handled similar to this video on front-end: https://www.youtube.com/watch?v=jsaOTcBe-dM

joshmorel:
@robinv85 the thing about the double cookie submit though, I'm wondering if that is required so we're not dependent on server to logout. It does seem complicated however. Maybe we are okay without it for soft launch and we can implement pre full MVP or shortly after?

backend/lib/plugins/auth.js
```diff
@@ -22,10 +23,42 @@ const checkAuth = async (req) => {
   req.userId = mongoose.Types.ObjectId(user[auth.jwtMongoIdKey]);
 };
 
+const cookieMaxAge = 60 * 60 * 24 * 7; // 1 week; TODO: ask about alternate?
+const tokenCookieOptions = () => {
+  let domain;
+  switch (env) {
```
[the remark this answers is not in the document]
joshmorel: `AUTH_APP_DOMAIN` or actually better `APP_DOMAIN` would be okay.

Leveraging `AUTH_APP_URL` is a little awkward as it's really a dummy value for our Auth0 scoped parameter. We'd need to use some ugly regex match as below and that's just super hard to reason about and potentially error prone.

I think this will work unless we change the AUTH_APP_URL to something totally unrelated and random. Likely not the right approach.

```
domain: env === "dev" ? "localhost" : AUTH_APP_URL.match(/https?:\/\/(\w*\.){1,}(?<domain>\w*\.\w*)\/$/).groups.domain
```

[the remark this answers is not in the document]
joshmorel: This is done I set as `APP_DOMAIN` with `localhost` as default so all the devs should be good without needing to update anything. The other envs have it set in `main.tf`.

backend/lib/plugins/auth.js
```diff
@@ -22,10 +23,42 @@ const checkAuth = async (req) => {
   req.userId = mongoose.Types.ObjectId(user[auth.jwtMongoIdKey]);
 };
 
+const cookieMaxAge = 60 * 60 * 24 * 7; // 1 week; TODO: ask about alternate?
```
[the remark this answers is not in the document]
joshmorel: Added with 24 hours as default. We can increase this but we either need to increase JWT Auth0 expiry or look into refresh token.

client/src/actions/authActions.js
joshmorel: I updated b/e to send null if no authentication. So the errors should only be for unexpected DB/server things (just logged for now).

joshmorel:
@robinv85 I resolved the issues and also added a double cookie to enable offline logout - similar to this https://medium.com/@thbrown/logging-out-with-http-only-session-ad09898876ba - although the content of the non-httpOnly cookie is really dummy content it doesn't need to match anything.

Both are necessary. If the user logs out the 2nd non-httpOnly one is cleared. The server checks that that exists during `checkAuth` and if not clears the httpOnly token cookie. 

cookie-logout

joshmorel:
The logout is not working in review for some reason. I will put back to draft & look into fixing.

joshmorel:
It seems good now. Besides `localhost` we need the `domain` to clear the cookie. This is what seems like a good solution to a) use httpOnly cookie to secure the JWT from XSS attacks (`SameSite=strict;` protects from CSRF) & b) enable offline logout so we don't need to worry about user being offline or some misc server/network error with regards to logging out.

Probably a good thing (login & logout) to add to cypress e2e testing now that it's setup - I'll look into that over the weekend. 

Certainly option to other idea or suggestion for improvement.

backend/config.js
```diff
@@ -9,8 +9,10 @@ const configData = envSchema({
   schema: S.object()
     .prop("AIRTABLE_API_KEY", S.string())
     .prop("AIRTABLE_BASE_ID", S.string())
+    .prop("APP_DOMAIN", S.string().default("localhost").required())
```
robinv85, line 12: either add a default value or set it required, but not both, this is usually an antipattern.

backend/config.js
```diff
@@ -9,8 +9,10 @@ const configData = envSchema({
...
     .prop("AUTH_APP_URL", S.string())
     .prop("AUTH_CLIENT_ID", S.string().required())
+    .prop("AUTH_COOKIE_MAX_AGE_SECONDS", S.number().default(86400).required())
```
robinv85, line 15: same here

backend/lib/endpoints/users.js
```diff
@@ -12,34 +12,49 @@ const {
...
-      throw app.httpErrors.notFound();
+  app.get(
+    "/current",
+    { preValidation: [app.authenticateOptional] },
+    async (req) => {
+      const { userId } = req;
+
+      // don't error if not authenticated
+      // client always queries on start to check if authenticated
+      if (!userId) return null;
```
robinv85, line 23: not sure if this is a good reason not to return an error? I think it might be better to keep this still as a `401` and make sure that the client handles that response appropriately

backend/lib/plugins/auth.js
```diff
@@ -16,16 +21,46 @@ const cache = new NodeCache({
...
 };
 
 const authPlugin = async (app) => {
+  app.register(fastifyCookie);
+
   app.register(fastifyJwt, {
     algorithms: ["RS256"],
     audience: `${auth.domain}/api/v2/`,
+    cookie: {
+      cookieName: "token",
```
robinv85, line 62: might want to use the `TOKEN_COOKIE ` constant that was declared above

backend/lib/plugins/auth.js
```diff
@@ -16,16 +21,46 @@ const cache = new NodeCache({
...
+  return {
+    domain: appDomain,
+    path: "/",
+    expires: new Date(0),
+  };
+};
+
+const checkAuth = async (req, reply) => {
+  // user has logged out - clear token cookie
+  if (!(REMEMBER_COOKIE in req.cookies) && TOKEN_COOKIE in req.cookies) {
```
robinv85, line 45: so here it's only checked to delete the token cookie if necessary
wondering if it should return in that case instead of also continue below for the JWT check?

backend/lib/endpoints/users.js
joshmorel: Yeah I wasn't sure on this one - my concern was having an errorred returned every time the app starts up. But actually I think what would be better now that we have 2nd cookie - only query /users/current/ on start-up if there is the non-httpOnly remember token present.

backend/lib/plugins/auth.js
joshmorel: I wanted to treat it that if a request is made without the remember cookie it's like making it without a token at all - so delete the **real** token and let jwtVerify do it's thing (i.e. throw error due to no authorization in token cookie). `optionalAuthenticate` routes, such as GET /posts, for example, will hang without other code changes if I just return, so I think the approach is good but if you have another suggestion let me know.

backend/lib/endpoints/users.js
joshmorel: @robinv85 I fixed the 3 other minor issues and implemented an alternate approach for this issue. Take a look and let me know if this is good (it seems to work nicely). The final suggestion  I think the current state is good but open to other suggestions.

PR #734, opened 2020-06-17
Author References Synchronization

### LINKED ISSUE #574: Address sync author element issues
Currently we are syncing author name & type within mongoose SchemaType `set` functions.

After modifying some code so I could post without authentication or location it doesn't seem to be working.  More here: https://github.com/FightPandemics/FightPandemics/pull/526#issuecomment-636324738

I think the reason is the update is async while the setter function is synchronous. If this isn't possible with setters another option is to try out [post-save middleware](https://mongoosejs.com/docs/middleware.html)

When addressing this also sync **author.photo**

Naraujo13:

Synchronize the author's name, photo, and type reference at posts and comments when someone changes any of this information at his profile or at the org profile.

Closes #574 
Supersedes #526 

For performance reasons, we have an author document at Posts and Comments that is a copy of the authors name, type, photo, and location. The three first need to be synced, that is, when the user updates the name, type or photo of his account or of his organization, the author subdocuments for his posts and comments need to also be updated to reflect this change. 

The author's location, per requirements, is NOT updated at the subdocuments when the author changes his location at his profile.


backend/lib/endpoints/users.js
```diff
@@ -61,52 +63,38 @@ async function routes(app) {
...
 
-      return updatedUser;
-    },
-  );
+      // -- Update Author References if needed
+      const { firstName, lastName, photo } = body;
+      if (firstName || lastName || photo) {
+        const updateOps = {};
+        if (firstName || lastName) {
+          updateOps["author.name"] = updatedUser.name;
```
adisen, line 71: Where is updated user initialized?

backend/lib/endpoints/users.js
```diff
@@ -61,52 +63,38 @@ async function routes(app) {
...
+        }
+        if (photo) {
+          updateOps["author.photo"] = updatedUser.photo;
+        }
 
-  app.get(
```
adisen, line 68: Did you remove this by mistake?

backend/lib/endpoints/users.js
```diff
@@ -61,52 +63,38 @@ async function routes(app) {
...
+          Post.updateMany(
+            { "author.id": updatedUser._id },
+            { $set: updateOps },
+          ),
+        );
+        if (postErr) {
+          req.log.error(postErr, "Failed updating author refs at posts");
+        }
 
-  app.post(
```
adisen, line 85: This too.

backend/lib/endpoints/users.js
Naraujo13: Line 65 at 
```
const [updateErr, updatedUser] = await app.to(
        Object.assign(user, body).save(),
      );
```

joshmorel:
The changes requested by @adisen were the same as mine and definitely resolved now so I will go ahead and merge.

PR #739, opened 2020-06-17
Cypress e2e test general structure and folders

galeza:

Add Cypress e2e structure and folders.

Cypress explanation: https://www.notion.so/fightpandemics/Instruction-how-to-instal-and-use-Cypress-bc6b0443f7aa4e93bbbd092d1635b7a6


cypresse2e/cypress.json
```diff
@@ -0,0 +1,4 @@
+{
+    "baseUrl": "http://localhost:3000/",
```
stavares843, line 2: should the baseUrl be the localhost or http://staging.fightpandemics.work/?

cypresse2e/cypress/fixtures/example.json
```diff
@@ -0,0 +1,5 @@
+{
+  "name": "Using fixtures to represent data",
+  "email": "hello@cypress.io",
+  "body": "Fixtures are a great way to mock data for responses to routes"
+}
```
stavares843, line 5: should we remove this file?

galeza: Removed

cypresse2e/cypress.json
```diff
@@ -0,0 +1,4 @@
+{
+    "baseUrl": "http://staging.fightpandemics.work/",
```
idjevm, line 2: Shouldn't it be `https`? Like this it will be redirected to https but probably better just to have https here.

cypresse2e/cypress/support/commands.js
```diff
@@ -0,0 +1,25 @@
+// ***********************************************
+// This example commands.js shows you how to
```
idjevm, line 2: We could probably remove this example file ?

cypresse2e/cypress/plugins/index.js
```diff
@@ -0,0 +1,21 @@
+/// <reference types="cypress" />
+// ***********************************************************
```
idjevm, line 2: Another example file? should we remove or will you add stuff here?

cypresse2e/cypress/support/index.js
```diff
@@ -0,0 +1,20 @@
+// ***********************************************************
+// This example support/index.js is processed and
```
idjevm, line 2: Do we need this example file?

cypresse2e/cypress.json
galeza: Ok, corrected

cypresse2e/cypress.json
galeza: Hi @stavares843 
Finally I put localhost as you requested.
There is an easy way to pass url param while starting cypress like this: $(node_modules location)/node_modules/.bin/cypress run --spec "cypress/integration/e2eTests/**/**/" --config baseUrl=http://localhost:3000/ --headed
we must just add baseUrl=xxx param
We can test in githubActions:
config:
          baseUrl=http://localhost:3000 or
env:
          CYPRESS_baseUrl: http://localhost:3000

https://github.com/cylc/cylc-ui/actions/runs/54043029/workflow

cypresse2e/cypress/plugins/index.js
galeza: Hi @idjevm 
This file will be used to put plugin modules configuration like:
module.exports = (on, config) => {
  /** the rest of your plugins... **/
  require('cypress-log-to-output').install(on)
}
We will add stuff here.

cypresse2e/cypress/support/commands.js
galeza: Hi @idjevm 
This file will be use to to define or overwrite commands. This file is loaded before any test files are evaluated.

cypresse2e/cypress/support/index.js
galeza: Hi @idjevm 
This file will run before every single spec file. We can here put reusable behaviour such as Custom Commands or global overrides -> available to all of spec files.

galeza:
Hi @robinv85 
Probably the best solution is to have data- attribute selectors. I would go this way.

PR #741, opened 2020-06-17
Prettier - Github Action

stavares843:

Adding Prettier on Github Action as a suggestion to format the code automatically.

https://github.com/marketplace/actions/prettier-action

flow
- checks the master branch and formats the code opening a PR automatically once per day - needs to be checked if works


robinv85:
the comma's should be added, see https://github.com/FightPandemics/FightPandemics/blob/master/.prettierrc.yml
Prettier is configured to always add trailing comma's, so it seems like something wasn't working right 
but they should already be added by the pre-commit hooks
i don't think we need a github action as well besides the already configured formatting, but that one needs to work correctly
currently it's just making things more complex and causing more conflicts.

PR #745, opened 2020-06-17
update copy

stavares843:

update copy


README.md
```diff
@@ -161,4 +161,4 @@ When a pull request is merged to master, it will automatically be deployed to th
 
 ### Production
 
-Coming soon...
+http://fightpandemics.com/
```
idjevm, line 164: Here. sorry, yeah we should use https and not http

stavares843: updated

PR #709, opened 2020-06-13
Feature/view organization profile

Chimaobi-Johnson:
### What does this pull request aim to achieve? 💯

1. Completed UI for view organization page and Implemented functionality
2. Implemented functionality for edit organization profile
3. Implemented functionality for edit organization account

UNRESOLVED ISSUES
1. there is currently no error handling code for the backend, so there is no way of seeing an error on the UI
2. the location input in the edit organization account does not display a default value
3. there is no way to view organizations you created (Didnt see a desktop page for this on figma )
4. there is no backend query to see organizations created by the logged in user (I didnt have time to look into this)
5. there is currently no feed in the organization profile.
6. there are currently no svg icons for appstore and googlestore in the organization profile.
7. the location input on create organization profile only accepts address with zipcode

_Please be concise and link any related issue(s):_

### 🚨Before submitting this pull request🚨:

_Please do **NOT** submit this PR if you have not done the following:_

1. I have checked that no one else is working on similar changes.
2. I have read the latest [**README**](https://github.com/FightPandemics/FightPandemics/blob/master/README.md) and understand the development workflow.
3. The name of this branch is: **`feature/<branch_name>`** (need to have cloned the repo not forked).
4. This branch is rebased/merged with the **latest master**.
5. There are no merge conflicts.
6. There are no console warnings and errors.
7. On the right hand side of this PR, I have linked the appropriate issue(s) under "Issues" and added a project progress under "Projects".

PR #749, opened 2020-06-18
Proper location requirements

### LINKED ISSUE #748: Set proper locationSchema requirements
locationSchema should be required on user create (but not user update).

It currently is required on organization create but `    .prop("location", organization.location.required())` this results in the last property of location object being required (`zip`) not the whole object.

```json
{
    "statusCode": 400,
    "error": "Bad Request",
    "message": "body.location should have required property 'zip'"
}
```

On the object itself, `coordinates` and `address` should be required - they are also provided by geoservice while the others may not be.

joshmorel:

Resolves #748 

This relates to an issue I saw when reviewing #709 . Considering making PR into that branch but instead some items are not exactly related as I want to address all of proper location requirements.

Then we can merge master into #709

* address/coordinates should be required on location object
* location object should be required on create user
* location object should be required on create org BUT the current way it was defined was incorrect, it was making the last location property required (zip) not the object

### DIFF
--- a/backend/lib/endpoints/schema/organizations.js
+++ b/backend/lib/endpoints/schema/organizations.js
@@ -28,12 +28,13 @@ const createOrganizationSchema = {
     .prop("global", organization.global.required())
     .prop("industry", organization.industry.required())
     .prop("language", organization.language)
-    .prop("location", organization.location.required())
+    .prop("location", organization.location)
     .prop("name", organization.name.required())
     .prop("needs", organization.needs)
     .prop("ownerId", organization.ownerId.required())
     .prop("type", organization.type.required())
-    .prop("url", organization.url),
+    .prop("url", organization.url)
+    .required(["location"]),
 };
 
 const getOrganizationSchema = {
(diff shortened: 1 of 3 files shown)


backend/lib/endpoints/schema/organizations.js
```diff
@@ -28,12 +28,13 @@ const createOrganizationSchema = {
     .prop("global", organization.global.required())
     .prop("industry", organization.industry.required())
     .prop("language", organization.language)
-    .prop("location", organization.location.required())
+    .prop("location", organization.location)
```
robinv85, line 31: hmm so isn't this required anymore? I might be missing something, thought it would also have to be required in some validation schema's (e.g. create org)

joshmorel: The problem was `organization.location.required()` makes locationSchema.zip required not the whole object. Adding it to required property (see line 37) makes this work as expected while the other primitive required properties also still work.

PR #759, opened 2020-06-20
Fix/linting formatting

robinv85:

double checked the linting &  formatting in the `client` directory
tested CLI, editor, and pre-commit hook with husky/lint-staged and it seems to work fine
there were a bunch of ESLint warnings, an error for unused file and a lot of places where the Prettier formatting wasn't correct so fixed those... i'm not sure how those get through since the pre-commit hook should fix those automatically anyway


robinv85:
@joshmorel alright yeah good point I didn't think about that, for me I think it might not even work if I don;t install everything locally too first ( I'll check).
I will update these merge conflicts and then make a separate PR for the back-end too, this is mostly frontend

PR #766, opened 2020-06-20
created backend endpoint to submit data to SendGrid (will do FE in another PR) 

### LINKED ISSUE #731: Send onboarding email to marketing list
If a non-logged in user provides their email during onboarding add them to the appropriate SendGrid list: https://sendgrid.com/docs/API_Reference/Web_API_v3/Marketing_Campaigns/index.html

idjevm:
In this PR, I am integrating the SendGrid API to create a contact to collect the user's data during Onboarding if they click submit. 
This PR resolves #731


client/src/utils/create-sendgrid-contact.js
```diff
@@ -0,0 +1,31 @@
...
+const sendgrid_list_id = process.env.REACT_APP_SENDGRID_LIST_ID;
+
+let sendgridContactsObject = {
+  contacts: [
+    {
+      list_ids: [sendgrid_list_id],
+    },
+  ],
+  list_ids: [sendgrid_list_id],
+};
```
idjevm, line 15: This is because SendGrid expects this object, specifically on the request body, if it is not exactly like this, then the req will fail. Also, the documentation has the wrong URL, which was annoying and thought it was issues with Auth but Nah, they just don't have really good documentation...

client/src/utils/create-sendgrid-contact.js
```diff
@@ -0,0 +1,31 @@
...
+
+const createSengridContact = async (stepsData) => {
+  Object.assign(sendgridContactsObject.contacts[0], stepsData);
+  try {
+    const res = await axios.put(sendgrid_contacts_url, sendgridContactsObject, {
```
idjevm, line 20: Since I am not using the response here, perhaps I can just have the axios request but we will see if I log something or do something with the response (it returns a pretty useless job id)...

client/src/pages/OfferHelp.js
```diff
@@ -22,11 +22,15 @@ import {
...
   WizardSubmit,
 } from "components/StepWizard";
+import createSendgridContact from "../utils/create-sendgrid-contact";
 
 const INITIAL_STATE = {
   answers: [],
 };
 
+// for now just creating an object to collect the data that I need
+let stepsData = {};
```
idjevm, line 32: Might think of having context instead of this obj, but for now this should work fine.

client/.env
```diff
@@ -1 +1,4 @@
 EXTEND_ESLINT=true
+REACT_APP_SENDGRID_CONTACS_API_URL=TODO
+REACT_APP_SENDGRID_API_KEY=
+REACT_APP_SENDGRID_LIST_ID=
```
idjevm, line 4: Please let me know if you need help finding these values for testing or adding them to the environmental variables.

client/src/utils/create-sendgrid-contact.js
```diff
@@ -2,25 +2,30 @@ import axios from "axios";
...
+    ],
+    list_ids: [sendgrid_list_id],
+  };
+  return sendgridContactsObject;
 };
 
 const createSengridContact = async (stepsData) => {
-  Object.assign(sendgridContactsObject.contacts[0], stepsData);
+  let contacts = populateSendgridContactsObj(stepsData);
+  console.log(contacts);
```
idjevm, line 23: Will remove in the next iteration!

client/.env
joshmorel: Do you recall why we had to check this in? I don't think we'll want to check these REACT_APP... ones if this is the case I think we'll need to handle these similarly to GTM env vars by adding via GitHub Action Workflows. 

Most devs won't need it, so we just want to make sure it does nothing if the env var isn't set.

```
       env:
          GTM_ID: ${{ secrets.PRODUCTION_GTM_ID }}
        run: >
          echo "REACT_APP_GTM_ID=${GTM_ID}" > ./client/.env
```

idjevm: Removed :)

backend/lib/endpoints/sendgrid.js
```diff
@@ -0,0 +1,48 @@
...
+const {
+  config: { sendgrid: sendgridConfig },
+} = require("../../config");
+
+const populateSendgridContactsObj = (body) => {
+  return {
+    contacts: [
+      {
+        email: body.email,
+        ...body.location,
```
robinv85, line 12: doesn't this have to be assigned to `location` property instead of spreading all the props?

client/src/pages/NeedHelp.js
```diff
@@ -25,6 +26,9 @@ const INITIAL_STATE = {
   answers: [],
 };
 
+// for now just creating an object to collect the data that I need
+let stepsData = {};
```
robinv85, line 30: yeah so this is similar to the first version of the steps wizard flow that I made in the first week
because I wasn't sure how that library worked
but since then, I think `props.update()` was simply used to update the state the is kept in the parent component of the wizard, which is a simple solution to manage the state here
so at first sight, you kept that `props.update()` call but also added an object at the top of the file here which is probably not needed

backend/lib/endpoints/sendgrid.js
idjevm: No, it actually needs to be spreaded. There is no location field on sendgrid, so instead I am spreading this object, which contains city, state, and country, and those exist on sendgrid, so they get applied.

client/src/pages/NeedHelp.js
idjevm: So I will be ignoring this for now since I will be only merging the Backend on this PR and wait for Cees to be done with his work for the filters, after that I will just make the api call from onsubmit() of these two pages.

PR #768, opened 2020-06-21
Fix header and footer issue for feed

jyclam:
Fix header and footer issue for feed


ppagarawal:
I confirm that we will have the footer at the bottom of the feed for the soft launch. I have noted the changes which will be addressed in future releases.

PR #769, opened 2020-06-21
Load post comments API integration

### LINKED ISSUE #550: Load Comments.
The idea is that after the user clicks in the comment button, it loads 5 comments and shows a load more button if there are more than 5. This "load more" button will retrieve the next comments from the backend.

`skip` and `limit` properties should be added to the current endpoint.

The new endpoint should be `posts/:postId/comments` and this will be merged in https://github.com/FightPandemics/FightPandemics/pull/378.

### LINKED ISSUE #713: Post text limit and "View More" functionality
env: [https://staging.fightpandemics.work/feed](url)

**For signed in users**
Expected: The posts are limited to 25 words --> User clicks "View More" --> Post shows all text

Fix: 
- Limit posts to 25 words
- When "View More" is clicked --> expand post to show rest of text

**For users unregistered or not logged in**
Expected: The posts are limited to 25 words --> User clicks "View More" --> User is directed to sign up page

Fix: 
- When "View More" is clicked --> user is directed to sign up page

Issue seen in: Desktop and Mobile

### LINKED ISSUE #757: Invalid redirection bug
Clicking on the feed redirects the user to login page.

Expected behaviour:
There should not be any involuntary redirection. I believe that this issue was fixed earlier but now has been reproduced.

Please refer to the gif below to replicate

feedbug

### LINKED ISSUE #798: Viewing a post in feed in a new tab results in uncaught TypeError, post is not rendered
**Steps to reproduce:**

1. Navigate to /feed
2. Open `View More` for any post in a new window

Error shown when replicated on local dev build: https://pastebin.com/mwQNtu9X

Problem might be the same as described here? https://github.com/ReactTraining/react-router/issues/5883

### LINKED ISSUE #800: [Desktop] User cannot post comments on the posts in the Feed Section as there is no post button available. Also user cannot access comment button directly from the post list.
Steps to Reproduce:
Open "https://staging.fightpandemics.work/"
Login using valid email/password
Click on 'Feed' -> Comment on any Post
The comment would not get posted as there is no post button available.

Expected Result:
Users should be able to post comments on any post.

Actual Result:
The user is not able to post comments on any post.

Comment Bug

BMariscal:
Addresses these issues:
Closes #550
Closes #713
Closes #757
Closes #798
Closes #800


joshmorel:
Hey @BMariscal thanks for this. I'll take more of a look at the code tomorrow but one UX/UI comment for now, if it's out of scope & a different PR let me know. 

The comment button in /feed - it acts kind of like a like button, which doesn't make sense. Either it should just be a count OR a button where clicking takes you into comment within the PostPage.

BMariscal:
@joshmorel  thanks! I'll make it redirect to the comments in the post's page

joshmorel:
@BMariscal for logged in users this is looking good.

For not logged in however - isn't the expected behaviour to resolve issue #713 to redirect them to login?

> For users unregistered or not logged in
> Expected: The posts are limited to 25 words --> User clicks "View More" --> User is directed to sign up page
> 
> Fix:
> 
>     When "View More" is clicked --> user is directed to sign up page
>  

Right now it is going to the post:
image

Also on the feed I get a bunch of errors in the log - because of https://github.com/FightPandemics/FightPandemics/blob/52dcf9553e3e3747e71da562d113e2a60b623bca/client/src/components/Feed/Post.js#L122  (the route requires auth). I think we should just skip this if `!isAuthenticated`

BMariscal:
@joshmorel I'm waiting for a confirmation on this but I think the expected behavior (I've asked before) is for anyone, registered or not, to be able to see the post's full content and to not be redirected to the sign in page. I do need to fix the access to the comments for users that are not logged in, thanks for pointing that out!

client/src/components/Feed/NestedComments.js
```diff
@@ -106,13 +106,13 @@ const NestedComments = ({ isAuthenticated, comment }) => {
...
 
-  const commentActions = [
-    renderTimeStamp,
-    renderLikeButton(),
-    renderReply(),
-    renderNumLikes(),
-    renderReplyInput,
-  ];
+  // const commentActions = [
+  //   renderTimeStamp,
```
BMariscal, line 110: commented out for now to remove replies/likes for comments.

BMariscal:
@joshmorel Product got back to me and you're correct! Users who are not registered should not have access to the post's full content. THANKS!!  I'm working on the implementation rn.

BMariscal:
Including this ticket as well: https://github.com/FightPandemics/FightPandemics/issues/757

client/src/components/Feed/Post.js
```diff
@@ -50,80 +62,137 @@ const Post = ({
...
-          <span className="view-more">View More</span>
-        )}
-      </div>
+      {post && post.content.length > CONTENT_LENGTH &&
+        < div onClick={onClick}>
+          {isAuthenticated ?
+            (loadContent ?
+              (<span className="view-more" > View Less</span>) :
+              (<span className="view-more" > View More</span>)) :
+            (<span className="view-more" > View More</span>)
```
idjevm, line 187: Do we need two view more here?

client/src/components/Feed/Post.js
```diff
@@ -33,12 +37,20 @@ const Post = ({
...
   const { postId } = useParams();
+  const history = useHistory();
   const [showComments, setShowComments] = useState(false);
   const [copied, setCopied] = useState(false);
-  const [currentPost, setCurrentPost] = useState({});
+  const [firstLoad, setFirstLoad] = useState(true);
+  const [loading, setLoading] = useState(false)
+  const [loadVisibility, setLoadVisibility] = useState(true)
   const [modalVisibility, setModalVisibility] = useState(false);
+  const [comment, setComment] = useState([]);
```
robinv85, line 51: looks like a lot of state here, probably better to replace multiple state hooks with `useReducer`

client/src/components/Feed/Post.js
```diff
@@ -50,80 +62,137 @@ const Post = ({
...
+    let response;
+    const postId = post._id;
+
+    const skip = 0;
+    const endPoint = `/api/posts/${postId}/comments?limit=${limit.current * page.current}&skip=${skip}`;
+
+    try {
+      response = await axios.get(endPoint);
+    } catch (error) {
+      console.log({ error });
```
robinv85, line 88: need to display or handle an error appropriately?

client/src/components/Feed/Post.js
```diff
@@ -50,80 +62,137 @@ const Post = ({
...
+    }
+    if (response && response.data) {
+      const currentComments = response.data
+      const previousComments = [...comments.current]
+      let allComments = [...previousComments, ...currentComments]
+        .sort((a, b) => (a.createdAt > b.createdAt) ? -1 : 1)
```
robinv85, line 94: when loading more pages of comments, aren't those already sorted server-side?

client/src/components/Feed/PostSocial.js
```diff
@@ -42,71 +48,107 @@ const PostSocial = ({
...
-      <SvgIcon src={commentGray} className="social-icon-svg" />
-    ) : (
-      <SvgIcon src={comment} className="social-icon-svg" />
-    );
+    return showComments ||
+      numComments > 0 ||
+      (history &&
+        history.location &&
+        history.location.state &&
+        history.location.state.comments) ? (
```
joshmorel, line 61: Curious why the 3rd condition to display the blue heart icon. The first two make sense, but the 3rd using the history state, is hard to understand the rationale for. Maybe worth a comment? 
```
history &&
        history.location &&
        history.location.state &&
        history.location.state.comments
```

Also, FYI & optional, on client only we can use safe navigation operator for shorter code, e.g.: `history?.location?.state?.comments`

client/src/components/Feed/Post.js
BMariscal: thanks!

client/src/components/Feed/Post.js
BMariscal: Thanks! This is super helpful!

client/src/components/Feed/PostSocial.js
BMariscal: I thought optional chaining was a Typescript thing. I didn't know you could use it in JavaScript. The last time I used JS was 2.5 years ago or so. This is super neat. Thanks!

robinv85: yeah it's ECMAScript and available in our current create-react-app version
but not on the backend with our Node.js version, or we would have to add babel transpilation

client/src/components/Feed/Post.js
BMariscal: It is sorted. Thanks for bringing this up. I was adding the newly created comment to the comments list in the wrong place.  I've removed the sorting.

BMariscal:
omg omgomg!!

PR #789, opened 2020-06-23
Cypresse2e github action configuration file [do not merge now]

galeza:

First draft for Cypress e2e github actions

- I used: wait-on: http://localhost:3000
- to check - what we should have in the start/run 

- [] The name of this branch is: **`feature/<branch_name>`** (need to have cloned the repo not forked).


.github/workflows/review_cypress_e2e_tests.yml
```diff
@@ -0,0 +1,29 @@
+name: Cypress end-to-end tests
+
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - backend/**
+      - client/**
```
stavares843, line 9: would the idea to be running the tests on each PR? for that seems it needs

```
on:
  pull_request:
```

.github/workflows/review_cypress_e2e_tests.yml
```diff
@@ -0,0 +1,29 @@
...
+  cypress-run:
+    runs-on: ubuntu-16.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
+      - name: Cypress run
+        uses: cypress-io/github-action@v2
+        with:
+          browser: chrome
+          headless: true
```
stavares843, line 21: if is headless it doesn't have a graphical user interface, what screenshots are going to use?

.github/workflows/review_cypress_e2e_tests.yml
```diff
@@ -0,0 +1,29 @@
...
+          start: 
+          wait-on: http://localhost:3000
```
stavares843, line 23: the baseUrl says https://localhost:3000/ instead of http://localhost:3000/

.github/workflows/review_cypress_e2e_tests.yml
galeza: @stavares843 you mean:
on:
  pull_request:
    types: [opened]

stavares843: i used 

on:
pull_request:

and worked fine

example 

https://github.com/FightPandemics/FightPandemics-android/pull/38/files

galeza: @stavares843 

ok, so I will ask about it

.github/workflows/review_cypress_e2e_tests.yml
galeza: I'm removing this part

.github/workflows/review_cypress_e2e_tests.yml
galeza: I'm correcting this.

.github/workflows/review_cypress_e2e_tests.yml
```diff
@@ -0,0 +1,25 @@
...
+    runs-on: ubuntu-16.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
+      - name: Cypress run
+        uses: cypress-io/github-action@v2
+        with:
+          browser: chrome
+          working-directory: cypresse2e
+          start: npx cypress run
```
robinv85, line 18: interesting... didn't know Github actions use `npx` ... I guess this always automatically uses the latest version

galeza:
@idjevm 
Hi! 
I changed URL from http to **https** as requested in this PR: #739 
but when I test it locally (using: npx cypress run) I see this error:

_https://localhost:3000/
_We attempted to make an http request to this URL but the request failed without a response.
We received this error at the network level:_
  _> Error: write EPROTO 140516428572184:error:100000f7:SSL routines:OPENSSL_internal:WRONG_VERSION_NUMBER:../../third_party/boringssl/src/ssl/tls_record.cc:242:___

Are you sure that we should use https?

galeza:
Please do not merge now

galeza:
Hi @stavares843 

That's why I didn't want to merge. I see all the time: waiting for check information. Is it possible to force/trigger this actions?

.github/workflows/review_cypress_e2e_tests.yml
```diff
@@ -0,0 +1,23 @@
+name: Cypress end-to-end tests
+
+on:
+  pull_request:
+    types: [opened]
```
stavares843, line 5: seems is not running because of the ''types: [opened]''

i opened a PR with the same github action and removed that line to see if would run and its running

https://github.com/FightPandemics/FightPandemics/pull/911/files

stavares843:
seems is failing because is trying to go to https://github.com/FightPandemics/FightPandemics/cypresse2e/package-lock.json and it returns not found

but theres a package-lock.json file here:
https://github.com/FightPandemics/FightPandemics/blob/master/client/package-lock.json

stavares843:
in case its useful
https://github.com/cypress-io/github-action/issues/59

galeza:
Hi @stavares843 
Could you look at the whole solution?

We have a problem with localhost:
Screen Shot 2020-06-30 at 21 30 40

stavares843:
will take a look

idjevm:
@galeza does that mean that this expects localhost to be running on your machine? did you have it running when you tried this? Also, what is the difference between this and having the Build URL? Will it automatically test for UI/backend stuff? and allow engineers to merge without you guys having to go for every PR and check them ?

idjevm:
I feel like somehow you should be able to use the Build URL rather than local host -> 
image

mannykary:
@idjevm We can't use the build URL because the end-to-end tests are supposed to be run before deployment. Reason being we wouldn't want a deployment to production happen before the end-to-end tests are run, only to find out that the tests are failing, right? The point of end-to-end tests is to catch any issues before deploying.

I mentioned this a few times in Slack, but I think we need to add a new container to our docker compose file to run the end-to-end tests, so that we can just run `docker-compose up` to spin up the end-to-end tests container. Agnieskza mentioned this: https://github.com/cypress-io/cypress-example-docker-compose. See our Slack discussion here: https://fightpandemics.slack.com/archives/C010V0DDHLN/p1593772641049900?thread_ts=1593713310.025000&cid=C010V0DDHLN

idjevm:
> @idjevm We can't use the build URL because the end-to-end tests are supposed to be run before deployment. Reason being we wouldn't want a deployment to production happen before the end-to-end tests are run, only to find out that the tests are failing, right? The point of end-to-end tests is to catch any issues before deploying.
> 
> I mentioned this a few times in Slack, but I think we need to add a new container to our docker compose file to run the end-to-end tests, so that we can just run `docker-compose up` to spin up the end-to-end tests container. Agnieskza mentioned this: https://github.com/cypress-io/cypress-example-docker-compose. See our Slack discussion here: https://fightpandemics.slack.com/archives/C010V0DDHLN/p1593772641049900?thread_ts=1593713310.025000&cid=C010V0DDHLN

That makes total sense, thanks for pointing me to the right direction. So then this has to happen in the local environment as you mentioned, 🤔 Do we will have to put down the service and re-spin it up to actually make sure that the e2e with cypress gets run?

mannykary:
> That makes total sense, thanks for pointing me to the right direction. So then this has to happen in the local environment as you mentioned, 🤔 Do we will have to put down the service and re-spin it up to actually make sure that the e2e with cypress gets run?

I think for now the easiest approach is to just add the e2e container to `docker-compose.yml`. Then in the workflow we would add a job at the very beginning for e2e tests. I think we would need to do something like this:
```
# spin up the containers required to run the app within the build environment (in the background)
docker-compose up -d client backend-service mongo
docker-compose up e2e  # run the end-to-end tests
```

The second command should complete with the appropriate exit code depending on whether the tests pass or fail, since this is used to determine whether the workflow should continue on to the next step or fail right away.

galeza:
Closing. We will proceed with docker-compose solution -> add a cypresse2e container

Issue #545, opened 2020-05-31
Improve verify-email page content/layout


joshmorel:
I don't see any design for this but we need to improve both the layout and content

Currently it is:
 
![image](https://user-images.githubusercontent.com/10546720/83362728-84007880-a361-11ea-9b0f-27ba1a01f43a.png)

**Content:** Change `<h1>Please verify your email</h1>` to:

```html
<h1>Check your inbox</h1>
<p>We just emailed a link to {{ THE@EMAIL }}. Click the link to confirm your account and complete your profile. If you don't see a message within a few minutes please check your spam folder. </p>
```

**Layout:** Center the content horizontally and move the content down on the page so it starts ~25-30% from the top of the page.

Yu-Alexa-Zhao:
Designer here! Few questions: (1) Is this issue for both mobile and desktop? , (2) we have a pop-up design for the mobile version but the content is different, should we change the content to what you have rn?
![Sign up-modal](https://user-images.githubusercontent.com/44085699/83434617-d0eb5a00-a408-11ea-9ecf-36ebb1e9c511.png)

Yu-Alexa-Zhao:
This is a design for email-sign-up design (if it works I can export this to Zeplin). Do you have the "resend email" feature? If we do, I can add a button below and main paragraph
![sign-up-email-verify](https://user-images.githubusercontent.com/44085699/83435917-fc6f4400-a40a-11ea-84e8-49e42528f78e.png)

manualzuru:
@Yu-Alexa-Zhao was this added to Zeplin?

joshmorel:
@manualzuru  I'm pretty sure this can be worked on. We have the design except maybe for resend component - but someone can certainly start working on this.

Yu-Alexa-Zhao:
Yes it was exported to zeplin:https://zpl.io/bJJMeDK

rabbi08:
> I don't see any design for this but we need to improve both the layout and content
> 
> Currently it is:
> 
> ![image](https://user-images.githubusercontent.com/10546720/83362728-84007880-a361-11ea-9b0f-27ba1a01f43a.png)
> 
> **Content:** Change `<h1>Please verify your email</h1>` to:
> 
> ```
> <h1>Check your inbox</h1>
> <p>We just emailed a link to {{ THE@EMAIL }}. Click the link to confirm your account and complete your profile. If you don't see a message within a few minutes please check your spam folder. </p>
> ```
> 
[...]

PR #794, opened 2020-06-23
Verify Email Page

bittricky:
This pull request introduces changes that will address issue #545. 

#545 

- [X] The name of this branch is: **`feature/<branch_name>`** (need to have cloned the repo not forked).


client/src/hooks/reducers/authFormReducer.js
```diff
@@ -19,6 +19,7 @@ export const authFormReducer = (state, action) => {
   switch (type) {
     case AUTH_FORM_LOGIN:
     case AUTH_FORM_SIGNUP:
+      return { ...state, email: payload.email };
```
adisen, line 22: What's up with this?

client/src/pages/VerifyEmail.js
```diff
@@ -1,11 +1,107 @@
...
+            </p>
+          </VerifyDirectionsContainer>
+        </div>
+      </VerifyEmailRightContainer>
+    </VerifyEmailContainer>
   );
 };
 
-export default VerifyEmail;
+const mapStateToProps = ({ state }) => ({
```
joshmorel, line 103: I get this error as is so I think this needs to be `session` instead of `state`

image

```suggestion
const mapStateToProps = ({ session }) => ({
```





```suggestion
const mapStateToProps = ({ session }) => ({
```

client/src/pages/VerifyEmail.js
```diff
@@ -1,11 +1,107 @@
...
+  email: state.email,
```
joshmorel, line 104:
```suggestion
  email: session.email,
```

idjevm, line 104: This is returning undefined when I tried to sign up. Might have to resolve this before merging this.

idjevm: Also, this caused to not show anything on the verify email page.

client/src/pages/VerifyEmail.js
```diff
@@ -1,11 +1,107 @@
 import React from "react";
+import { connect } from "react-redux";
```
idjevm, line 2: Is there any chance you could implement `Context` instead of Redux? We are trying to move away from Redux... @robinv85 Any thoughts here?

client/src/pages/VerifyEmail.js
idjevm: Yup same error on my end.

client/src/hooks/reducers/authFormReducer.js
bittricky: I needed to access email from state to display it on the page. In the login component we were saving email to state. Wanted explicitly return it here.

client/src/pages/VerifyEmail.js
bittricky: I will revisit it and get it resolved.

bittricky: Updates the connection to session in commit 427850d

client/src/pages/VerifyEmail.js
bittricky: Updates the connection to session in commit 427850d

client/src/pages/VerifyEmail.js
bittricky: Updates the connection to session in commit 427850d

client/src/pages/VerifyEmail.js
robinv85: this example is simply getting the state that's already in redux
with the `mapStateToProps(({ session }) =>...` below
so I think it's out of scope to request refactoring the global state management for this PR :)

client/src/pages/VerifyEmail.js
```diff
@@ -1,11 +1,107 @@
...
+
+const VerifyEmail = ({ email }) => {
   return (
-    <div className="text-center">
-      <h1>Please verify your email</h1>
-    </div>
+    <VerifyEmailContainer>
+      <VerifyEmailLeftContainer>
+        <EnvelopeImageContainer>
+          <img className="EnvelopeSVG" src={emailVerify} alt="" />
```
robinv85, line 84: why empty alt tag, maybe just type "verify email" which can help for accessibility

idjevm: True. Missed this. Yeah type something as Robin said. That will also avoid having more console warnings

client/src/pages/VerifyEmail.js
robinv85: `isAuthenticated` simply checks if user has logged in and got a token (which is now also stored in a cookie but I don't think this is related here, unless somehow loading the cookie doesn't load the other data such as `emailVerified` which means that has to be fixed somehow).
but there are additional checks to do, if the email verification check is not handled correctly it might be related to the value in the redux store, is this fixed now @joshmorel ?

client/src/pages/VerifyEmail.js
bittricky: @robinv85 @idjevm added a description to the alt attribute in 27a6b82

PR #799, opened 2020-06-23
handle auth error edge case

### LINKED ISSUE #796: Skipping user profile completion, user not created but cookies persist
So when they refresh browser or go to /feed the user isn't found. For feed this results in forbidden error: 

image

joshmorel:

Resolves #796 

This was an edge case after initial sign-up where the Auth0 user is authenticated but Mongo user not created yet. If user closes the tab or navigates with the browser bar to let's say /feed they still have log-in cookies but there is no user so we get an error.

In first commit I added at least some error message instead of "Something went wrong" to make other issues that may occur easier to troubleshoot:

image

2nd commit includes setting error state if an error is caught during `initAuth` and that error is checked in RouteWithSubRoute redirect logic - logging them out in such a case.

Open to alternate approaches.

### DIFF
--- a/client/src/actions/authActions.js
+++ b/client/src/actions/authActions.js
@@ -2,8 +2,9 @@ import axios from "axios";
 import { 
   checkRememberCookie,
   clearRememberCookie,
-} from "utils/cookie"
+} from "utils/cookie";
 import {
+  AUTH_ERROR,
   AUTH_LOGOUT,
   SET_AUTH_LOADING,
   SET_USER,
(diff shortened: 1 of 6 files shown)


joshmorel:
Thanks @robinv85. The global messaging is a good idea. Should we create an issue for this or discuss with product/design? 

Also, I'm a little unsure of the redirection logic handling in RouteWithSubRoute. It seems like it's getting convoluted and hard to reason about - might be a good thing to look into refactoring but not sure the best approach?

I'll merge this one as-is for now so at least we're handling this error.

PR #802, opened 2020-06-23
Fix Header and MobileTabs positioning

jyclam:

Fix Header and MobileTabs interaction so spacing is consistent across all pages whether the page has just a Header or Header and MobileTabs both.


ppagarawal:
@jyclam Can you please link the ticket so that I can see the ticket attached. The context just wend off my mind :)

Issue #801, opened 2020-06-23
Org. Profile- UI + Functionality Updates


vikjoshi:
On the Create Org Page, let's make the following updates: 
![image](https://user-images.githubusercontent.com/41554004/85470510-fda41480-b563-11ea-9360-eca58ed97933.png)

1. Add all industries from the "Soft Launch MVP- Industry List" tab into the "Industry" dropdown field from this Google Doc- https://docs.google.com/spreadsheets/d/1kAfWucnSdnbuqMF8qBrql7dDCRRujsvslr3dB4UwHoY/edit?copiedFromTrash#gid=795292381
2. Move the checkboxes for Privacy and Terms of use to be above the "Create Profile" button and make them mandatory fields
3. Remove "notion" as default from the Org. Name field. 
4. For Email field, we can leave a default to clarify the format but let's use "example@example.com" and remove "this is a fake"

PR #803, opened 2020-06-24
Org profile ui func

cvaldez98:

Changes the organization profile form
closes https://github.com/FightPandemics/FightPandemics/issues/801

### DIFF
--- a/client/src/pages/CreateOrganizationProfile.js
+++ b/client/src/pages/CreateOrganizationProfile.js
@@ -37,9 +37,7 @@ import {
   initialState,
 } from "hooks/reducers/organizationReducers";
 import axios from "axios";
-import {
-  inlineLabelStyles,
-} from "constants/formStyles";
+import { inlineLabelStyles } from "constants/formStyles";
 
 const { type, industry } = createOrganizationProfile;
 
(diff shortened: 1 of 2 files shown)


client/src/pages/CreateOrganizationProfile.js
```diff
@@ -156,7 +154,7 @@ const CreateOrgProfile = (props) => {
             <Input
               type="text"
               required
-              placeholder="notion"
+              placeholder=""
```
idjevm, line 157: Perhaps we should add a different place holder text instead of having it empty @vikjoshi ?

client/src/pages/CreateOrganizationProfile.js
```diff
@@ -327,6 +312,19 @@ const CreateOrgProfile = (props) => {
               <Link to="/terms-conditions">Terms and Conditions</Link>
             </StyledCheckbox>
```
idjevm, line 313: You have to add the rule to this to be required as per the issue. This two checkboxes has to be checked before the user can create a profile.

client/src/pages/CreateOrganizationProfile.js
vikjoshi: @idjevm , the field is "Organization Name" which is pretty self-explanatory so my thinking is that we don't need any placeholder text there

idjevm: That makes sense, then I would probably suggest to keep consistency and also remove the place holder for the email the `example@example.com`, and use email validation util to make sure that the user inputs a valid email, the validate email util also display an error message I believe.

idjevm: @vikjoshi Let us know about my last comment ^.

vikjoshi: @idjevm agree with your last comment- let's go ahead and remove the placeholder text for email and use error handling/email validation for error messages instead. Thank you!

vikjoshi:
Thanks @cvaldez98 @idjevm @MohamedEl1 for all the great work! Tested in staging and it looks good to me

Issue #787, opened 2020-06-22
user should be able to logged in using <enter> key also.


rabbi08:
Steps to reproduce:

1. Open "https://staging.fightpandemics.work"
2. Click on 'Login'
3. Enter valid email/password used during SignUp
Press <enter>
You will observe nothing happen

So then. 4. Click on 'Sign In' button

User is successfully logged in.

Issue: User should be able to SIGN IN using <enter> key along with clicking on 'Sign In' key.
Please note that <enter> key does not work during sign ups too.

idjevm:
Reopening this one @adisen since we would like maybe to find a way to do this without messing up the styled, let's continue the work on #937

tejasm2:
Hey @rabbi08 , is this issue implemented already or still needs to be worked on?

rabbi08:
> Hey @rabbi08 , is this issue implemented already or still needs to be worked on?

Yes this issue still exist and it should be fixed.

adisen:
Yes. The PR has been since. No movement 
https://github.com/FightPandemics/FightPandemics/pull/937

guvarallo:
Hey @adisen is there anyone working on this? If not I can do it.

adisen:
@guvarallo No. I was supposed to work on it. But couldn't. So you can sure pick it up. 
Tag me in the PR when you have something

rabbi08:
Verified in STG, working fine:
User is able to login using <enter> key, Refer below GIF
https://share.getcloudapp.com/2Nuj5K7E

User is able to 'Join Now' using <enter> key, Refer below GIF:
https://share.getcloudapp.com/2Nuj5KWg

PR #813, opened 2020-06-24
Feature submit on enter

adisen:
To allow login and signup form to be submitted on "Enter"

Issue [787](https://github.com/FightPandemics/FightPandemics/issues/787)


adisen:
> Thanks @adisen .
> 
> Functionality works but we need keep the existing design for disabled buttons, at least `background-color: rgb(66, 90, 242);` is the main difference I see.
> 
> Also there's some other unintended affect of changing from using button from "antd-mobile" to "antd" button here e.g (left is this PR, right is staging):
> 
> image
> 
> 
> 
> 
> 
> 

I'll look into it. "antd-mobile" doesn't support the kind of button needed and I also feared some of the other Buttons have broken. Fixed the one I saw.

client/src/components/Button/BaseButton.js
```diff
@@ -67,6 +67,7 @@ const BaseButton = styled(Button)`
       pointer-events: none;
     `}
 
+
```
idjevm, line 70: remove this extra line

adisen: Okay

idjevm: oh nooo, now this is merged into master. It's fine though. but please remove extra white lines next time.

idjevm:
@joshmorel I think commits like this one, it will be really better to have squash and merged. We have 7 commits that are irrelevant to the actual change that was made.

adisen:
Okay @idjevm. Noted

PR #848, opened 2020-06-26
use Individual consistently as type

### LINKED ISSUE #846: Filter for provider=Individuals does't work due to spelling mismatch
Should be "Individual" but type `IndividualUser` is lowercase unlike `OrganizationUser`  so need to update but then also update staging DB.

joshmorel:

Resolves #846 

I think ideally we'd have a common types/constants among F/E & B/E but I guess this would have to be an NPM package. Any thoughts on this @robinv85 

I changed IndividualUser type to "Individual"  (uppercase) for consistency with OrganizationUser. I will update staging DB once this is merged. Interestingly, post creation would have error if we were using Author.js but we're not for anything. @joaofnds do you recall why we couldn't use the subschemas?

### DIFF
--- a/backend/lib/models/IndividualUser.js
+++ b/backend/lib/models/IndividualUser.js
@@ -1,7 +1,7 @@
 const { Schema } = require("mongoose");
 const { model: User } = require("./User");
 
-const INDIVIDUAL_USER_TYPES = ["individual"];
+const INDIVIDUAL_USER_TYPES = ["Individual"];
 
 function fullName(firstName, lastName) {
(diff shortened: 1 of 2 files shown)


joshmorel:
> @joshmorel yeah that works, I've used that approach before, but it was with a private repo, or with an organization on npm, nowadays there are probably a lot more possibilities, even Github also has a repository.
> Not sure if we could a local folder as a dependency, that could probably even work too and a relative path in the package.json

I'll create a different issue (non urgent) and we can determine the best approach. I'll go ahead and merge this one to fix the issue.

PR #849, opened 2020-06-26
Displaying the Organizations list in the create post modal

nr431207:
#### displaying the organizations list in the create post modal and removing the default `notion` lists :

Screen Shot 2020-06-26 at 1 39 24 PM
_Please be concise and link any related issue(s):_
#738 

- [x] I have checked that no one else is working on similar changes.
- [x] I have read the latest [**README**](https://github.com/FightPandemics/FightPandemics/blob/master/README.md) and understand the development workflow.
- [x] The name of this branch is: **`feature/<branch_name>`** (need to have cloned the repo not forked).
- [x] This branch is rebased/merged with the **latest master**.
- [x] There are no merge conflicts.
- [x] There are no console warnings and errors.
- [x] On the right hand side of this PR, I have linked the appropriate issue(s) under "Issues" and added a project progress under "Projects".

PR #854, opened 2020-06-27
Fix/Proposal - Profile page create post functionality

kylejvh:
Fixes bug introduced by #849 (I missed this in review and merged it to master)

Prop `user` for `CreatePost` was missing for the profile page.

Side note: 
I looked to see if we had designs laid out for the Profile page in Figma, and there wasn't anything to go off of. 

Does it make sense to be able to create a post from your profile page?

I was thinking it may be better to just refactor the button on the profile page to redirect you to the feed and open up the create post modal there. Thoughts?

### DIFF
--- a/client/src/components/CreatePost/CreatePost.js
+++ b/client/src/components/CreatePost/CreatePost.js
@@ -52,13 +52,13 @@ const Step1 = () => {
 const Step2 = ({ user }) => {
   const createPostContext = useContext(CreatePostContext);
   const { setForm, currentStep, setCurrentStep } = createPostContext;
- 
+
   return (
     currentStep === 2 && (
       <>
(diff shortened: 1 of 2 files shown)


client/src/components/CreatePost/CreatePost.js
```diff
@@ -147,7 +147,7 @@ const CreatePost = (props) => {
     >
       <Wrapper {...props}>
         <Step1 />
-        <Step2 user={props.user}/>
+        <Step2 user={props.user} />
```
idjevm, line 150: No need this space here, check your prettier settings

idjevm: aah nvm, just realized that this is the right settings xD

PR #859, opened 2020-06-27
Edit and Delete a comment in a post

### LINKED ISSUE #486: [WEB] [MOBILE] Edit a comment in a post
**Goal**: To be able to edit a comment by a logged in User.

**Edit a comment**:
As a logged in user, I want to edit a comment so that I can rectify my spellings/errors/thoughts

Web:
image

image

image

Mobile:
image

image

**Acceptance Criteria:**
Only the original poster should be able to edit the comment in a post
The poster should be logged it to edit the comment
The time of update should be stored
When the user clicks on edit, he should be able to see the existing comment and modify it
The comment must accept the max number of characters.
Save button must update the comment
The user should be able to edit the comment any number of times
If a user has logged from multiple devices and editing the same comment, only the latest comment must be visible
Mobile specific: The keypad must appear and accept the max number of characters.
Send button from the keyboard must update the comment

Please feel free to reach out to me in case you need any clarifications.

BMariscal:
Comment author has access to the `Edit` and `Delete` buttons, and can: 

Edit a post a comment in a post:
PATCH request to `/api/posts/${postId}/comments/${commentId}`;

Delete comment in a post:
DELETE request to `/api/posts/${postId}/comments/${commentId}`;

I also included the comment `createdAt` time stamp and the ISO object-to-string mapping. 

Closes #486


joshmorel:
Thanks @BMariscal 

We should have confirmation modal - but also for posts. However this can be a new issue after this is merged.

There is one design from Zeplin. Maybe we can create one component that works for both.

image

https://app.zeplin.io/project/5ecaed1b2522d9280f1a4792/screen/5ed034b05b5521976f3c3223

Also, something to ask product team - when we have comment replies (post soft launch) how do we handle deletes then. I like the idea as you will see often "(deleted)" with the replies still persisting (so the mongo document is updated, not actually deleted). But not sure if only those with replies or all should be like that.... Also if something has been edited an indicator would be good e.g. "21 hours ago (edited)"

I'll look through the code soon.

BMariscal:
@joshmorel Thanks for the feedback! i've alerted @ppagarawal about the changes you've suggested to improve the feature. Unfortunately, I'm juggling a lot of things right now and don't have the bandwidth to work on these feature changes in the near future.

ppagarawal:
> Looks good - although I find this odd. The time ago only updates as you write a new comment. Might be best to make it non-reactive? So basically only updates on page refresh. But probably won't affect most people at first so we can change later based along with further user/PM feedback.
> 
> comment-time-ago-reactive

I recommend to make it non-reactive at the moment. Updates to this on page refresh is fine.

ppagarawal:
> Thanks @BMariscal
> 
> We should have confirmation modal - but also for posts. However this can be a new issue after this is merged.
> 
> There is one design from Zeplin. Maybe we can create one component that works for both.
> 
> image
> 
> https://app.zeplin.io/project/5ecaed1b2522d9280f1a4792/screen/5ed034b05b5521976f3c3223
> 
> Also, something to ask product team - when we have comment replies (post soft launch) how do we handle deletes then. I like the idea as you will see often "(deleted)" with the replies still persisting (so the mongo document is updated, not actually deleted). But not sure if only those with replies or all should be like that.... Also if something has been edited an indicator would be good e.g. "21 hours ago (edited)"
> 
> I'll look through the code soon.

Please show the message [Deleted] for deleted comments. For the same comment, the user name of the comment-or must be visible.

client/src/components/Feed/NestedComments.js
```diff
@@ -74,62 +108,144 @@ const NestedComments = ({ isAuthenticated, comment }) => {
...
   });
 
+  const handleCommentEdit = (e) => {
+    setEditedComment(e.target.value);
+  };
+
+  const handleSubmit = async () => {
+    let response;
+    const postId = comment.postId;
+    const commentId = comment._id;
```
robinv85, line 153: can use destructuring on 1 line, also to rename like `const { postId, _id: commentId } = comment`
Also I just  realized, sometimes the `_id` is changed on the backend to `id` but I guess this is not consistent for every response

client/src/components/Feed/NestedComments.js
```diff
@@ -74,62 +108,144 @@ const NestedComments = ({ isAuthenticated, comment }) => {
...
+            SET_COMMENT,
+            "commentId",
+            commentId,
+            "comment",
+            response.data.content,
+          );
+          setEditComment(!editComment);
+        }
+      } catch (error) {
+        console.log({ error });
```
robinv85, line 172: can either remove or use `console.error` too, maybe we want to show that something went wrong to the user? 
i often ask this to designers to be more specific about error states in the UI since this is often just left out

client/src/components/Feed/Post.js
```diff
@@ -183,6 +183,35 @@ const Post = ({
...
+    const commentId = comment._id;
+    if (isAuthenticated && comment.author.id === user.id) {
+      const endPoint = `/api/posts/${postId}/comments/${commentId}`;
+
+      try {
+        response = await axios.delete(endPoint);
+      } catch (error) {
+        console.log({ error });
+      }
+      if (response && response.data) {
```
robinv85, line 198: ok so as far as I understand this is not like what I previously explained about "eventual consistency" but it only updated the UI after the request returned successfully
this is not really a problem, it might look less responsive in the UI eventually though, I think it generally looks much faster if the UI is updated first while the request is still in progress and then rolled back if something went wrong

joshmorel:
> lgtm as far as the code is finished, do we want to merge this in its current state and add the other parts later?

@robinv85 I think so... @ppagarawal  - handling of comment deletion I think we can change this when implementing comment replies (post soft launch) no?

client/src/components/Feed/Post.js
joshmorel: I'm going to merge maybe we should open another ticket for eventual consistency

joshmorel: Or if you're all right approve and merge then we can move stuff to other ticket ([delete], eventual consistency, etc).

joshmorel:
Merging so we can test for soft launch - will create new issues for desired modifications.

Issue #858, opened 2020-06-27
[Desktop][Mobile] Add Checkboxes to accept Terms & Conditions and Privacy Policy on Create Individual Profile Page.


tejasm2:
As a user,
Given that I am on the Create your Profile page,
Then I should have an option (checkboxes) to accept Terms & Conditions and Privacy Policy.

Desktop:
![Create Profile- Terms n Condi 0](https://user-images.githubusercontent.com/66698276/85917165-7dd4bd80-b875-11ea-9656-4b843cf47553.png)

Mobile:
![Create Profile- Terms n Condi Mobile 0](https://user-images.githubusercontent.com/66698276/85917458-d1e0a180-b877-11ea-9864-7e7485f6f1ec.png)

idjevm:
Isn't this issue already resolved?? @tejasm2 where did you get this from?

 Where are you checking??

idjevm:
Closing this as this already exists.

idjevm:
Seems like it is still an issue for the user profile! We have this already for ORG profile!

pthoma14:
@idjevm thank you for clarifying this issue will be resolved with the #860 PR. I'm going to remove the "MVP-Critical"  tag

PR #860, opened 2020-06-27
Feature/add tos and privacy policy checkboxes to create user profile page

idjevm:
1. Solves a small bug that it is on the create profile page
2. Adds the checkboxes of privacy policy and terms and conditions for the create profile page
3. Made changes to show `creating profile...` when user clicks create profile
#858

### DIFF
--- a/client/src/pages/CreateUserProfile.js
+++ b/client/src/pages/CreateUserProfile.js
@@ -170,7 +170,7 @@ const CreateProfile = ({ email, history }) => {
     } else if (!conditions) {
       alert("You must agree to our terms and conditions before proceeding");
       return;
-    } else if (!location.address) {
+    } else if (!location) {
       // all location objects should have address (+coordinates), others optional
       return setError(
         "location",
--- a/client/src/pages/CreateUserProfile.js
+++ b/client/src/pages/CreateUserProfile.js
@@ -137,7 +137,6 @@ const CreateProfile = ({ email, history }) => {
   const [location, setLocation] = useState({});
   const [privacy, setPrivacy] = useState("");
   const [conditions, setConditions] = useState("");
-
   const dispatch = useDispatch();
   const {
     clearError,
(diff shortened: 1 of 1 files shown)


client/src/pages/CreateUserProfile.js
```diff
@@ -195,8 +155,21 @@ const CreateProfile = ({ email, history }) => {
...
+    if (!privacy) {
+      alert("You must agree to our privacy policy before proceeding");
+      return;
+    } else if (!conditions) {
+      alert("You must agree to our terms and conditions before proceeding");
+      return;
+    } else if (!location) {
       // all location objects should have address (+coordinates), others optional
       return setError(
         "location",
```
MohamedEl1, line 175: Minor: Maybe we can use multiple condition in ternary operator here instead of if statements or a switch case statement here?

client/src/pages/CreateUserProfile.js
```diff
@@ -178,6 +135,9 @@ const handleCheckboxChange = ([evt]) => evt.target.checked;
 
 const CreateProfile = ({ email, history }) => {
   const [location, setLocation] = useState({});
+  const [privacy, setPrivacy] = useState("");
+  const [conditions, setConditions] = useState("");
+
```
MohamedEl1, line 140: Minor: Extra White Space

jyclam:
LGTM. Existing code could use a DRY refactor but that can be left for cleanup another time

client/src/pages/CreateUserProfile.js
idjevm: Tru tru! thank you :)

client/src/pages/CreateUserProfile.js
idjevm: No necessarily, cuz you are not checking the if the value is one or the other, but unique values and their values..

MohamedEl1: Alright sounds good then!

PR #873, opened 2020-06-27
add safe navigation to user?.organizations?.

### LINKED ISSUE #870: Undefined org error after create profile
Due to no organization sent back after POST /users

Will return empty array

joshmorel:

We get this error after profile creation. I'm trying to figure out the best way for the B/E to always send the same data regardless of login/signup/create/etc but until then if we use the full safe operator like `user?.organizations?.` that is best. @MohamedEl1 do the same when adding the mobile.

Resolves #870 

```
TypeError: "o.organizations is undefined"
    mc Header.js:110
    React 6
    unstable_runWithPriority scheduler.production.min.js:19
    React 6
    unlisten Router.js:33
    r history.js:155
    notifyListeners history.js:173
    notifyListeners history.js:172
    _ history.js:288
    push history.js:369
    confirmTransitionTo history.js:145
    push history.js:350
    e CreateUserProfile.js:217
    s runtime.js:45
    _invoke runtime.js:274
    t runtime.js:97
    Babel 2


<!--### 🚨Before submitting this pull request🚨:-->

<!--_Please do **NOT** submit this PR if you have not done the following:_-->

### DIFF
--- a/client/src/components/Header.js
+++ b/client/src/components/Header.js
@@ -108,15 +108,14 @@ export default ({ authLoading, onMenuClick, isAuthenticated, user }) => {
         <Link to="/create-organization-profile">Add Organisation</Link>
       </Menu.Item>
       <Menu.Divider />
-      {user
-        ? user.organizations.map((organization) => (
-            <Menu.Item>
-              <Link to={`/organization/${organization._id}`}>
-                {organization.name}
-              </Link>
-            </Menu.Item>
-          ))
-        : null}
+      {user?.organizations?.map((organization) => (
+        <Menu.Item>
+          <Link to={`/organization/${organization._id}`}>
+            {organization.name}
+          </Link>
+        </Menu.Item>
+        ))
+      }
 
       <Menu.Divider />
       <Menu.Item>

### CODE REVIEW

File: client/src/components/Header.js [hunk 1]
```diff
@@ -108,15 +108,14 @@ export default ({ authLoading, onMenuClick, isAuthenticated, user }) => {
...
-      {user
-        ? user.organizations.map((organization) => (
-            <Menu.Item>
-              <Link to={`/organization/${organization._id}`}>
-                {organization.name}
-              </Link>
-            </Menu.Item>
-          ))
-        : null}
+      {user?.organizations?.map((organization) => (
```
MohamedEl1, line 111: Looks good to me, whats the difference between the two? the second one checks for organization as well as the user ?

joshmorel: Yes. Basically if `user` is `undefined` OR `organizations` is `undefined` it won't do anything.

It's a [new ecmascript feature](https://github.com/tejacques/es-safe-navigation) so transpiled via webpack to something like `users && user.organizations && user.organizations.map`

PR #887, opened 2020-06-28
Feature/org posts in org profile

### LINKED ISSUE #825: Have posts from organization show up under organization profile
MVP-Critical: As a user associated with an organization, if I go to my organization page, under "My Activity" I should see the posts that anyone in my organization has created. As part of the logic, you should use the Author ID to filter posts for the specific organization

image

Post-MVP Launch: Activity should also show any posts that the user has liked or commented on

B/E Ticket that is complete: #811 

From @joshmorel :For Activity, use posts created by organization (authorId={orgId})

cvaldez98:

Fixes #825 

- [X] The name of this branch is: **`feature/<branch_name>`** (need to have cloned the repo not forked).


client/src/components/Feed/Post.js
```diff
@@ -139,13 +141,13 @@ const Post = ({
...
     limit.current = currentLimit * page;
-  };
+  });
 
   useEffect(() => {
     if (postId) {
       loadComments();
     }
-  }, []); // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [loadComments, postId]); // eslint-disable-next-line react-hooks/exhaustive-deps
```
joshmorel, line 150: I guess these were added with automatic eslint fix on commit? Usually I'd suggest not including unrelated commits but since we added `--fix` as part of lint-staged (not yet working with docker) I'm okay to try to sort this out here. 

This one is creating an infinite loop so we need to use the `eslint-diable-next-line` properly. I'll commit here and keep testing.

```suggestion
  // eslint-disable-next-line react-hooks/exhaustive-deps
  }, []);
```

client/src/components/Feed/Post.js
```diff
@@ -84,7 +86,7 @@ const Post = ({
     }
   };
 
-  const loadComments = async () => {
+  const loadComments = useCallback(async () => {
```
idjevm, line 89: @joshmorel This one most likekly got added cuz of the `--fix`?

client/src/pages/NeedHelp.js
```diff
@@ -153,8 +153,8 @@ const NeedHelp = withRouter((props) => {
 
   useEffect(() => {
     setTransition(!transition);
-  },[]);
-  
+  }, [transition]);
```
idjevm, line 156: Also need to update this one here. This will cause an infinite loop.

client/src/pages/OfferHelp.js
```diff
@@ -177,7 +177,7 @@ const OfferHelp = withRouter((props) => {
 
   useEffect(() => {
     setTransition(!transition);
-  },[]);
```
idjevm, line 180: Also here

client/src/components/Feed/Post.js
joshmorel: Might not be an issue but I'd like to at least understand why it was eslint added it :thinking:

client/src/pages/NeedHelp.js
cvaldez98: Yeah, I messed it up with the ES linter.

client/src/components/Feed/Post.js
idjevm: Yeah I think it is because of this -> 
image

client/src/pages/OfferHelp.js
joshmorel: Yeah this `exhaustive-deps` thing we need to look more into. Either it's something we should ignore completely or we're not using react hooks properly.

client/src/components/Feed/Post.js
idjevm: I mean this warning causes the trigger to --fix and it fixes it and put is back. I think is the way that some engineers use react hooks, it is not exactly the proper way?

joshmorel: Or maybe not. I removed it and didn't get any eslint errors.  Any idea @cvaldez98 ?

cvaldez98: It complained about a missing useCallback dependency when I first added the dependency and the code was already there.

client/src/pages/OfferHelp.js
idjevm: I think it is the latter more than anything else.

galeza:
Hi!

This is my scenario:

1. I created an organisation in my Profile. 
2. Then I select my_organisation from the drop-down:
Screen Shot 2020-06-30 at 12 00 06

3. I added post as an organisation
4. Post is not visible under Profile -> my_organisation

What should be expected behaviour for this feature?

joshmorel:
The reason is - if I set the visibility filter to "City" and the org is in a different city than my individual user I won't see it on the profile to location visibility filters. 

I wonder if we should include that here https://github.com/FightPandemics/FightPandemics/issues/904#issue-647447338

Try creating posts with visibility worldwide and it should work.

joshmorel:
We will address visibility filters on B/E with PR for issue #904

cvaldez98:
Sounds good @joshmorel Can we approve this PR then since we have a separate PR for it?

Issue #655, opened 2020-06-09
Desktop/Mobile web UI- Hide email address on View User profile screen when viewing other user's profile


indirabhowmick:
- [ ] Hide email address on View User profile screen when viewing other user's profile
https://zpl.io/bWzjANx

larryabim:
Am I allowed access to zeplin so I can see the issue

indirabhowmick:
@larryabim - please request for access from @manualzuru 
@robinv85 - Tagging Robin since he is working on this feature.

robinv85:
@indirabhowmick I worked on the view profile for your own user, but we don't have support to view another user profile at the moment, and you can never see the email address.

indirabhowmick:
@robinv85 @larryabim - I agree with @robinv85 . We do not have the design for it. For now, when a user clicks on other user's profile, we can display the view profile page, but hide the email address. Nothing should be editable on that page. 
![image](https://user-images.githubusercontent.com/57240870/84581258-37bb2c80-ad94-11ea-9d87-d8ec9df3c714.png)

Tagging @Yu-Alexa-Zhao @manualzuru since we do not have a design for it.

pthoma14:
@joshmorel I see you mentioned this issue yesterday. Is this issue still open or should it be closed?

pthoma14:
@indirabhowmick is anyone assigned to this issue? Do you believe we can push this UI issue to the real MVP launch instead of the soft launch?

indirabhowmick:
@pthoma14 - functionality has already been implemented. @robinv85 @joshmorel @idjevm - Please confirm

pthoma14:
@indirabhowmick I've confirmed that this functionality is currently working.

joshmorel:
Yeah this one wasn't linked because it was the 2nd tagged after "Resolves" keyword

pthoma14:
Ahh got it. Thanks @joshmorel

PR #889, opened 2020-06-28
Feature/user public profile

### LINKED ISSUE #805: Route for IndividualUser Public Profile
Exclude email

Also need to consider location what to keep private (coordinates) vs public (city/state) - but also address and whether to hide (need IndividualUser field hideAddress).

joshmorel:

Resolves #805  #655 

This includes (3 commits): 

* B/E: updating backend route so email and (conditionally) address is not returned but other required data are
* F/E: re-using `/profile` page for other user's profile (`/profile/:id`) as well as own profile with conditional rendering
* F/E: adding links to author profile from post header (user: `/profile/:id`  or `/organization/:id` depending on type)

I'm still not yet fully up to speed on React so alternate suggested approaches are welcome.

### DIFF
--- a/client/src/components/Feed/Post.js
+++ b/client/src/components/Feed/Post.js
@@ -34,6 +34,8 @@ import {
 import SvgIcon from "../Icon/SvgIcon";
 import statusIndicator from "assets/icons/status-indicator.svg";
 
+const INDIVIDUAL_AUTHOR_TYPE = "Individual";
+
 export const CONTENT_LENGTH = 120;
 const Post = ({
   currentPost,
@@ -42,6 +44,7 @@ const Post = ({
   fullPostLength,
   handlePostDelete,
   handlePostLike,
+  includeProfileLink,
   isAuthenticated,
   loadMorePost,
   numComments,
@@ -75,6 +78,9 @@ const Post = ({
       post.author.name.match(/\b\w/g).join("").toUpperCase()) ||
     "";
 
+  const authorProfileLink = `/${
+    post.author.type === INDIVIDUAL_AUTHOR_TYPE ? "profile" : "organization"
+  }/${post.author.id}`;
   // mock API to test functionality
   /* to be removed after full integration with user api */
   const [shared, setShared] = useState(false);
@@ -205,7 +211,7 @@ const Post = ({
     return !postId && post && isAuthenticated ? (
       <Link
         to={{
-          pathname: `post/${post._id}`,
+          pathname: `/post/${post._id}`,
           state: {
             post: post,
             postId: post._id,
@@ -245,6 +251,10 @@ const Post = ({
         </span>
       }
     />
+  );
+
+  const renderHeaderWithLink = (
+    <Link to={authorProfileLink}>{renderHeader}</Link>
   );
 
   const renderContent = (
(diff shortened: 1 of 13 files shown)


joshmorel:
That's a good option @idjevm I'm open to it - @indirabhowmick what do you think?.

joshmorel:
Weird I get an error (location isn't defined) when visiting an organization page in staging & review although local is okay... There were no changes to organization profile page itself here as this is the issue @cvaldez98 is working on.

indirabhowmick:
@joshmorel @idjevm - I like the idea that @idjevm has proposed.

client/src/components/Feed/Post.js
```diff
@@ -346,7 +362,7 @@ const Post = ({
           }}
         >
           <div className="card-header">
-            {renderHeader}
+            {renderHeader()}
```
robinv85, line 365: Personally I don't really think creating separate functions for parts of the rendered output makes much sense 
this is usually a good idea for things like mapped functions, like `items.map(renderItem)` but otherwise I don't really see an advantage to create multiple separate functions unless they are extracted to another file

joshmorel: Are you speaking more general - all the `renderSocialIcons`, `renderShareModel`, `renderHeader` etc and not just my specific change? So basically - they should be moved to a separate component file or just include within the main structure.

Anyways I thought I had made a change to address this for my specific function but I now I realize probably not.

joshmorel:
@indirabhowmick this is working now so - all individual profiles are profile/:id

@robinv85 - to determine if own user I originally looked into using `session.user` but this became awkward - especially with a new page load because you need to wait until current user is fetched. So instead I just added a field returned from the b/e - `ownUser`.

errors.txt
```diff
@@ -0,0 +1,548 @@
...
...y; city promotion and branding; quality employment; industrial reactivation and sports and economic activity. \n\n \n\nCampaign to foster local consumption and improve the promotion of Barcelona\n\nA major local, national and international communication campaign is also being started to back local consumption and relaunch Barcelona’s international profile.\n\nOn a local and national level the goal is to boost local consumption in the short to mid term. On an international level, the intention is to promote Barcelona abroad as soon as the circumstances allow.\n\n \n\nProposed socio-economic agreement for Barcelona\n\nThe municipal government has also reached out to municipal opposition groups to construct a major socio-economic city agreement with the following aims:\n\nWork to reactivate the economy.\nAgree on action and measures.\nOvercome the crisis as soon as possible.\nMove forward in the transformation to a greener and digitally-based model.\nFurther informationOpen in a new window\n\nLatest update 17/04/2020","post_objective":["Offer"],"post_tags":["Economics"],"post_duration":"Forever","country":["Spain"],"state":["Catalonia"],"city":["Barcelona"],"from_whom":["Goverment"]}.
```
idjevm, line 7: Do we need this file?

idjevm: I think it is just from your local? so let's remove it

joshmorel: Oopsie. Def not. I will rewrite commit so it will be gone from history.

idjevm: Sounds good :) Yeah in this cases comes in handy removing that commit from history.

joshmorel: Oops used the wrong command that completely changed older commit shas. I should have rebased instead I guess.

PR #890, opened 2020-06-28
Add Datadog Fargate task monitoring

### LINKED ISSUE #762: Add Datadog
https://docs.datadoghq.com/integrations/ecs_fargate/#pagetitle

NOTE: not critical for MVP soft launch.

mannykary:

Adds a Datadog container to the Fargate task definition which will monitor the performance of both the client and backend containers. This container is only added in production; the staging and review environments do not require application performance monitoring.

In order to conditionally add the Datadog container to the task definition, I had to refactor the code in `task-definition.tf` to define each container as a Terraform map, rather than JSON. Then it was simply a matter of checking the value of `fp_context` and including the Datadog definition only if `fp_context` is `production`.

I also created a Slack channel called `#performance-monitoring`, where Datadog alerts will be sent to.

Resolves #762

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -11,7 +11,7 @@ variable "fp_context" {
 }
 
 variable "commit_hash" {
-  type = string
+  type    = string
   default = ""
 }
 
@@ -53,6 +53,11 @@ data "aws_ssm_parameter" "logger_host" {
 
 data "aws_ssm_parameter" "logger_port" {
   name = "/fp/logger/port"
+}
+
+data "aws_ssm_parameter" "datadog_api_key" {
+  count = var.fp_context == "production" ? 1 : 0
+  name  = "/fp/datadog/key"
 }
 
 locals {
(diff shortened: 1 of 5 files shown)


mannykary:
Converting to draft as I need to test that the metrics are logged to Datadog and make fixes if necessary.

main.tf
```diff
@@ -11,7 +11,7 @@ variable "fp_context" {
 }
 
 variable "commit_hash" {
-  type = string
+  type    = string
```
idjevm, line 14: Remove extra spaces here.

mannykary: This is Terraform style convention: https://www.terraform.io/docs/configuration/style.html

Running `terraform fmt` automatically formats Terraform code to their convention.

idjevm: Got you! Thank you for clarifying!

mannykary:
This is ready to be reviewed. I have already deployed this to production by adding a `v1.0.2` tag to a commit in this branch, and tested that Datadog is receiving the metrics.

I will add a Slack integration for Datadog, but this shouldn't require any code.

mannykary:
I have decided to implement the other things in another [ticket](https://github.com/FightPandemics/FightPandemics/issues/894) since there is a bit of set up involved for trace performance monitoring of Node.js apps, so I will set this back to ready for review.

Issue #786, opened 2020-06-22
Run husky in docker & local


joshmorel:
We use husky to run eslint before committing but this doesn't work when running docker. Warning:

```
Can't find Husky, skipping pre-commit hook
You can reinstall it using 'npm install husky --save-dev' or delete this hook
```

We want this to work for people working with and without docker. Here is one potential resolution but it might work only for docker: https://github.com/MadeInHaus/react-starter/pull/30/files

Needs to be added to both `client/package.json` & `backend/package.json`

joshmorel:
@robinv85 from what I've been reading it seems the recommended path for a monorepo is to add a root level npm project for husky/lint-staged.

https://github.com/typicode/husky#monorepos
https://www.horacioh.com/writing/setup-lint-staged-on-a-monorepo

This [repo is an example using lerna](https://github.com/sudo-suhas/lint-staged-multi-pkg) but with two non-lerna alternates. [This seems](https://github.com/sudo-suhas/lint-staged-multi-pkg#install-lint-staged-at-the-package-root) like the best choice for us (at least at first). The 3rd one I don't think will work with docker-compose. The main `lerna` option might work with docker-compose with some tweaks but maybe `lerna` is overkill is us?

Everyone would need to install Node.js regardless of if using local or docker set-up but I don't think that's too big a deal.

If you have any other suggestions or thoughts on this let me know.

robvenn:
@joshmorel interesting information, I will look at the links you provided and give it some thought
I am not sure if the long term vision here is to keep the backend and the web frontend in a monorepo, also because there are mobile apps as clients?
About Lerna, I have seen that this is often successfully used for libraries and tools such as Babel and Lodash but I am wondering if this would fit our use case since it's not just a multi-package monorepo but we are running multiple separate Docker services (back- and front-end in the same repo).
Eventually this might also affect the deployment process and such, maybe @mannykary has some valuable insights here.

Last one looks ok, I guess I haven't really thought of using the `package.json` in the root folder like that

joshmorel:
@robinv85 - For `lerna` I think you're right - not just overkill but inappropriate for us.

I'm not sure about the long term vision but in the medium term I can't see this changing anytime soon (~6 months). For one thing, I have a feeling the feasibility of running cypress e2e tests might depend on keeping f/e & b/e together in one repo.

mannykary:
[...]

PR #891, opened 2020-06-28
Feature/post as org

### LINKED ISSUE #874: Make organization author if creating post as organization
We need both backend and frontend changes but backend is predominat issue.

We can split out F/E to different ticket if needed in `posts.js` `app.post(`.

Backend we need to accept organization object with `createPostSchema`. If it's set then we need to use that organization as author instead of authenticated user in . 

Frontend we need actually send organization - set here https://github.com/FightPandemics/FightPandemics/blob/7d220104fa870e5f57ee68c0c024f754d4550944/client/src/components/CreatePost/CreatePost.js#L67 - but not actually sent to B/E.

joshmorel:

Resolves #874.

Instead of write away merging @cvaldez98 right away I do want a review here.

But you can merge this into your other branch for organization activity if you want to because this does enable posting as an org. 

@Naraujo13 for B/E - I considered two other approaches:

1. Passing whole organization to B/E and using that. But this could actually enable someone else to post as an org they don't own via curl/postman/etc so we do need to look up the ownerId in the DB to make sure it matches posting userId
2. If `organizationId` is set use "organizations" virtual then look through the array to achieve the similar. This seemed awkward however.

Open to other suggestions

@nr431207 @BMariscal @robinv85 for F/E this seems okay. I think I'm using context here correctly. The name form for the context was a little weird (we also have formData state) and I considered changing it to just organizationsId but maybe more could be included in there later.

We need to change post edit/delete so owners can edit organization posts. I will make a new issue.

### DIFF
--- a/backend/lib/endpoints/posts.js
+++ b/backend/lib/endpoints/posts.js
@@ -213,6 +213,7 @@ async function routes(app) {
     },
     async (req, reply) => {
       const { userId, body: postProps } = req;
+      // return postProps;
       const [userErr, user] = await app.to(User.findById(userId));
       if (userErr) {
         req.log.error(userErr, "Failed retrieving user");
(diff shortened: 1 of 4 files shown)


client/src/components/CreatePost/CreatePost.js
```diff
@@ -82,7 +83,7 @@ const Step2 = ({ user }) => {
 
 const Step3 = ({ onCancel }) => {
   const { currentStep, setCurrentStep, setPostId } = useContext(
-    CreatePostContext,
+    CreatePostContext
```
idjevm, line 86: no trailing comma?

joshmorel: Yeah this is an issue when I apply prettier --write via docker & didn't catch this mistake. We have `trailingComma: 'all' in root level .prettier.yml. We've got to find a better way to make this work smoothly in docker - see https://github.com/FightPandemics/FightPandemics/issues/786

idjevm: Ohh interesting though! That explains a lot! @CeesWang is probably not running docker, which is why his code gets reformatted on commit and messes some of his stuff...

backend/lib/endpoints/posts.js
```diff
@@ -222,13 +223,35 @@ async function routes(app) {
...
+      const { organizationId } = postProps;
+      if (organizationId) {
+        const [orgErr, org] = await app.to(User.findById(organizationId));
+        if (orgErr) {
+          req.log.error(userErr, "Failed retrieving organization");
+          throw app.httpErrors.internalServerError();
+        } else if (org === null) {
+          req.log.error(userErr, "Organization does not exist");
+          throw app.httpErrors.forbidden();
+        } else if (org.ownerId.toString() !== userId.toString()) {
```
Naraujo13, line 237: Instead of converting both to string we could use the `.equals()` from the ObjectId

idjevm: so from `org.ownerId.toString() !== userId.toString()` to -> `org.ownerId.equals(userId)`? I also think that would be better. I guess we can revisit later.

Naraujo13: Yep, that's correct

PR #898, opened 2020-06-29
Add liked and likes count at GET /posts/:postId

### LINKED ISSUE #843: Include Post liked and likesCount in GET posts/:postId
So F/E work in this PR works #834 

Just like here in GET posts/ https://github.com/FightPandemics/FightPandemics/blob/d959a2ef4e605e827d039caf6e54d6005fb9ff10/backend/lib/endpoints/posts.js#L181

Naraujo13:

Add the following fields to the response of the `GET /posts/:postId` endpoint:
- `liked`: boolean value with whether the user current logged user (if present) already liked this post or not
- `likesCount`: number of likes currently at this post

Both fields were already returned at `GET /posts/`, so this just adds the same structure to the single post response.

Closes #843

### DIFF
--- a/backend/lib/endpoints/posts.js
+++ b/backend/lib/endpoints/posts.js
@@ -352,7 +352,7 @@ async function routes(app) {
       return {
         comments,
         numComments,
-        projectedPost,
+        post: projectedPost,
       };
     },
   );
--- a/backend/lib/endpoints/posts.js
+++ b/backend/lib/endpoints/posts.js
@@ -285,6 +285,7 @@ async function routes(app) {
       schema: getPostByIdSchema,
     },
     async (req) => {
+      const { userId } = req;
       const { postId } = req.params;
       const [postErr, post] = await app.to(Post.findById(postId));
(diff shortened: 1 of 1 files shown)


backend/lib/endpoints/posts.js
```diff
@@ -341,10 +342,17 @@ async function routes(app) {
...
+        ...post.toObject(),
+        liked: post.likes.includes(mongoose.Types.ObjectId(userId)),
+        likesCount: post.likes.length,
+      };
+
       return {
         comments,
         numComments,
-        post,
+        projectedPost,
```
joshmorel, line 355:
```suggestion
        post: projectedPost,
```

Renaming the return element will break the frontend. I think the above fix will work.

idjevm: Haha I thought about this, but wasn't sure.

joshmorel: I'm going to make the commit here and test on review branch. Everything else is good.

PR #913, opened 2020-06-30
Integrate sendgrid on the frontend

### LINKED ISSUE #731: Send onboarding email to marketing list

idjevm:
This PR integrates sendgrid on the frontend :). 
Resolves #731

### DIFF
--- a/backend/lib/endpoints/schema/sendgrid.js
+++ b/backend/lib/endpoints/schema/sendgrid.js
@@ -4,7 +4,9 @@ const { strictSchema } = require("./utils");
 const sendgridContactSchema = {
   body: strictSchema()
     .prop("email", S.string().format("email").required())
-    .prop("location", S.object().required()),
+    .prop("location", S.object().required())
+    .prop("helpType", S.string())
+    .prop("postType", S.string()),
 };
 
 module.exports = { sendgridContactSchema };
--- a/backend/lib/endpoints/schema/sendgrid.js
+++ b/backend/lib/endpoints/schema/sendgrid.js
@@ -6,7 +6,8 @@ const sendgridContactSchema = {
     .prop("email", S.string().format("email").required())
     .prop("location", S.object().required())
     .prop("helpType", S.string())
-    .prop("postType", S.string()),
+    .prop("postType", S.string())
+    .prop("providers", S.array()),
 };
 
 module.exports = { sendgridContactSchema };
--- a/backend/lib/endpoints/schema/sendgrid.js
+++ b/backend/lib/endpoints/schema/sendgrid.js
@@ -4,7 +4,7 @@ const { strictSchema } = require("./utils");
 const sendgridContactSchema = {
   body: strictSchema()
     .prop("email", S.string().format("email").required())
-    .prop("location", S.object().required())
+    .prop("location", S.anyOf([S.object(), S.string()]))
     .prop("helpType", S.string())
     .prop("postType", S.string())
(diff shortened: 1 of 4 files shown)


backend/lib/endpoints/schema/sendgrid.js
```diff
@@ -4,7 +4,10 @@ const { strictSchema } = require("./utils");
 const sendgridContactSchema = {
   body: strictSchema()
     .prop("email", S.string().format("email").required())
-    .prop("location", S.object().required()),
+    .prop("location", S.anyOf([S.object(), S.string()]))
```
idjevm, line 7: The reason I am doing this is because if the location is not set, then it is a string (per the way that it is on initial state) done by @CeesWang.

backend/lib/endpoints/sendgrid.js
```diff
@@ -31,18 +31,20 @@ const getHeaders = () => {
...
-    }
-  });
+  app.put(
+    "/create-contact",
+    { schema: sendgridContactSchema },
+    async (req, res) => {
+      try {
+        let contacts = populateSendgridContactsObj(req.body);
+        axios.put(sendgridConfig.contactsApiUrl, contacts, getHeaders());
+        res.send(`Contact with email=${req.body.email} created successfully!`);
```
idjevm, line 41: respond to fe  with this string.

client/src/pages/NeedHelp.js
```diff
@@ -160,6 +161,13 @@ const NeedHelp = withRouter((props) => {
     setState({ ...updatedAnswers });
     if (key === "email") {
       localStorage.setItem("needHelpAnswers", JSON.stringify(updatedAnswers));
+      if (value) {
```
idjevm, line 164: L64 will check if the value is not falsy, if it is not then  we will make the API request, cuz Skip is being handled on submit and this will execute as well on skip.

client/src/pages/OfferHelp.js
```diff
@@ -184,6 +185,13 @@ const OfferHelp = withRouter((props) => {
     setState({ ...updatedAnswers });
     if (key === "email") {
       localStorage.setItem("offerHelpAnswers", JSON.stringify(updatedAnswers));
+      if (value) {
```
idjevm, line 188: Same as NeedHelp.js comment.

idjevm:
I forgot to handle on the backend what if a string is passed in location, then location will be undefined and won't be able to spread.

backend/lib/endpoints/schema/sendgrid.js
CeesWang: We can set it to an empty object if the location is not provided.

idjevm: Yeah, working around this on the backend, cuz in the frontend, if I do that, it will break the filters you did.

idjevm:
This is good to go :) Tested and works fine with null / empty / falsy values for the other objects. As long as the email is passed, it will create an email ONLY with the email if the other fields are undefined.

idjevm:
Will go ahead and merge this @joshmorel, if you have any comments, I can address them later on. So far for first implementation this is good to go. At least now we will be able to capture the emails of the people who put them during onboarding :D

Issue #871, opened 2020-06-27
Bad request on create org no ownerId


joshmorel:
This is hapening inconsistently and I don't know why.

![image](https://user-images.githubusercontent.com/10546720/85929008-3bb07980-b87f-11ea-8897-9333b917fd5c.png)

Perhaps we can just do this on the backend from req userId.

joshmorel:
Thought's on this one @robinv85 ? Do you think it would be best to move this logic to B/E, `const { userId: ownerId } = req`

PR #922, opened 2020-06-30
Add Edit/Delete menu to organization posts

nr431207:
### What does this pull request aim to achieve? 💯
Adds Edit/Delete menu to org posts in `Feeds` , `Post`, and `Organization Profile` page : 
Screen Shot 2020-07-02 at 5 40 23 PM
Screen Shot 2020-07-02 at 5 40 00 PM
Screen Shot 2020-07-02 at 5 40 12 PM

_Please be concise and link any related issue(s):_
#906 

- [x] I have checked that no one else is working on similar changes.
- [x] I have read the latest [**README**](https://github.com/FightPandemics/FightPandemics/blob/master/README.md) and understand the development workflow.
- [x] The name of this branch is: **`feature/<branch_name>`** (need to have cloned the repo not forked).
- [x] This branch is rebased/merged with the **latest master**.
- [x] There are no merge conflicts.
- [x] There are no console warnings and errors.
- [x] On the right hand side of this PR, I have linked the appropriate issue(s) under "Issues" and added a project progress under "Projects".

PR #926, opened 2020-06-30
Feature/only update own org

### LINKED ISSUE #919: Require user to be owner to update organization profile
Current anyone can do it.

Also on f/e hide edit & create post buttons if not owner.

joshmorel:

Resolves #919 #871

This does a few related things. I guess best practice would be to make smaller PRs but in this case with looming launch it seems a little more efficient to do 1.

1) Instead of using ownerId from F/E, use auth user's id on B/E
2) Require user to be owner to allow update/delete
3) Only show edit/post buttons on org profile if owner

 I included update to delete (2 above) even though we aren't using it anywhere on F/E. I think that's fine but if you want me to undo I can (revised as separate commit).

### DIFF
--- a/backend/lib/endpoints/organizations.js
+++ b/backend/lib/endpoints/organizations.js
@@ -23,14 +23,26 @@ async function routes(app) {
       schema: deleteOrganizationSchema,
     },
     async (req) => {
-      // TODO: make sure user can only delete organizations they own
-      const result = await Organization.findByIdAndRemove(
-        req.params.organizationId,
-      );
-      if (result === null) {
-        return new httpErrors.NotFound();
+      const {
+        params: { organizationId },
+        userId,
+      } = req;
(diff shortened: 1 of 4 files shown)


cvaldez98:
Is the ability to edit a post only available from the User profile page? Because if I create a post from an org, then click on the post I made, I won't have the option to edit it.

joshmorel:
> Is the ability to edit a post only available from the User profile page? Because if I create a post from an org, then click on the post I made, I won't have the option to edit it.

Editing organization posts is also an open PR #922

joshmorel:
> Just wanna know, if there is any reason, why the type comparison is preferred to the strict equality operator in lines 34, in the `backend/lib/endpoints/organizations.js` file. aside from that, every other thing seems fine.

The `==` was already there so at the time I decided not to change it. But yes strict equality would be preferable here because mongoose will always return `null` if not found. Fixed.

Issue #910, opened 2020-06-30
Fix styles in the onboarding flow <> other places in the platform. 


idjevm:
I believe that PR #813 messed up some stuff in the frontend. 

Please refer to this slack post -> https://fightpandemics.slack.com/archives/C010J3BMS9H/p1593525056033700

idjevm:
also #912 @adisen let's just address them here.

omarzoghayyer:
Also the submit button on >>Needs help page >> Step 3

idjevm:
Yeah the onboarding is the Needs help/ Offer Help pages, but this ticket addresses all the buttons across the platform. Most of them got messed up, so working with @adisen to find a solution. @adisen make sure you check all buttons, filters, creating a post, mobile/desktop, onboarding page, creating profile/organization. I noticed that a lot of them got messed up.

adisen:
Working on it. Gonna push the code today.

idjevm:
So we will close this one since I have reverted #813.

PR #937, opened 2020-07-01
Feature/fix button styles

adisen:
Fix styling issues with the Buttons.

Submit Button now from the antd button which is an HTML button compared to BaseButton which is an HTML link from the antd-mobile button. This allows for SubmitButton to submit a form on enter keypress
Issue #910


adisen:
@jyclam I don't get the second question.

jyclam:
@adisen I see that a lot of code is duplicated between the BaseButton and your Submit Button. It'll be easier to maintain if we avoid duplicating code like this. I think we can avoid this by adding an event handler for the enter key for the existing element, or, refactor the duplicate code into a new component which BaseButton and Submit Button both use and further add styles to.

https://github.com/FightPandemics/FightPandemics/blob/b5ff22fc50ab2bf77dd5c9c85fb0ce97022fd48f/client/src/components/Button/BaseButton.js

https://github.com/FightPandemics/FightPandemics/blob/85a18e345add2c00de4cc8cff5335c6b1c1fb1fe/client/src/components/Button/SubmitButton.js

joshmorel:
> @adisen I see that a lot of code is duplicated between the BaseButton and your Submit Button. It'll be easier to maintain if we avoid duplicating code like this. I think we can avoid this by adding an event handler for the enter key for the existing element, or, refactor the duplicate code into a new component which BaseButton and Submit Button both use and further add styles to.

I like @jyclam 's idea @adisen . Maybe a reset to where everything was working prior to this PR that fixes it is a good idea - we can just add the event listener to the component. I know there's been a lot of work here but sometimes we need to go back before we can go forward.

joshmorel:
> There are a few things here and there but this is almost ready to go! Please take a look and give it a test, all buttons need to look as expected. Please @adisen fix the way that filters look on mobile.

@idjevm Is it the lack of spacing or is there something else for this section?

@jyclam @adisen I guess we'll have to do something soon because we want to make the decision to soft launch or not and the current state of master is definitely not good. 

I will add @robinv85 here for opinion

adisen:
Sorry. 
Well I think if there's a way we can get rid of the duplicated code. It'll be better. 
We'll have a "form submit button" and a "Normal button which is a link"
I think adding an event listener on every page where we have a form to submit might lead to more bloated code.

idjevm:
Hey @joshmorel, so the filters on mobile should look like this -> 
image

But currently, they look like this -> 
image

This happened on the changes that were merged. 

Yeah I think we should revert back that PR that was merged. I will be taking a look today and give you guys an update, the thing is that it is not easily revertable since lots of changes have happened on master since that PR was merged but yeah I will work with Israel to resolve this issue quickly.

joshmorel:
> Hey @joshmorel, so the filters on mobile should look like this ->

Ah okay but these issue does seem indeed fixed in this PR. Maybe by more recent commits.

The only one I noticed is this issue - no spacing between buttons:

image

Where as this was prior (looking at prod before we merged the issue):
image

Anyways, I'm starting to change my mind :sweat_smile:  going back might not be the best choice. This seems close and the repeated code isn't a huge issue at the moment.

Whoever added `antd-mobile` into our app - that was a mistake. It is meant for react native projects and only support iOS and Android environments - not browsers! https://mobile.ant.design/docs/react/introduce#Environment-Support

I'm going to create a new (non-urgent) ticket to migrate everything to `antd`. @adisen - I think this one is close maybe? Just look into the minor spacing issue identified.

joshmorel:
@adisen Okay so we merged a reversion so at least it is working as before so other testing can continue more easily #1025 .

I don't think the event listener is a good approach ideally we'd use `antd` button but so the styling is consistent with the design and works as expected across the app. Minimize code duplication BUT if you can't get away from that's okay. Putting styling 

There is no rush now - the no enter to submit is not a show stopper.

idjevm:
Moved this to Draft as it is a WIP.

adisen:
Have we come to a conclusion on how to move ahead with this?

mannykary:
Based on the comments in #910 (which this PR is attempting to resolve), it appears that issue was closed by reverting a PR that introduced the regressions. So I am assuming that this PR is not MVP critical for soft launch, which is why I have updated the base branch to `rc/v1.1` so that we can freeze the `master` branch. I have also remove the MVP Critical tag from this PR.

adisen:
True Manny

idjevm:
Yup @mannykary that's right; thank you for updating this PR!

adisen:
@idjevm is there any other thing that needs to be worked on?

adisen:
This has been here for a while. @idjevm have you had time to check as discussed. tag: @joshmorel

joshmorel:
I will close this I think we solved in a different way? In any case it would be best to start again with a new branch from staging as this is so old.

PR #955, opened 2020-07-01
Every post, regardless of length, has a line

### LINKED ISSUE #950: Line is missing on posts that is less than 120 char
env: [https://staging.fightpandemics.work/feed](url) and also on profile (indv and org) page 

Posts that have texts that are less than 120 char is missing the line separating the text and the post engagement icons:
Screen Shot 2020-07-01 at 3 45 15 PM

Expected:
We should always see the line in every post.

BMariscal:
Closes #950


idjevm:
🔥

PR #958, opened 2020-07-02
On-hover, post page buttons match styleguide colors

### LINKED ISSUE #957: Fix the on-hover colors of edit and delete post buttons to match our styleguide
"The Hover color of buttons is a blue that isn't in our styleguide"

BMariscal:
Closes #957

### DIFF
--- a/client/src/components/Feed/StyledCommentButton.js
+++ b/client/src/components/Feed/StyledCommentButton.js
@@ -1,0 +1,26 @@
+import { theme } from "constants/theme";
+import { Button } from "antd";
+import styled from "styled-components";
+
+
+const { darkGray, lightGray, royalBlue, white } = theme.colors;
+
+export const StyledCommentButton = styled(Button)`
+  color: ${darkGray};
+  border-color: transparent;
+  border-radius: 1rem;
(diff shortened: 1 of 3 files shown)


client/src/components/Feed/StyledCommentButton.js
```diff
@@ -0,0 +1,25 @@
...
+  color: ${darkGray};
+  border-color: transparent;
+  border-radius: 1rem;
+  :hover{
+    color: ${royalBlue};
+    border: 1px solid ${white};
+  }
+`;
+
+export const StyledLoadMoreButton = styled(Button)`
```
silenceway, line 17: You can extend from previous StyledCommentButton and only add the border-color rule.

BMariscal: I thought about doing this but I figured it'd be best to keep them loosely coupled since they're buttons in different components and different areas of the page. It's an easy fix regardless and I can change it if you feel strongly about it. Thanks for taking a look!

silenceway: Let's go the current way.

PR #961, opened 2020-07-02
Fix organization initials on profilePic

omarzoghayyer:
Screen Shot 2020-07-01 at 8 25 49 PM

This PR should fix the organizations Initials (still working on it)

### DIFF
--- a/client/src/pages/OrganizationProfile.js
+++ b/client/src/pages/OrganizationProfile.js
@@ -195,7 +195,7 @@ const OrganizationProfile = () => {
       const nameArr = name.split(" ");
       if (nameArr.length < 2) {
         firstName = nameArr[0];
-        lastName = firstName.split("")[0]
+        lastName = firstName.split("")[1]
       } else {
         firstName = nameArr[0];
         lastName = nameArr[1];
--- a/client/src/pages/OrganizationProfile.js
+++ b/client/src/pages/OrganizationProfile.js
@@ -42,7 +42,7 @@ import {
   DrawerHeader,
   CustomDrawer,
 } from "../components/Profile/ProfileComponents";
-import { getInitials } from "utils/userInfo";
+import { getOrgInitials } from "utils/userInfo";
 import {
   LINKEDIN_URL,
   TWITTER_URL,
@@ -193,12 +193,12 @@ const OrganizationProfile = () => {
       const offerHelp = Object.values(objectives).some((val) => val === true);
       const { address } = location;
       const nameArr = name.split(" ");
-      if (nameArr.length < 2) {
+      if (nameArr.length < 1) {
         firstName = nameArr[0];
         lastName = firstName.split("").pop();
       } else {
         firstName = nameArr[0];
-        lastName = nameArr[1];
+        lastName = nameArr[0];
       }
 
       return (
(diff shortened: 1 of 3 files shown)


client/src/pages/OrganizationProfile.js
```diff
@@ -193,12 +193,12 @@ const OrganizationProfile = () => {
       const offerHelp = Object.values(objectives).some((val) => val === true);
       const { address } = location;
       const nameArr = name.split(" ");
-      if (nameArr.length < 2) {
+      if (nameArr.length < 1) {
```
silenceway, line 196: Is this working?
if `nameArr = [ "Firstname" ];` <- Length 1 false
It's only true with `nameArr = [];` and it's useless.

omarzoghayyer: no its not working! I was trying to use similar method to display 2 initials instead of 1

client/src/pages/OrganizationProfile.js
```diff
@@ -193,12 +193,12 @@ const OrganizationProfile = () => {
...
         firstName = nameArr[0];
         lastName = firstName.split("").pop();
       } else {
         firstName = nameArr[0];
-        lastName = nameArr[1];
+        lastName = nameArr[0];
```
idjevm, line 201: I mean if you have an array with the whole name, i.e `["Omar", "Zoghayyer"]` then this will give out initials such as `OO` because you are setting the last name to the first element of the array, assuming that it is an array with the full name broken down in parts.

client/src/utils/userInfo.js
```diff
@@ -1,8 +1,10 @@
...
   return firstName.charAt(0).toUpperCase() + lastName.charAt(0).toUpperCase();
 };
-
 export const getInitialsFromFullName = (fullName) => {
   const [firstName, lastName] = fullName.split(" ");
   return firstName.charAt(0).toUpperCase() + lastName.charAt(0).toUpperCase();
 };
+export const getOrgInitials = (firstName) => {
+  return firstName.charAt(0).toUpperCase()
+};
```
idjevm, line 10: You probably want to do something similar to lines 4-6. Where the full org name is passed into the function returns only the first of each word that was separated by space. Here you can also only return two initials, of the very first two words on the org name, so sort of split the array, and return the first char of the first word and the first char of the second word, ignore everything else for now.

client/src/pages/OrganizationProfile.js
idjevm: and based of my comment below, you don't have to change anything in this line here.

client/src/utils/userInfo.js
idjevm, line 10:
```suggestion
export const getOrgInitials = (fullOrgName) => {
  const orgNameSplit = fullOrgName.split(" ");
  return orgNameSplit[0].charAt(0).toUpperCase() + (orgNameSplit[1]) ? orgNameSplit[1].charAt(0).toUpperCase() : ""; 
};
```

idjevm: Or some logic like this one

silenceway: That change can be integrated in `getInitialsFromFullName` function and it doesn't require a new one.

idjevm: That's truee! I think we should make the body of that funtion more generic, so not first last name and first name. What do you think @silenceway ?

idjevm:
Hey @silenceway after talking with @omarzoghayyer, decided to close this one as it got a bit too messy and also omar asked for some help so I opened this PR #1005 that has the solution already but I am open to any suggestions. Let's keep the convo going on the other PR

Issue #857, opened 2020-06-27
Create a send button for posting comments


ppagarawal:
Please create a send button for posting comments

![image](https://user-images.githubusercontent.com/6276256/85916153-81fbdd80-b86b-11ea-9069-bd0f5a03eeb8.png)

- [ ] The user is able to create a comment via keyboard and button, both
- [ ] The send button is placed on right end of the text box

idjevm:
@BMariscal Are you covering this on your PRs too? I could put my hands on this one but I know a lot of things are going on your end on the posts so probably leave this to you?

PR #963, opened 2020-07-02
Feature/ Send button for comments in post view

nr431207:
Adding send button for comments in post 
Screen Shot 2020-07-01 at 8 27 05 PM

#857 


client/src/components/Input/AutoSize.js
```diff
@@ -1,15 +1,42 @@
...
 const AutoSize = ({ placeholder, value, onChange, onPressEnter }) => {
   return (
-    <TextArea
-      autoSize
-      value={value}
-      onChange={onChange}
-      onPressEnter={onPressEnter}
-      placeholder={placeholder}
-    />
+    <div>
```
idjevm, line 28: Do we need this`<div>` here?

nr431207: we can only return one element in the return section, since I have two elements I needed to wrap them in a div to count as one

BMariscal: For these cases, you can use fragments instead of <div></div>: https://reactjs.org/docs/fragments.html  short syntax: `<></>`

So insteand of `<div></div>` use `<></>`

nr431207: thats right :) I'll change is!

PR #964, opened 2020-07-02
Fixed comments age reactivity issue

### LINKED ISSUE #938: Fix comments age reactivity issue
The time ago only updates as you write a new comment. Best to make it non-reactive (like YouTube) so basically only updates on page refresh. 

But probably won't affect most people at first.

Or we can send from b/e.

comment-time-ago-reactive

### LINKED ISSUE #959: Display time in comments with spacing
The time within the comments is currently display without spacing (i.e. "10days ago" instead of "10 days ago")

BMariscal:
Previously, the time ago comment timestamps only updated as you wrote a new comment. 
Now, the timestamps are only updated on page refresh.

Changes:
* moved timeAgo function to the backend and created a Comment post-save hook to set the timeElapsed property to a relative time string generated from the createdAt ISO Date object. 
* Moment relative time timestamps now only update on page refresh. 

Closes #938
Closes #959

### DIFF
--- a/backend/lib/models/Comment.js
+++ b/backend/lib/models/Comment.js
@@ -16,7 +16,6 @@ const commentSchema = new Schema(
       ref: "User",
       type: [ObjectId],
     },
-    timeElapsed : String,
     parentId: {
       ref: "Comment",
       type: ObjectId,
@@ -49,11 +48,12 @@ commentSchema.index({ "author.id": 1, createdAt: -1 });
 commentSchema.index({ likes: 1 });
 /* eslint-enable */
 
-// post-save hook to set timeElapsed property with relative time string generated from createdAt.
-commentSchema.post('save', function(doc) {
-  this.set({ timeElapsed: translateISOtoRelativeTime(doc.createdAt) });
-  this.save();
-});
+commentSchema.set('toObject', { virtuals: true })
+commentSchema.set('toJSON', { virtuals: true })
+
+//set virtual 'timeElapsed' property on Comment POST and Comment PATCH.
+commentSchema.virtual("timeElapsed")
+  .get(function () { return translateISOtoRelativeTime(this.createdAt) });
 
 const Comment = model("Comment", commentSchema);
 
--- a/backend/lib/models/Comment.js
+++ b/backend/lib/models/Comment.js
@@ -1,5 +1,6 @@
 const { Schema, model, ObjectId } = require("mongoose");
 const { schema: authorSchema } = require("./Author");
+const { translateISOtoRelativeTime } = require("../utils");
 
 const commentSchema = new Schema(
   {
@@ -15,6 +16,7 @@ const commentSchema = new Schema(
       ref: "User",
       type: [ObjectId],
     },
+    timeElapsed : String,
     parentId: {
       ref: "Comment",
       type: ObjectId,
@@ -46,6 +48,12 @@ commentSchema.index({ "author.id": 1, createdAt: -1 });
 // Index for like's foreign key for lookup performance
 commentSchema.index({ likes: 1 });
 /* eslint-enable */
+
+// post-save hook to set timeElapsed property with relative time string generated from createdAt.
+commentSchema.post('save', function(doc) {
+  this.set({ timeElapsed: translateISOtoRelativeTime(doc.createdAt) });
+  this.save();
+});
 
 const Comment = model("Comment", commentSchema);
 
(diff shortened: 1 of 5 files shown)


BMariscal:
> LGTM! I gave it a test and seems to be working fine, let's wait for the others to see if they have any suggestions. I did notice that we added a new package though `moment.js`, shouldn't we update the `package.json` and `package-lock.json`?

Hey JV. Momentjs was already installed; it is being used in the `/posts` endpoint for other stuff I just imported it into the utils file.

idjevm:
> > LGTM! I gave it a test and seems to be working fine, let's wait for the others to see if they have any suggestions. I did notice that we added a new package though `moment.js`, shouldn't we update the `package.json` and `package-lock.json`?
> 
> Hey JV. Momentjs was already installed; it is being used in the `/posts` endpoint for other stuff. I just imported it into the utils file.

Aah that makes sense xD Thank you thank youuu! Awesome work @BMariscal

backend/lib/models/Comment.js
```diff
@@ -15,6 +16,7 @@ const commentSchema = new Schema(
       ref: "User",
       type: [ObjectId],
     },
+    timeElapsed : String,
```
joshmorel, line 19:
```suggestion
    timeElapsed : String,
```

I don't think it makes sense to store this in the DB.

We can add a virtual to the model with similar logic to post save hook below. FYI we want to return with find & update as well (virtuals are not automatically always returned by default) - see https://stackoverflow.com/a/26909782/4582054.

BMariscal: a-ha! thanks! I was looking for a way to dynamically set the computed property without storing it in the DB. Thanks for that info!

BMariscal: I had to use  `forEach` https://docs.mongodb.com/manual/reference/method/cursor.forEach/ to set the `timeElapsed` property on GET because you cannot query against a virtual property https://stackoverflow.com/questions/27536383/querying-on-a-virtual-property-in-mongoose

BMariscal:
> This is good the main thing that's not working is when I update a comment it doesn't update timeElapsed on the F/E.
> 
> But maybe we create a new issue for post soft launch. I think (but we need to confirm with PMs) that we should also include " (edited)" suffix in such a case.

I like "edited" tag!
Ah, so I don't think the timeElapsed property should be replaced with the updatedAt value on PATCH. I like how Reddit does it. They show both the time it was created and when it was edited, e.g., `186 points
 · 5 hours ago · edited 3 hours ago` 

But that's not my decision to make, I'll reach out to a PM. Thanks a bunch @joshmorel!

Naraujo13:
> > This is good the main thing that's not working is when I update a comment it doesn't update timeElapsed on the F/E.
> > But maybe we create a new issue for post soft launch. I think (but we need to confirm with PMs) that we should also include " (edited)" suffix in such a case.
> 
> I like "edited" tag!
> Ah, so I don't think the timeElapsed property should be replaced with the updatedAt value on PATCH. I like how Reddit does it. They show both the time it was created and when it was edited, e.g., `186 points · 5 hours ago · edited 3 hours ago`
> 
> But that's not my decision to make, I'll reach out to a PM. Thanks a bunch @joshmorel!

A bit late here, but nice solution @BMariscal! 
I also like reddit format, for that we could have an elapsedt time relative to `updateAt` and the edit can be created by comparing update and creation timestamps. We can add that in the aggregation projection, so it would return `edited` as a boolean value

BMariscal:
> > > This is good the main thing that's not working is when I update a comment it doesn't update timeElapsed on the F/E.
> > > But maybe we create a new issue for post soft launch. I think (but we need to confirm with PMs) that we should also include " (edited)" suffix in such a case.
> > 
> > 
> > I like "edited" tag!
> > Ah, so I don't think the timeElapsed property should be replaced with the updatedAt value on PATCH. I like how Reddit does it. They show both the time it was created and when it was edited, e.g., `186 points · 5 hours ago · edited 3 hours ago`
> > But that's not my decision to make, I'll reach out to a PM. Thanks a bunch @joshmorel!
> 
> A bit late here, but nice solution @BMariscal!
> I also like reddit format, for that we could have an elapsedt time relative to `updateAt` and the edit can be created by comparing update and creation timestamps. We can add that in the aggregation projection, so it would return `edited` as a boolean value

Thanks for your input @Naraujo13! I'll make sure to add that info to the ticket

Issue #908, opened 2020-06-30
Fix home page sizing to fit different screen


ppagarawal:
As a User, I want to have consistent UI experience across devices.

Currently, the components of the homepage are not proportionate to the different screen sizes. The frontend needs to be proportionate across screens. This change has to be implemented across all the different features.

Eg: 
![image (4)](https://user-images.githubusercontent.com/6276256/86091054-db147d00-bac8-11ea-8ae3-c3ace1cda991.png)

@Yu-Alexa-Zhao Can you please share the proportion of the component in the screen.

pthoma14:
@ppagarawal this is a great issue! Thanks for logging this. 

Maybe for these UI type issues we don't use the "MVP-Critical" tag but rather "MVP-Release 1" and "Good First Issue" tags

Thoughts?

Yu-Alexa-Zhao:
Is it possible to - 
1) For both the left and right sections to take up the entire height? 
2) Ideally the two boxes would take up to 75% of the height of the div ...

Not sure if I am being descriptive here...
![image](https://user-images.githubusercontent.com/44085699/86192057-a29b9000-bb16-11ea-8b90-f8cb6c038a56.png)

Joeu:
@Yu-Alexa-Zhao , can you check how it is now in https://staging.fightpandemics.work/ and give you thoughts about it? I just pushed a fix for other issue that deals with this blue part on the home page.
I'd also like to know which screen sizes we are going to support, so we can change the proportions using the font-size approach.

Yu-Alexa-Zhao:
Yes I think the design now makes perfect sense to me

idjevm:
I am not sure about supporting 4k resolutions, I mean I guess we should consider having a bit screen resolution but most probably will ok without 4k res.

Joeu:
I made a lot of changes to deal with sizing and proportion. Now we are able to set the proportion by adjusting the font-size in the GlobalStyles.js file.

@idjevm , @jyclam , @manualzuru , @Yu-Alexa-Zhao , @ppagarawal , please check and make comments regarding the screenshots below (representing 4k screens):

![4kScreenProportions1](https://user-images.githubusercontent.com/1004896/86392135-57e25b00-bc71-11ea-8ee8-0dc49151a703.png)
![4kScreenProportions2](https://user-images.githubusercontent.com/1004896/86392149-5ca70f00-bc71-11ea-9517-40b00fa97a03.png)
[...]

PR #965, opened 2020-07-02
Feature/908 supporting different screen sizes

Joeu:

Unit changing from px to rem units, allowing us to change elements proportions by just changing the value of the font-size in the GlobalStyles file.

_Please be concise and link any related issue(s):_ #908 

### DIFF
--- a/client/src/components/Feed/PostCard.js
+++ b/client/src/components/Feed/PostCard.js
@@ -89,6 +89,7 @@ const PostCard = styled(Card)`
         margin-top: -1.8rem;
 
         .status-icon {
+          width: 1rem;
           margin-right: 0.8rem;
         }
       }
--- a/client/src/components/Feed/PostCard.js
+++ b/client/src/components/Feed/PostCard.js
@@ -162,6 +162,7 @@ const PostCard = styled(Card)`
         }
 
         .total-number {
+          font-size: ${medium};
           display: inline;
 
           @media screen and (min-width: ${mq.tablet.narrow.minWidth}) {
--- a/client/src/components/Feed/PostCard.js
+++ b/client/src/components/Feed/PostCard.js
@@ -89,7 +89,7 @@ const PostCard = styled(Card)`
         margin-top: -1.8rem;
 
         .status-icon {
-          width: 1rem;
+          width: fit-content;
           margin-right: 0.8rem;
         }
       }
--- a/client/src/GlobalStyles.js
+++ b/client/src/GlobalStyles.js
@@ -62,7 +62,7 @@ a {
 }
 
 .app-drawer .am-drawer-sidebar .am-list {
-  width: 200px;
+  width: 20rem;
   padding: 0;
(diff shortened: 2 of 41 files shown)


idjevm:
Let's open this up, besides the changes that @BMariscal pointed out, there just might be little things that we will need to fix here and there but Loooks already WAAAY better of how it is right now

Joeu:
Requested changes applied, @idjevm  and @BMariscal .

client/src/components/Feed/PostCard.js
```diff
@@ -88,6 +88,7 @@ const PostCard = styled(Card)`
         margin-top: -1.8rem;
 
         .status-icon {
+          width: 1rem;
```
BMariscal, line 91: The size of the blue button next to the country name is too big atm but this should fix it: 
```suggestion
          width: fit-content;
```

BMariscal:
When you click on the "Create Post" button, the icons inside the buttons need to be resized and realigned.

BMariscal:
If you go the user's profile page, the "Create Post" button is still too small:

BMariscal:
Drop down arrows within the visibility buttons, in this case 'City' and 'Month' need to be fixed.

Joeu:
Just pushed changes related to the modals, I'm still counting on you to help me cover the whole layout, @idjevm and @BMariscal .  As we have no specs for higher resolutions, we will keep finding places where changes will be needed.
The main goal of my changes is to change proportions by changing the font-size in GlobalStyles.

BMariscal:
Thanks for all the work you've done here @Joeu 
My screen resolution is just "1440 x 900" 

The icons inside the Create Post modal are too small:


But this is how you can fix them: 

add a class, in this case I called it `.icon`, to the images and unset the height and width.

BMariscal:
Also this still too small
 

Please rebase and fix the merge conflicts. Thanks for all your work!

BMariscal:
> We are almost there, just a few more changes though:
> image
> 
> image
> 
> image
> 
> image
> 
> Org Profile:
> image
> 
> I think it needs a bit more work, in org profile, post tags across orgs / profile / feed . And also the same buttons and arrows a previously mentioned by Bri, let me know I can slack you and show you more xD

hmm, the sign in button and the tags look good to me, even on 2560 x 1440 resolution. What resolution are you testing these with? 

The org profile page, not the user profile, still needs work. Why does the same button shared across the app not have the same styling? These should be set in one place, GlobalStyles, and inherited everywhere else.

Joeu:
I agree with you about the button, JV, we need to dig more into that to see why we have WizardSubmit and SubmitButton.

Joeu:
Check again the tags on the Feed, please.

Joeu:
My point is that we should push the code and fix those small issues later, the main goal was to change the units from pixels to rem. But I agree with you and I can say that more tweaks like that will appear as we go through application flow.

PR #966, opened 2020-07-02
Clear Modal state on Modal close

### LINKED ISSUE #876: Removing the state of the modal after a successful post
So after successfuly creating a post, the state of the create post button is not being reset, 
Ideally we would want to be able to create a new post instead of having the message from the last post. 

Refer to the gif here: 
https://share.getcloudapp.com/o0uvZeWE

jyclam:
User can hit 'Create a post' button repeatedly and create new posts. Currently, the success content is shown after first post and remains even after closing and reopening.

closes #876

### DIFF
--- a/client/src/components/CreatePost/CreatePost.js
+++ b/client/src/components/CreatePost/CreatePost.js
@@ -144,13 +144,13 @@ const Step4 = () => {
   );
 };
 
-const CreatePost = ({ toggleModal, ...props }) => {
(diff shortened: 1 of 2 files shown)


jyclam:
Hang on, I might have broken something during final clean up. Let me review first.

@idjevm can you let me know where we access those and I'll take a look?

jyclam:
@idjevm 
> Can we add this also to the org profile create post and individiual profile create post ?
Done, they use the same component

PR #974, opened 2020-07-02
Add icon to manifest.json

kylejvh:
Aims to fix issue #442.

Adds a new svg icon in the manifest.json for use where appropriate. After looking into it, this is the icon used in Chrome when a user adds a shortcut to their homescreen. 

fp logo mobile


kylejvh:
I added the svg icon I used to the public folder. I would have preferred to serve it from the assets folder with the other icons, but I couldn't get the src path worked out correctly. If anyone has any opinions on this or could tell me how to fix it, let me know. If not, it's working as is.

kylejvh:
Also not sure why it's showing a warning for the way the sizes are provided:

icon size errors


MDN shows this as valid in their examples:
https://developer.mozilla.org/en-US/docs/Web/Manifest/icons

joshmorel:
> Also not sure why it's showing a warning for the way the sizes are provided:
> 
> icon size errors
> 
> MDN shows this as valid in their examples:
> https://developer.mozilla.org/en-US/docs/Web/Manifest/icons

It seems to want only one size that matches the actual dimensions of the icon.

But this does seem valid (albeit this is not an area I'm too experienced in) :man_shrugging:

joshmorel:
Despite the warnings it works totally fine on my Android.

Issue #973, opened 2020-07-02
Update texts


doheekim93:
The following pages have texts that require changes
1. Nav Bar (for users _not_ signed in)
2. Nav Bar (for users signed in)
3. Home Page
4. Sign In Page
5. Join Now Page
6. Help Board Page

### Changes needed for each page:
**1. Nav Bar (for users _not_ signed in)**

Desktop:
Feed --> Help Board
Login --> Sign In
Register --> Join Now
<kbd>
    <img width="1440" alt="Screen Shot 2020-07-02 at 4 18 06 PM" src="https://user-images.githubusercontent.com/66695502/86412163-0394a680-bc84-11ea-8350-03d05461d782.png">
</kbd>

Mobile:
Login --> Sign In
Register --> Join Now
<kbd>
    <img width="200" src="https://user-images.githubusercontent.com/66695502/86413010-cb8e6300-bc85-11ea-8240-284ba6a9230f.jpg">
</kbd>

\
**2. Nav Bar (for users _signed in_)**

Desktop:
Feed --> Help Board
<kbd>
    <img width="1440" alt="Screen Shot 2020-07-02 at 5 09 28 PM" src="https://user-images.githubusercontent.com/66695502/86413468-e6ada280-bc86-11ea-9901-dfa47928dc79.png">
</kbd>

Logout --> Sign Out
<kbd>
    <img width="321" alt="Screen Shot 2020-07-02 at 4 21 59 PM" src="https://user-images.githubusercontent.com/66695502/86413842-e661d700-bc87-11ea-9fdb-abb6ab3bc154.png">
</kbd>

Mobile:
Feed --> Help Board
Logout --> Sign Out
<kbd>
    <img width="200" src="https://user-images.githubusercontent.com/66695502/86413541-1197f680-bc87-11ea-895e-174a4ad60d32.jpg">
</kbd>

\
**3. Home Page**

Desktop:
View Community Postings --> View Help Board
<kbd>
    <img width="700" alt="Screen Shot 2020-07-02 at 4 19 51 PM" src="https://user-images.githubusercontent.com/66695502/86414081-959eae00-bc88-11ea-80e2-c77c8cbc3d93.png">
</kbd>

Mobile:
View Community Postings --> View Help Board
<kbd>
    <img width="200" src="https://user-images.githubusercontent.com/66695502/86414241-fd54f900-bc88-11ea-82a8-f15f7debcb99.jpg">
</kbd>

\
**4. Sign In Page**

[...]

PR #975, opened 2020-07-02
update copy - issue #973

stavares843:

https://github.com/FightPandemics/FightPandemics/issues/973

### DIFF
--- a/client/src/pages/Faq.js
+++ b/client/src/pages/Faq.js
@@ -94,7 +94,7 @@ const Faq = () => {
             </p>
             <p>
               To join now as a user click the menu button and select the
-              "Join Now/Sign In" option. There you will be able to join now with
+              "Join Now/Sign In" option. There you will be able to join with
               your email or social media credentials.
             </p>
             <p>
@@ -368,7 +368,7 @@ const Faq = () => {
             <summary>How do I join now?</summary>
             <p>
               To join the community, click the "Menu" button and
-              select "Join Now/Sign In". There you will be able to join now with
+              select "Join Now/Sign In". There you will be able to join with
               your choice of email or social media credentials.
             </p>
             <p>
--- a/client/src/pages/Faq.js
+++ b/client/src/pages/Faq.js
@@ -93,8 +93,8 @@ const Faq = () => {
               posts and comments on FightPandemics platform.
             </p>
             <p>
-              To sign up as a user click the menu button and select the
-              "Register/Log in" option. There you will be able to sign up with
+              To join now as a user click the menu button and select the
+              "Join Now/Sign In" option. There you will be able to join now with
               your email or social media credentials.
             </p>
             <p>
@@ -197,7 +197,7 @@ const Faq = () => {
             <p>
               From the homepage, select "Give help" to see all the posts created
               by users asking for help and filter the results by location and
-              type. You need to be logged in to reach out to them. You can also
+              type. You need to be signed in to reach out to them. You can also
               create your own postings describing the help you're able to offer.
             </p>
           </details>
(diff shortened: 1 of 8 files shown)


stavares843:
closed PR because product team is discussing about it

idjevm:
@stavares843 Pleaseee use more descriptive names for commit messages & PR names 🙏 We already have like many many update copy PRs merged and commits merged, not a good thing to do if we do need to go back or figure where something was changed.

stavares843:
reopened because product team decided to include these modifications

stavares843:
https://github.com/FightPandemics/FightPandemics/issues/973 doesn't include faq

jyclam:
You're right, but it should probably be changed along with this. Otherwise, we'd be pointing people at buttons that no longer exist.

stavares843:
updated

client/src/pages/Faq.js
```diff
@@ -411,7 +411,7 @@ const Faq = () => {
           <details>
             <summary>How can I reset my password?</summary>
             <p>
-              Click on "Login" from the homepage and then click on "Forgot
+              Click on "Sign In" from the homepage and then click on "Forgot
               password" and follow the instructions
```
idjevm, line 415: Can you change line 389 and 424 to say something like `Your sign in credentials will be the same as` instead of your login credetials will be the same as..

client/src/pages/Faq.js
```diff
@@ -422,7 +422,7 @@ const Faq = () => {
             <p>
               Please check if your password has at least 8 letters. If you
               already have an account but forgot your login password, click on
```
idjevm, line 424: Changing your sign in password on line 424

client/src/pages/Login.js
```diff
@@ -527,7 +527,7 @@ const Login = ({ isLoginForm, forgotPassword }) => {
             <WhiteSpace />
             {!forgotPassword && (
               <SectionDiv className="text-center">
```
idjevm, line 529: Above here, there is line 523, it says Back to Login page, change it to back to sign in page.

stavares843:
#973 doesn't include privacy policy

client/src/pages/Faq.js
stavares843: did you ask product about that copy?

client/src/pages/Faq.js
stavares843: updated

client/src/pages/Login.js
stavares843: updated

client/src/pages/Faq.js
doheekim93: Yes, we want to have "sign in" instead of "login" in all places!! Thanks @stavares843

doheekim93: Also, please replace all "Sign Up" to "Join Now"!!

stavares843: done

client/src/pages/Faq.js
```diff
@@ -89,12 +89,12 @@ const Faq = () => {
...
             <p>
-              As a registered user you have unlimited access to view and create
+              When you join the community, you will have unlimited access to view and create
               posts and comments on FightPandemics platform.
             </p>
             <p>
-              To sign up as a user click the menu button and select the
-              "Register/Log in" option. There you will be able to sign up with
+              To join now as a user click the menu button and select the
+              "Join Now/Sign In" option. There you will be able to join now with
```
idjevm, line 97: Maybe change here the wording to just say `... able to join with ...` instead of `... able to join now with ...` What do you think @doheekim93 ?

doheekim93: Yea, agree with @idjevm "able to join with" flows better!

stavares843: updated

stavares843:
thats unrelated to #973

idjevm:
This is almost good to go, just update the `Header.js` file and then I think it should be close to be good to go.

PR #1005, opened 2020-07-04
Stetting orgs and posts avatar icon to only 1 initial or 2 initials

### LINKED ISSUE #918: Profile "Logo" is different than which is displayed on the POSTs added by the respective Organizations
1. Open "https://staging.fightpandemics.work"
2. Login using valid email/Password
3. Navigate to Profile -> [Org_name_added before]
4. Create a POST

Issue# You wil observe that the profile logo displayed on Company profile is different than its display on the POST added by them.

Refer below GIF. 

https://share.getcloudapp.com/NQuglJ90

idjevm:
# What does this pull request aim to achieve? 💯
This PR just fixes the initials in posts and org profile depending on name. If name only 1 word, then the avatar will be 1 letter, but if the name is 2 words or more, then the name will be 2 letters
This PR resolves #918 

Here is the outcome of this PR: 
image

### DIFF
--- a/client/src/components/Feed/Post.js
+++ b/client/src/components/Feed/Post.js
@@ -33,6 +33,7 @@ import {
 } from "hooks/actions/postActions";
 import { isAuthorOrg } from "pages/Feed";
 import { authorProfileLink } from "./utils";
+import { getInitialsFromFullName } from "utils/userInfo";
 
 // Icons
 import SvgIcon from "../Icon/SvgIcon";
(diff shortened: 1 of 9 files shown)


idjevm:
Yeah, I was looking at one that we already have actually 🤔 It is on the User info utils. I might just implement it there so that way we can use it here and also on the org profile

rabbi08:
Validate #918 
Profile logo is displayed as first initials  of the first and second names.

PR #1006, opened 2020-07-04
add simple mock geolocation client for dev

joshmorel:

This has come up several times for onboarding so I want to get something super simple working now.

It's simple mock client of the `@googlemaps/google-maps-services-js` `Client` that returns a fixed result for the 3 different endpoints:

image

If we use multiple api services for service rotation later we'll change it but I wanted at least something so we're not handling all these questions and errors.

If anyone asks to the weird result while doing other dev we can point them to the readme. For people needing to work on this feature we can assist them in getting their own key.


joshmorel:
I guess I could at least return a different result depending on which they click so they can create users/filters with a handful of different locations. I'll finish this tomorrow.

joshmorel:
I re-thought the approach I quickly put together last night:

Instead of mocking the gmaps library client I'm mocking our own `Geo` component that uses the gmaps client. This is less code but more importantly is more independent of gmaps if we decide to do service rotation or another external service altogether. The result of one method - `getAddressPredictions` - is still coupled to the structure of the gmaps result, but this can be updated later when we choose to do service-rotation (interfaces might be a good idea here).

If there is no key on dev environment they will see only five results no matter what is typed:

image

Picking one will actually return the related result (my 1st version returned one no matter what is picked) while reverse-geolocation (from share my location) will always return the same result.

You can test the current state (gmaps) is still working via the [review branch](https://featuregeolocation-dev-mock.fightpandemics.xyz). Testing the mock requires checking out this branch locally then deleting or commenting out `GOOGLE_MAPS_API_KEY=AIza...` in `backend/.env` before starting.

PR #1008, opened 2020-07-05
Link comment avatar and name to profile

### LINKED ISSUE #944: Clicking on profile avatar in Comments should redirect user to the profile page
Similar to how it's done in Post:

https://github.com/FightPandemics/FightPandemics/blob/1ae090dc0a094b979c1cc579aab194e45b9a6f56/client/src/components/Feed/Post.js#L81

Should probably move `authorProfileLink` as a utility function so it's the same for post Post & Comment

Currently, I think only individuals can comment but we should account for potential for orgs to comment later.

For any posted comments, as a logged in user, clicking on the avatar of the author of the comment should take me to their profile (individual profile if the comment is made by an individual and org profile is the comment is made by an organization).

jyclam:
* Refactor authorProfileLink to Feed/utils
* Links comment author name and avatar to profile

closes #944

### DIFF
--- a/client/src/components/Feed/Post.js
+++ b/client/src/components/Feed/Post.js
@@ -32,11 +32,11 @@ import {
   TOGGLE_COMMENTS,
 } from "hooks/actions/postActions";
 import { isAuthorOrg } from "pages/Feed";
+import { authorProfileLink } from "./utils";
 
 // Icons
 import SvgIcon from "../Icon/SvgIcon";
 import statusIndicator from "assets/icons/status-indicator.svg";
-const INDIVIDUAL_AUTHOR_TYPE = "Individual";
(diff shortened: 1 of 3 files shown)


client/src/components/Feed/Post.js
```diff
@@ -83,9 +83,6 @@ const Post = ({
       post.author.name.match(/\b\w/g).join("").toUpperCase()) ||
     "";
 
-  const authorProfileLink = `/${
-    post.author.type === INDIVIDUAL_AUTHOR_TYPE ? "profile" : "organization"
-  }/${post.author.id}`;
```
omarzoghayyer, line 88: Thank you for the clean up.

PR #1051, opened 2020-07-07
On comment PATCH, display an 'edited' tag and the relative update time timestamp in the comment

### LINKED ISSUE #1001: On comment PATCH, display an 'edited' tag and the relative update time timestamp in the comment
On comment edit, the comment should display something like
`5 hours ago · edited 3 hours ago` where
`${relative create time timestamp} · edited ${relative update time timestamp}`

Similar to Reddit's approach:

---
 
This will require some work on the backend as well as some frontend changes. 

---
Please see the comments here for some ideas on how to approach this implementation:
https://github.com/FightPandemics/FightPandemics/pull/964

BMariscal:
Changes:

* Added the property `edited` to the Comments aggregation on GET and set its value to false if  createdAt timestamp is **>=** the updatedAt timestamp. 

* Added a virtual property called `editTimeElapsed` that's a relative time stamp of the updatedAt timestamp. `editTimeElapsed` is only displayed in the frontend if the `edited` property is set to true. 

Closes #1001

### DIFF
--- a/client/src/components/Feed/StyledComment.js
+++ b/client/src/components/Feed/StyledComment.js
@@ -16,7 +16,7 @@ const StyledComment = styled(Comment)`
     .ant-comment-inner {
       padding: 1rem 0;
       .ant-comment-content {
-        width: 40rem;
+        width: 45rem;
         .ant-comment-content-author-time {
           cursor: default;
           padding-right: 25px;
--- a/client/src/components/Feed/StyledComment.js
+++ b/client/src/components/Feed/StyledComment.js
@@ -16,8 +16,7 @@ const StyledComment = styled(Comment)`
     .ant-comment-inner {
       padding: 1rem 0;
       .ant-comment-content {
-        min-width: 40rem;
-        max-width: 45rem;
+        width: 40rem;
         .ant-comment-content-author-time {
           cursor: default;
           padding-right: 25px;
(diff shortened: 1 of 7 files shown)


client/src/components/Feed/StyledComment.js
```diff
@@ -16,7 +16,7 @@ const StyledComment = styled(Comment)`
     .ant-comment-inner {
       padding: 1rem 0;
       .ant-comment-content {
-        max-width: 30rem;
+        width: 45rem;
```
joshmorel, line 19: This seems to be causing issues with narrower widths:

image

Maybe if post page area wasn't so narrow that would be okay but it doesn't solve it for mobile:

image


Can we break we let it break onto multiple lines if necessary?

BMariscal: ah thanks! great catch!

BMariscal:
> Thanks @BMariscal ! Besides the styling issue. I personally like what YouTube is doing here, all the logic is handled on the B/E:
> 
> image
> 
> So if we have one function `elapsedTimeText` that handles everything we wouldn't need to pass back all 3 and then handle variable logic on the F/E.
> 
> So I think that would be an improvement (we likely will need to repeat for posts at some point) although not critical.
> 
> FYI - it seems Reddit handles all the logic for this kind of thing the F/E.
> 
> image

hmm I think that's a bit cleaner. I'll give it a shot. Thanks @joshmorel

Issue #842, opened 2020-06-26
Mongoose findOneAndModify deprecation warning (low priority)


joshmorel:
We get this warning where we are using `findOneAndUpdate` 
```
(node:29) DeprecationWarning: Mongoose: `findOneAndUpdate()` and `findOneAndDelete()` without the `useFindAndModify` option set to false are deprecated. See: https://mongoosejs.com/docs/deprecations.html#findandmodify
```

We should use `    useFindAndModify: false,` option

kaydeejay:
I'd like to fix this bug

kaydeejay:
I've added the `{ useFindAndModify: false }` option to the query in `backend/lib/endpoints/posts.js`, but now when I try to 'like' a post on the UI, the request seems to go through successfully but the like icon does not change. Oddly enough, if I load the page and the post is already liked, I can 'unlike' it, but not 'like' it again. [Demonstration](https://imgur.com/PskFmAM)

joshmorel:
Hey Kevin - this issue actually seems to be related to
________________________________

joshmorel:
Resolved by #1095

PR #1095, opened 2020-07-09
fixes mongo useFindAndModify deprecation error

kaydeejay:
Resolves issue #842 
- In backend/config.js, adds `useFindAndModify: false` to mongo params
- Fixes mongo/mongoose useFindAndModify deprecation warning


idjevm:
Yeah Thanks @joshmorel, yeah the prehook commit is the one that I talked with Kevin to get it resolved. cuz it won't let him commit if it is not in order

PR #1109, opened 2020-07-10
adding SendGrid API Key and SendGrid Contact List ID to main.tf

### LINKED ISSUE #1108: Add SendGrid config to terraform file
@mannykary has already added the secrets to the store, now we just need to modify the `main.tf` file to be able to pull those secrets from the aws store

idjevm:
resolves #1108


idjevm:
> LGTM - what is the plan to test that it's working in review/staging?
> 
> We need to also update `SENDGRID_CONTACTS_API_URL` to just be a constant instead of env variable.
> 
> So do we merge this, add the SSM parameters and then there will be a 2nd PR to address the above when you'll test or should all be part of the same PR?

So Many has added the SSM parameter already I believe. I will actually update this PR to update the URL as well! Thanks for reminding me haha 

As for the plan to test, yeah, I am not sure, what do you think we should do? I know in my local is working haaa, but not sure if it will be working in staging...

joshmorel:
> As for the plan to test, yeah, I am not sure, what do you think we should do? I know in my local is working haaa, but not sure if it will be working in staging...

With this one built it should work if @mannykary  added the keys for review (or was it only staging/prod or only prod?).

I tried with my email and I didn't see it populate in SendGrid. 

We ideally should have different lists for staging/review/prod or we'll get junk in our prod onboarding list due to QA. But having a staging/review one is useful if we make changes and want to confirm it is populating properly before deploying to prod.

PR #1100, opened 2020-07-09
Fix various GTM bugs

jyclam:
### What does this pull request aim to achieve? 💯
Fix various GTM bugs persisting after PR #1091

closes #1081

PR #1111, opened 2020-07-10
Fix bug for check boxes not being clickable introduced by #1091

idjevm:
There was a bug introduced in 1091, so @jyclam I will rever those two files, and that should help with solving this bug. 

Thanks @nadinejuraschek for pointing it out!!!


idjevm:
My bad @jyclam, not 1091 but #1100. This checkboxes are for the create profile and I think this should not affect what you did on #1091

jyclam:
@idjevm Do you have more details on how the checkbox was not clickable? It seems to work fine from my testing

PR #1112, opened 2020-07-10
mobile button size to add post in feed doubled

### LINKED ISSUE #1107: [Mobile] The "+" icon on feed is very small in mobile version.
The "+" icon on the feed is very tiny in the mobile version.

Create post icon small

nadinejuraschek:
closes #1107

- increased the size of the button on mobile (2.2rem to 4rem) so it is big enough to tap with a finger

### DIFF
--- a/client/src/components/Feed/FeedWrapper.js
+++ b/client/src/components/Feed/FeedWrapper.js
@@ -6,13 +6,6 @@ const FeedWrapper = styled.div`
   font-family: ${display};
   width: 100%;
   position: relative;
-
-  .create-post {
-    position: fixed;
-    bottom: 5%;
-    right: 5%;
-    cursor: pointer;
-  }
 `;
 
 export default FeedWrapper;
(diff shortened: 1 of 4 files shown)


joshmorel:
> Looks good! But ideally we would want to update these create post button across all the places (create post from org profile) and from mobile as well. Also delete the `package-lock.json` (helping you with this)

`Profile` already has `5rem` but `OrganisationProfile` does not so it's too small. So yeah I would update that or just add width/height rem to component.

```
              <CreatePostIcon
                id={GTM.user.profilePrefix + GTM.post.createPost}
                src={createPost}
                onClick={() => setModal(!modal)}
                style={{ width: "5rem", height: "5rem" }}
              />
```

The feed though - I tested it on an actual phone and it works fine.

galeza:
Tested on Android 10
Button size:
Screenshot_2020-07-13-13-29-00-716_com android chrome

galeza:
Hi 
Button size is ok on the Feed screen and Profile screen.
I agree that it should be corrected also on the Organisation -> create post. Screenshot below:
Screenshot_2020-07-13-13-35-13-735_com android chrome

Issue #1090, opened 2020-07-08
Can't like a post after 1st signing up


joshmorel:
@kaydeejay I think this issue is related to #923 

This endpoint doesn't work after signing up because the POST users/ returns `_id` as `_id` not `id` 

```
const endPoint = `/api/posts/${postId}/likes/${user && user.id}`;
```

I wish we had always sent back `_id` as `_id` instead of sometimes `id` but for now can you fix this one?

kaydeejay:
Sure thing, @joshmorel ! Happy to take a look at it.

idjevm:
@joshmorel so if I refresh it works, doesn't it have to do with state or something like that?

joshmorel:
@idjevm No it's because of the data sent from the backend. After creation the user object has _id where as get current user (called on refresh) it's id

Renaming _id to id in some routes was a mistake imho

idjevm:
aaaaht that makes sense. :O Damn hopefully we can get this done soon @kaydeejay, let us know if you have any questions

kaydeejay:
Will do @idjevm , I plan to spend a good of time on this tomorrow and Sunday morning (EST). I'll reach out to you or @joshmorel if I have any questions. Thanks!

joshmorel:
Resolved by #1136

PR #1136, opened 2020-07-13
fixes ability to like posts after initial signup and before refresh (…

kaydeejay:
resolves issue #1090 

changes:
`const endPoint = \`/api/posts/${postId}/likes/${user && user.id}\`;`
to:
`const endPoint = \`/api/posts/${postId}/likes/${user && user._id}\`;``
in the following files:
  - /client/src/pages/Profile.js
  - /client/src/pages/OrganisationProfile.js
  - /client/src/pages/PostPage.js
  - /client/src/pages/Profile.js

### DIFF
--- a/client/src/pages/Feed.js
+++ b/client/src/pages/Feed.js
@@ -327,7 +327,7 @@ const Feed = (props) => {
     sessionStorage.removeItem("likePost");
 
     if (isAuthenticated) {
-      const endPoint = `/api/posts/${postId}/likes/${user && user._id}`;
+      const endPoint = `/api/posts/${postId}/likes/${user?.id || user?._id}`;
       let response = {};
 
       if (user) {
--- a/client/src/pages/Feed.js
+++ b/client/src/pages/Feed.js
@@ -327,7 +327,7 @@ const Feed = (props) => {
     sessionStorage.removeItem("likePost");
 
     if (isAuthenticated) {
-      const endPoint = `/api/posts/${postId}/likes/${user && user.id}`;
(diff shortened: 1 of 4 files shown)


client/src/pages/Feed.js
```diff
@@ -327,7 +327,7 @@ const Feed = (props) => {
     sessionStorage.removeItem("likePost");
 
     if (isAuthenticated) {
-      const endPoint = `/api/posts/${postId}/likes/${user && user.id}`;
+      const endPoint = `/api/posts/${postId}/likes/${user && user._id}`;
```
joshmorel, line 330: Oh I thought you were going to change backend too from our conversation on slack. 

This will work after sign-up but now it doesn't work later on because of: https://github.com/FightPandemics/FightPandemics/blob/a09c8b314a8d818b1c1964a524564f423979ddd6/backend/lib/endpoints/users.js#L31. 

Let's take more time to do overall full stack update (to use `_id` only) with a resolution to #923 

For now to fix the bug do something similar to https://github.com/FightPandemics/FightPandemics/blob/a09c8b314a8d818b1c1964a524564f423979ddd6/client/src/components/Header.js#L123

```suggestion
      const endPoint = `/api/posts/${postId}/likes/${user?.id || user?._id}`;
```

kaydeejay: Will do! I'll fix the bug for now, but I too would like to fix this issue in a more permanent way.

joshmorel: Agreed - assign yourself to the other issue and start working on that one after this is merged.

PR #1160, opened 2020-07-14
Match rc/* and hotfix/* branches for review workflow

mannykary:

Match `rc/*` and `hotfix/*` branches for review workflow so that a review build can be spun up.

The build for `rc/v1.1` will be accessible at rcv11.fightpandemics.work.


idjevm:
🤔 thinking about this, perhaps this can be added to master as well?

PR #1208, opened 2020-07-17
initializing gtm in Login page with userId

### LINKED ISSUE #1190: Extra GTM <noscript> added when navigating between pages
**Steps to recreate:**
1. Sign in to your account 
2. Go to homepage on staging https://staging.fightpandemics.work/
3. Click on **any** links eg `About Us`, `Help Board`, `Request Help`, `Offer Help`...
4. Observe that an extra <noscript> tag is added 

It's a GTM related element
```
      <iframe src="https://www.googletagmanager.com/ns.html?id=undefined&gtm_auth=&gtm_preview=&gtm_cookies_win=x"
        height="0" width="0" style="display:none;visibility:hidden" id="tag-manager"></iframe>
```

Screen Shot 2020-07-16 at 12 52 21 PM

Looks like every time we call `TagManager.initialize` it inserts a new one, https://github.com/alinemorelli/react-gtm/blob/master/src/TagManager.js#L43

nr431207:
fixes duplicate `<noscript>` tag that appears when switching between pages.
fixes #1190 

### DIFF
--- a/client/src/App.js
+++ b/client/src/App.js
@@ -28,8 +28,6 @@ export const tagManagerArgs = {
   gtmId: process.env.REACT_APP_GTM_ID,
 };
 
-// TagManager.initialize(tagManagerArgs);
-
 const mapDispatchToProps = {
   initAuth: initAuth,
 };
--- a/client/src/App.js
+++ b/client/src/App.js
@@ -28,7 +28,7 @@ export const tagManagerArgs = {
   gtmId: process.env.REACT_APP_GTM_ID,
 };
 
-TagManager.initialize(tagManagerArgs);
+// TagManager.initialize(tagManagerArgs);
 
 const mapDispatchToProps = {
   initAuth: initAuth,
--- a/client/src/pages/Home.js
+++ b/client/src/pages/Home.js
@@ -6,8 +6,6 @@ import ImageButton from "components/Button/ImageButton";
 import { theme, mq } from "constants/theme";
 import TextLabel from "components/Typography/TextLabel";
 import GTM from "constants/gtm-tags";
-import tagManagerArgs from "App";
-import TagManager from "react-gtm-module";
 
 const { typography } = theme;
 const { black, royalBlue, white, offWhite } = theme.colors;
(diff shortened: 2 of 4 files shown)


nr431207:
@joshmorel @idjevm @kylejvh initially we decided to keep the gtm initializing in the App.js  and also initing it in the NavLayer but that caused the `noscript` tag to appear for each page as the user switched between pages. I now put it in the Home page where the user lands after signing in. if I also keep initing in the App.js as well, there will be one extra `noscript` tag still.. thoughts?

jyclam:
@nr431207 I think the best approach would be to initialize in `App.js` and then pass in the userID in the datalayer after the user has logged in.

nr431207:
@jyclam I tried that approach, along with adding the dataLayer in App.js and then giving it the value of userId in Home component, but it did not include the dataLayer when I tested,  it somehow needs to be re-initialized again after feeding the userId to it. And I just realized, the user doesnt always necessarily land on Home page once they login, so it makes sense to initialize in App as you said, but again userId wont be accessed there since user hasnt logged in.

idjevm:
Talked with Sonia, she will reopen this from Master instead.

jyclam:
@nr431207 Hmm if we init in the login page as is, I don’t think we’re capturing users that don’t hit the login page. I think we also want to track non-logged in users right?

idjevm:
non-logged users do not have an userId... So I think the purpose of this is to only track logged in users.

jyclam:
Right non-logged in users don't have an id, I think we still want to track them? @doheekim93

idjevm:
Yeah but that's a different use case than here. So that would be implemented later for the datalayer where BI wants to track some info

jyclam:
But wouldn't we then need to come back and move the gtm init back to `app.js` at that point? We'd be attaching datalayers to this tagmanager then

doheekim93:
Yeah, we want to be able to track non logged in users as well!

idjevm:
Kind of, yeah working on checking the PR that Sonia open. Again, we are not tracking non logged users this way. Different use case.

jyclam:
Ah maybe there's something I'm missing. How are we tracking non logged in users?

PR #1229, opened 2020-07-20
Hotfix/dont log out privacy page

### LINKED ISSUE #1217: Revisiting other page after aborted /complete-profile bug (edge case, low priority)
This resolution #1187 has likely resolved the one edge-case error while creating another. 

This only happens if you visit another page after aborting `/complete-profile` and can be quickly resolved by refreshing the page . So it should NOT be considered a Soft launch priority.

I'll take some time to figure out the best way to address this.

error-after-aborted-complete-profile

### LINKED ISSUE #1228: Can't complete profile after opening new tab (privacy, terms & conditions)
This is because opening a new tab (or any start-up of the app) queries `/users/current` to authenticate and fetch the JWT cookie.  But because the user isn't created yet it clears the authentication cookie created in the 1st tab after sign-up before complete-profile.

@vikjoshi @pthoma14 @indirabhowmick  FYI

Looks like @rabbi08 was the only one to e2e test sign-up INCLUDING opening the privacy/terms+conditions tab.

This whole thing needs to be refactored as part of this #1164 - but I need to implement a hotfix for soft launch.

joshmorel:

Resolves #1228 

@rabbi08 @vikjoshi @indirabhowmick @pthoma14 this should definitely be merged before invites go out.

I switched from new tabs to modals for user sign up policy page links. This addresses the issue but IMHO this is also a better UX/design. In mobile particularly if the user opens a new tab it can be annoying to get back to the original sign up page.

The pages still exist so other links to this content still works. This includes on create org page, this is not an issue here, but I can update to use modals if you agree about the better option UX wise. 

Also Resolves #1217. I partially undid the earlier attempted fix for this.


client/src/pages/PostPage.js
```diff
@@ -225,6 +225,7 @@ const PostPage = ({ user, updateComments, isAuthenticated }) => {
       response.data.post.types.map((type) => typeToTag(type));
       let copiedpost = Object.assign({}, response.data.post);
       let postSubstring = response.data.post.content;
```
idjevm, line 227: Were this numcomments intended?

joshmorel: Okay this seems okay now

rabbi08:
Created Individual profile after clicking the ‘Privacy policy’ and ‘Terms and Conditions’. Here the NEW behavior is that they open in pop up and user. Closes it either by clicking on ‘x’ present on top left corner or ‘OK’ button present at the end of the page.

https://share.getcloudapp.com/BluO8kv9

Created Organisation Profile, though the here the ‘Privacy policy’ and ‘Terms and Conditions’ in new TAB, No change in the previous behavior.

https://share.getcloudapp.com/NQugP88K

Sanity Testing done around:
- Added the Post as Individual and Organisation.
- View Helpboard
- Onboarding User
- Filter as per location

joshmorel:
Thanks @rabbi08 . I do agree, @idjevm , that we should do the same for createorgprofile but it is not affected by the same issue (logged out because opens new tab) so I didn't want to add here given the other issue is critical to fix.

Are we okay to go ahead and merge? After I will push a tag so it deploys production.

idjevm:
@joshmorel Don't deploy just yet, I will get in a GTM tags PR soon!

joshmorel:
Ah okay @idjevm - we can merge both then push the tag. Merge this if you are fine now - create org profile is same as create user profile.

vikjoshi:
Tested on build url and this looks good to me!

PR #1268, opened 2020-07-25
Fix for issue #1267

### LINKED ISSUE #1267: [Desktop] Update the feedback icon
The feedback icon (envelope icon) needs to be updated by this new icon : https://zpl.io/bWOKBQM

feedback icon 0

Joeu:

_Please be concise and link any related issue(s):_ Fixes issue #1267 


joshmorel (changes requested):
Thanks @Joeu 

This is close to good to merge. But there are two thoughts:

1) It feels like the positioning should be adjusted a little higher (I think this issue existed before but is more noticeable now because of the speech bubble shape) . Thoughts @tejasm2 

Here is current state:

image

Here is removing the `padding-top: 0.4rem` from `NavLinks` `styled.div` `button`. Not sure exactly why it was set originally but I think this looks better.

image

2. `envelope.svg` isn't used anywhere so I think we should delete it as part of this PR. It's dead code now and we still can get it (from git history, zeplin, or figma) if need be.

tejasm2:
Hey @joshmorel , agree with you, the second image looks much better. May be we also need to align the menu buttons in line with FightPandemics ? Any suggestions on this?

joshmorel:
@tejasm2 - yeah it's a bit more out of scope for the issue so I'm leaning towards opening a new issue. It would be nice to know why the padding was added - it's not exactly clear looking at the history. Those who did are either on a break or left the project.

If we change

```
const NavLinks = styled.div`
  align-self: flex-end;
  padding-top: 1rem;
  white-space: nowrap;
  padding-top: 2rem;

```
TO:

```
const NavLinks = styled.div`
  align-self: flex-end;
  white-space: nowrap;
```

Here is what it looks like:

image


Thoughts @idjevm on creating a new issue vs fixing this here?

tejasm2:
Yeah this image looks perfect @joshmorel

joshmorel:
The one issue is @tejasm2 is the hover active is no longer adjacent to top of the main content so that might need some further work to get right. Maybe a new issue would be better?

image

tejasm2:
Yeah @joshmorel , u are right. We should have another ticket for this. Should I create one or will you be creating it? Also the feedback icon color is darker then the join us button...or is it just bold/thick?

joshmorel:
@tejasm2 I opened a new ticket: #1276 

>> Also the feedback icon color is darker then the join us button...or is it just bold/thick?

The "Join Now" button is a different color - `#6A7EFC` - which isn't in the design (it is the primary color - `#425AF2`) or even one of our colors in the style guide. Not sure why it was set that way originally. Thoughts @Yu-Alexa-Zhao . May be worth changing here because it will look better with new feedback icon (current on top, `#425AF2` on bottom).

image
image

tejasm2:
Yeah I think the color should be same.

yu-zhao-xyz:
Yes I agree with @joshmorel, using the primary color `#425AF2` in the design system is better.

idjevm:
@joshmorel @tejasm2 Sorry I got to late to the party haha but I might give it a shot here and also link the new issue. Will take a look.

tejasm2:
No worries brother

Joeu:
@joshmorel , I just deleted the envelope and removed the padding for the button.

joshmorel (approved):
@Joeu. That `button` is the only such element in the NavLinks styled component. So I think it's better to remove the rule there, then leave it and add and inline rule to override it. I added the commit. This is good to merge. Thanks

joshmorel:
I will create a new issue for the Join Now color. It is a good first issue so I can encourage someone who just joined to get into the code base.

idjevm (approved):
LGTM! I was testing some stuff locally for the header to be centered. I will double check on the other issue and submit a PR when I am done

PR #1269, opened 2020-07-25
Update git branching model

### LINKED ISSUE #1127: Update deployment branches - production, staging
`staging` will be replace `master` as main target branch and deploy to `staging` environment

`production` will be new and deploy to `production` environment

### LINKED ISSUE #1130: Document git branch/merge process
Create a draft document in Github

Related: #1127

mannykary:

Update Github workflows based on new branching and deployment model, and update Readme. Also update PR template.

I also did the following (which is not visible in code):
- Added `staging` and `production` branches, with merge/push restrictions to lead engineers only

Resolves #1130, #1127


mannykary:
TO DO after this PR is merged:
- [x] Update default branch to `staging`
- [x] Update all open PRs to point to `staging`
- [x] Merge `rc/v1.1` into `staging` and delete `rc/v1.1` branch (merge this branch using `Merge commit` method)
- [ ] ~Restrict merges to `Squash and Merge`~ (EDIT: I just realized that we shouldn't add this restriction because we wouldn't want to squash commits when merging from `staging` to `production`.
- [x] Delete `master` branch

README.md
```diff
@@ -150,20 +150,49 @@ We collaborate closely with the design and product team. The design team provide
...
+
+- A `hotfix/` branch is created if there is a critical fix that needs to be deployed to production as soon as possible, but staging has untested code that is not safe to deploy to production.
+    - This branch must be created off the `production` branch.
+    - Every push to `hotfix/` branches will automatically deploy to the review AWS environment. See the [Review branches](#review-branches) section for details on how to access this build.
+    - Two pull requests must be opened for hotfixes: one PR to merge into `production` and one PR to merge into `staging`.
+    - Hotfixes must be tested in the staging AWS environment before being merged to the `production` and `staging` branches. In order to deploy only the hotfix to the staging enviroment, create and push a git tag with a `hotfix-` prefix. This will trigger a deployment to the staging environment: `git tag -a hotfix-<GITHUB_ISSUE_NUMBER> && git push origin hotfix-<GITHUB_ISSUE_NUMBER>`.
```
idjevm: I am wondering about it here like if we merge the hotfix to prod and to staging and then we will merge staging to prod, we will have two commits that are doing the same thing in prod right? Should be fine but just wondering if I am thinking correctly. So we won't be back merging prod into staging, but instead, merge hotfixes to both staging and prod...

README.md
```diff
@@ -150,20 +150,49 @@ We collaborate closely with the design and product team. The design team provide
 
 ## Deployment
 
+### Git Branching Model
+
+- When a Github issue is started, create a branch with the `feature/` prefix. Every push to branches with this prefix will automatically deploy to the review AWS environment. See the [Review branches](#review-branches) section for details on how to access this build.
+
+- After a pull request for a `feature/` branch is approved, it is merged into the `staging` branch. This will trigger an automatic deployment to the staging AWS environment. Note that changes may take several minutes to take effect. The staging app is accessible at http://staging.fightpandemics.work.
+
+- When we are ready to release to production, a pull request is opened to merge the `staging` branch to `production`. After this pull request is approved and merged, it will trigger an automatic deployment to the production AWS environment. Note that changes may take several minutes to take effect. The production app is accessible at http://fightpandemics.com
```
idjevm: should we change the prod urls to `https://fightpandemics.com`? This will work too though

idjevm (approved):
besides the URL change, LGTM!

README.md
mannykary: Actually that's a good point; if we use the Github `Squash and merge` button to merge the hotfix branch into both staging and production, we'll have two distinct commit hashes for the same change. It's probably better to backmerge from production to staging instead to get the hotfix into staging. I'll update this.

README.md
mannykary: Either works, but probably best practice to use HTTPS

README.md
```diff
@@ -150,20 +150,50 @@ We collaborate closely with the design and product team. The design team provide
...
+
+- A `hotfix/` branch is created if there is a critical fix that needs to be deployed to production as soon as possible, but staging has untested code that is not safe to deploy to production.
+    - This branch must be created off the `production` branch.
+    - Every push to `hotfix/` branches will automatically deploy to the review AWS environment. See the [Review branches](#review-branches) section for details on how to access this build.
+    - Two pull requests must be opened for hotfixes: one PR to merge into `production` and one PR to backmerge `production` into `staging` after the initial PR is merged to `production`.
+    - Hotfixes must be tested in the staging AWS environment before being merged to the `production` and `staging` branches. In order to deploy only the hotfix to the staging enviroment, create and push a git tag with a `hotfix-` prefix. This will trigger a deployment to the staging environment: `git tag -a hotfix-<GITHUB_ISSUE_NUMBER> && git push origin hotfix-<GITHUB_ISSUE_NUMBER>`.
```
joshmorel:
```suggestion
    - Hotfixes must be tested in the review AWS environment before being merged to the `production` and `staging` branches. In order to deploy only the hotfix to the staging enviroment, create and push a git tag with a `hotfix-` prefix. This will trigger a deployment to the staging environment: `git tag -a hotfix-<GITHUB_ISSUE_NUMBER> && git push origin hotfix-<GITHUB_ISSUE_NUMBER>`.
```

Besides this substitution I think the docs and workflow files look good. Thanks @mannykary !

joshmorel (changes requested):

README.md
```diff
@@ -150,20 +150,50 @@ We collaborate closely with the design and product team. The design team provide
...
+
+- All pull request merges to the `staging` branch, and all `hotfix/` merges to the `production` branch are done using the `Squash and Merge` strategy. This allows for a cleaner commit history by combining all commits in a branch into a single commit. The message for this squashed commit must use the following convention:
```
robvenn, line 169: If i've read the other comments correctly this part should also be changed and for `hotfix` we can't squash to prevent different commit hashes ...

robvenn (commented):
Ehm yeah I missed a lot of the dicussions but lgtm 
I wonder why creating merge commits instead of rebase? Or merge using fast-forward if possible without the extra commits? This is usually a cleaner git history afaik.

README.md
mannykary: @josh It's actually staging. That's the reason why I added the tag trigger for `hotfix-` so that we can deploy any arbitrary branch the staging environment. We wouldn't want to test the hotfix only in the review environment as it isn't as "fully baked" as staging.

README.md
mannykary: Yeah, that's why I changed [this](https://github.com/FightPandemics/FightPandemics/pull/1269/commits/80220c554a90eee5f7aa288722dc2be3680c7864#diff-04c6e90faac2675aa89e2176d2eec7d8R164) to instead backmerge from the `production` branch after the hotfix PR is merged to `production`, rather than merge directly from the hotfix branch into `staging`. I also added [this](https://github.com/FightPandemics/FightPandemics/pull/1269#diff-04c6e90faac2675aa89e2176d2eec7d8R181) to indicate that when merging from `staging` to `production` or vice versa, we should use `Create a merge commit` instead of `Squash and merge`

README.md
joshmorel: Okay. In that case, the order of action as well as rationale for this is not quite clear. It reads like it's a) deployed to review environment (for testing I would assume?), b) PR is merged into `production` c) PR is backmerged into `staging`. But somewhere in between it is also pushed to `staging` environment. After what and why is not quite clear... How is not quite "fully baked"?

mannykary: Let me know if this is clearer: https://github.com/FightPandemics/FightPandemics/pull/1269/commits/be41a69d7c0fe0dc36eea2181e1bc128c086eb6e

When I say the review environment is not "fully baked", I mean that I don't consider the review environment to be a valid environment for QA testing for the following reasons:
- All the review builds read from and write to the same database, and also use the same Auth0 tenant, so data integrity is not really guaranteed
- It does not quite have the same infrastructure as the staging and production environments - the way it's set up is the load balancer forwards requests to the appropriate Fargate task by looking at the subdomain of the URL. This does not exist in staging/production.
- There is a cleanup job that kills review builds that haven't been deployed to in more than a day, so that could potentially throw off QA testing. I anticipate adding more cleanup jobs to rein in our resource usage so as to minimize expenses
- This hasn't been implemented yet, but I plan to vertically scale down the Fargate tasks to provision a lot less memory

The review builds are really more of a convenience for developer testing rather than official QA testing. That is not to say that we can't do ANY QA testing in the review build. I just wouldn't consider review build QA testing as the final green light to merge into the `production` branch and deploy to production. The staging environment mirrors the production environment more closely, so it would be more prudent to test in staging rather than the review environment.

joshmorel (approved):
LGTM let's try this out!

README.md
joshmorel: Thanks for clarifying @mannykary

PR #1274, opened 2020-07-26
Feature/fixed list pagination

### LINKED ISSUE #927: Add pagination for profile activity
We added a skip pagination (via `limit=-1`) to posts endpoint to get this working quickly.

However, some definitely take awhile to load for sourced by fp due to hundreds of posts. So we should add pagination similar to scroll on feed.

Ideally this would be some kind of higher order component but copy/paste is okay for now if that is not straight forward.

OzkanAbdullahoglu:
Resolves #927 


rabbi08:
Adding below issue for tracking:
Issue: while loading data for Orgnaisation profile, UI get distorted:

device: iphone 11 pro max
https://share.getcloudapp.com/lluYeAyZ

joshmorel (changes requested):
Thanks @OzkanAbdullahoglu . 

The loader indicator is working except it still shows up if the filter has no cases:

image

In such a case we don't want it there - just the message.

Also for profile activity can we also use the loader? It's hard to notice (I need to put devtools network throttling on) - but you see the "No activity" when it is loading (
https://featurefixed-list-pagination.fightpandemics.xyz/organisation/5eff67aa7002103f35cfca9d). Side note: I'm unsure why it's so much faster on activity compared to feed, my guess is - the geo distance sort is skipped).

Otherwise functionality seems good. 

Not sure how much of a deep dive I'll take into the code line-by-line as it's been so dramatically changed for integration with `react-virtualized` library.

OzkanAbdullahoglu:
@joshmorel @idjevm  Well I tried to cover everything, all edge cases, `emptyFeed()`, `Loader`.  Also, there is a current bug in staging right now.  I added some lines to prevent that one too. 
This is from staging. Current logic loses some of your posts after post deletion and after loading following page. 



And yes, Feed.js is a bit slower (in the first mount, on the other hand, scrolling performance is good) cause of filtering. I separated `objectiveURL()` and `filterURL()` from `loadPosts()`, by this way I aimed to not re-run these functions in every re-fetch, just run them when they needed.

OzkanAbdullahoglu:
React virtualized is a library for rendering large lists more effectively. We use  `WindowScroller` for a full-size scrolling view (like Twitter, Facebook, etc.), `InfiniteLoader`, for fetching when scrolling down, `List` to render elements. 

Examples and docs can be accessed below links.

https://github.com/bvaughn/react-virtualized
https://bvaughn.github.io/react-virtualized/#/components/List

joshmorel (approved):

client/src/components/Feed/Post.js
```diff
@@ -521,7 +521,6 @@ const Post = ({
           />
         </>
       ) : (
```
joshmorel:
```suggestion
      ) : (
        //Post in feed.
```

I don't think we should remove this comment - it's still helpful.

joshmorel: I will just reset this file to staging & repush

joshmorel (changes requested):

joshmorel (approved):

client/src/components/Feed/Post.js
OzkanAbdullahoglu: I haven't realised & also do not remember that I removed this comment. anyway, I agree.

idjevm (approved):
Let's give it a shot! Thanks for all the work on this one @OzkanAbdullahoglu !!!

rabbi08:
@idjevm @OzkanAbdullahoglu @joshmorel : Let me know if I should be logging new issue to track below observation:

Tested this PR on mobile and Desktop in STG:
- Desktop version is working fine: 
- Clicking on company profile on mobile, the UI is not distorted as reported above: PASS
- But on the mobile the loading spinner exist along with the message about no post. "Filter" is not working in STG on mobile. : FAIL
Refer below screenshot:
Devices Tested:
- iphone 11 pro max
- Tested on simulator: Samsung Galaxy, iphone XR

Android:
https://share.getcloudapp.com/NQu1LrR6
iOS:
https://share.getcloudapp.com/OAuJ0qxx

OzkanAbdullahoglu:
Hi @rabbi08 , yes that is my bad, I forgot a simple implementation for mobile. This is an easy fix, I can handle that under a new branch which I will bind to #1545

PR #1312, opened 2020-08-02
Fix Terraform deprecated host_header condition syntax

mannykary:

Deployment with Terraform is broken because of an update to a new version which deprecated the syntax for the `condition` block for the `aws_alb_listener_rule` resource.

Since this is a critical fix impacting review builds and staging, I will merge this in right away and post an announcement on Slack to rebase the latest from staging to fix deployments.

### DIFF
--- a/terraform-task-module/alb-listener-rule.tf
+++ b/terraform-task-module/alb-listener-rule.tf
@@ -17,8 +17,9 @@ resource "aws_alb_listener_rule" "main" {
     target_group_arn = aws_alb_target_group.app.arn
   }
   condition {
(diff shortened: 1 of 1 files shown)


idjevm:
Thanks @mannykary ! Yeah I noticed last night when I was helping someone with their branch, that it was something to do with terraform that was not letting it build

git-satya (commented):
thanks Manny. Syntax looks good as per AWS Terraform registry documentation. 
  
condition {
    host_header {
      values = [ ...  ]
    }

PR #1321, opened 2020-08-04
add _id as posts sort field for determinate sort/pagination

joshmorel:
### What does this pull request aim to achieve? 💯

Resolves issue #1320 

I do note in commit this still doesn't handle case of post added/deleted after page load but that is more complicated to address (and only affects specific point-in-time page views, instead of multiple).

@mannykary  I figured this is appropriate for a hotfix - not sure how many posts it is affecting but likely enough to warrant it. This is a good opportunity to test out the process. The `hotfix-` tag, I assume I add & push this after I get all dev approvals, but then do I message in `#testing` to tell QAs that I need someone to test and staging will not work for non-hotfix testing OR do we get an agreed to time? (we can discuss in slack) 

- [x] I have checked that no one else is working on similar changes.
- [x] I have read the latest [**README**](README.md) and understand the development workflow.
- [x] The name of this branch is: **`feature/<branch_name>`**, or **`hotfix/<branch_name>`** if this is a hotfix. The branch is created in a cloned version of this repo, **not a forked repo**.
- [x] This branch is rebased/merged with the **latest staging** (or the **latest production** if it is a hotfix).
- [x] There are no merge conflicts.
- [x] There are no console warnings and errors.
- [x] On the right hand side of this PR, I have linked the appropriate issue(s) under "Issues" and added a project progress under "Projects".

PR #1322, opened 2020-08-04
Hotfix/posts determinate pagination

### LINKED ISSUE #1320: Some posts are skipped due to lack of deterministic sorting
This is a tricky one to reproduce but can be seen in production by:

* Visiting https://fightpandemics.com/feed  (log off if currently logged in)
* Filtering for "Requesting help" + "Providers" = "Startup" & scroll til the bottom (total displayed posts 7)
* "Covid Blood Drive" will be one of the posts
* Update filter to "Requesting help" + "Providers" in ( "Startup", "University" ) & scroll til the bottom (total displayed posts 14)
* "Covid Blood Drive" will NOT be shown

In the DB for this filter there are actually 15 not 14.

posts-not-displayed-due-to-sorting

The reason is there are multiple with same `createdAt` which is used as sorting criteria - a common situation due to posts imported from AirTable. Sorting after `createdAt` is not determinant so usage of `$skip`/`$limit` results in the same post being fetched with each call. 

It's very hard to determine how many posts are being skipped because of this case but the above example shows that it does happen. FYI on the front-end the posts are being merged by `_id` key into an object instead of appended to an array, that's why there aren't duplicates displayed.

The fix is to sort on `_id`, this is an `ObjectId` where the first 4 bytes are second timestamp so all created at the same second will sort first, the rest will be unique meaning it will be a deterministic sort and `$skip`/`$limit` will work as expected (https://docs.mongodb.com/manual/reference/method/ObjectId/).

We should also add a secondary sort for geodistance in case of scenarios with the same distance (likely uncommon but worth handling). 

```javascript

      const sortAndFilterSteps = location
        ? [
            {
              $geoNear: {
                distanceField: "distance",
                key: "author.location.coordinates",
                near: {
                  $geometry: {
                    coordinates: location.coordinates,
                    type: "Point",
                  },
                },
                query: { $and: filters },
              },
            },
            { $sort: { distance: 1, _id: -1 } },
          ]
        : [{ $match: { $and: filters } }, { $sort: { _id: -1 } }];
```

joshmorel:
Resolves issue #1320

I do note in commit this still doesn't handle case of post added/deleted after page load but that is more complicated to address (and only affects specific point-in-time page views, instead of multiple).

Created a new branch since the other #1321 didn't deploy.

* [x] I have checked that no one else is working on similar changes.
* [x] I have read the latest README and understand the development workflow.
* [x] The name of this branch is: feature/<branch_name>, or hotfix/<branch_name> if this is a hotfix. The branch is created in a cloned version of this repo, not a forked repo.
* [x] This branch is rebased/merged with the latest staging (or the latest production if it is a hotfix).
* [x] There are no merge conflicts.
* [x] There are no console warnings and errors.
* [x] On the right hand side of this PR, I have linked the appropriate issue(s) under "Issues" and added a project progress under "Projects".

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -7,9 +7,9 @@ _Please be concise and link any related issue(s):_
 <!--_Please do **NOT** submit this PR if you have not done the following:_-->
 
 - [ ] I have checked that no one else is working on similar changes.
-- [ ] I have read the latest [**README**](https://github.com/FightPandemics/FightPandemics/blob/master/README.md) and understand the development workflow.
-- [ ] The name of this branch is: **`feature/<branch_name>`** (need to have cloned the repo not forked).
-- [ ] This branch is rebased/merged with the **latest master**.
+- [ ] I have read the latest [**README**](README.md) and understand the development workflow.
+- [ ] The name of this branch is: **`feature/<branch_name>`**, or **`hotfix/<branch_name>`** if this is a hotfix. The branch is created in a cloned version of this repo, **not a forked repo**.
+- [ ] This branch is rebased/merged with the **latest staging** (or the **latest production** if it is a hotfix).
 - [ ] There are no merge conflicts.
 - [ ] There are no console warnings and errors.
 - [ ] On the right hand side of this PR, I have linked the appropriate issue(s) under "Issues" and added a project progress under "Projects".
--- a/.github/workflows/production-workflow.yml
+++ b/.github/workflows/production-workflow.yml
@@ -1,9 +1,15 @@
-name: production
+name: Deploy to Production
 
 on:
   push:
-    tags:
-      - 'v*'
+    branches:
+      - production
+    paths:
+      - backend/**
+      - client/**
+      - main.tf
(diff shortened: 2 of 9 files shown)


idjevm (approved):
LGTM!

joshmorel:
EDIT: you can see here if filtering "Requesting help" and providers "Startup" (not logged in)  https://featureredirect-to-login.fightpandemics.xyz/feed there are only 6. But https://hotfixposts-determinate-pagination.fightpandemics.xyz/feed shows 7 ("123 Bite") was missing.

~So I was hoping I could see it working in this review branch but not another but they both work:  https://featureredirect-to-login.fightpandemics.xyz/feed~

~Because `_id` will be different for same post in different environment the current state pagination could still "work" in some cases.~

~I'll see if there is another example to demonstrate, but  you can pull locally with a mongodump I have to see the difference. ~

~If the later is required - the key for QA testing is simply that sort by distance (if logged on) and time (if not) still work  as before.~

Naraujo13 (approved):
LGTM

PR #1328, opened 2020-08-05
Removed unused Org Complete file and refactored ProfileCompleted.js per user path

### LINKED ISSUE #1306: Refactor and remove OrgProfileComplete page, instead use the ProfileCompleted page
The call to action button should say `View Help Board` unless the unlogged in user clicked on `Create a Post` and went through the sign up process, in which case it should say `Continue posting` 

# Currently the pages look like this:

## OrgProfileComplete
screencapture-localhost-3000-create-organisation-complete-2020-07-30-17_47_07

## ProfileCompleted
screencapture-localhost-3000-profile-completed-2020-07-30-17_46_32

kterry1:
- Removed OrgProfileComplete.js as ProfileCompleted.js is essentially doing the same thing. 
- Added sessionStorage to the Create a Post button in Feed.js when a user attempts to create a post but isn't logged in. 
- If the user isn't logged in and attempts to create a post, they will be taken to the sign in page, then when they see complete their profile and are taken to /profile-completed, the second CTA button will say "Continue Posting", otherwise, the same button will say "View Help Board".
- ProfileCompleted.js also replaced OrgProfileComplete.js for the /create-organisation-complete path and subsequently the ProfileCompletedButtonsWrapperProfile replaced the ButtonContainer component. 
This will fix and close out: #1306

- [x ] I have checked that no one else is working on similar changes.
- [x ] I have read the latest [**README**](README.md) and understand the development workflow.
- [x ] The name of this branch is: **`feature/<branch_name>`**, or **`hotfix/<branch_name>`** if this is a hotfix. The branch is created in a cloned version of this repo, **not a forked repo**.
- [x ] This branch is rebased/merged with the **latest staging** (or the **latest production** if it is a hotfix).
- [x ] There are no console warnings and errors.
- [x ] On the right hand side of this PR, I have linked the appropriate issue(s) under "Issues" and added a project progress under "Projects".


joshmorel (changes requested):
Thanks @kterry1 . I haven't looked into redirect but there are a few things we need to check for if we're going to combine the page components.

The 1st is in the current state the "View my Profile" is a link to the organisation profile. `CreateOrganisationProfile` is already including `orgId` as `state` with redirect so we can use that. 

import the `useLocation` hook from react-router-dom.

Before `return (`

```javascript
const location = useLocation();
```

Then instead we can have a conditional link depending on this:

```javascript
        <Link
          to={
            location?.state?.orgId
              ? `organisation/${location.state.orgId}`
              : `${PROFILE}/${user?.id || user?._id}`
          }
        >
```

2. Is we refetch the user so we're updating the profile nav organisation list state (otherwise you add an org and you need to refresh the page to see it).

So basically use similar code to the deleted `OrgProfileComplete` for this.

It's not necessarily, although not harmful, to refetch after user profile complete. So you could add logic before dispatching function `if (location?.state?.orgId) {`

Feel free to reach out in Slack for anymore questions. I'll take a look at the rest soon.

kterry1:
I'm not sure I entirely understand what's being asked here, but I'll review the code today and see if I have any further questions. Thanks!

Edit: I think I'm on board now. Will make the changes tonight and hope to have them available for review tomorrow unless I run into a snag.

joshmorel:
Hi @kterry1  -  might be better to check the current state in `staging` (branch or environment) to see what `OrgProfileComplete` is doing. They are "almost" doing the same thing but not quite.

On further thought:

* Personally, actually, I think it is better to have re-usable "ProfileComplete" components (for buttons etc) rather than the same whole component for both routes. We've run into a bit of a mess in using the same component for multiple auth-related pages.
* If we migrate to Next.js (likely) you need to have different files for each route anyways so I think going forward this is an overall better approach (more re-usable components, BUT each route is its own file)

kterry1:
@joshmorel Sorry, I didn't see your updated post until later. The changes I made were based off your first suggestion. Just saw that you thought about a ProfileCompleted component. The second commit should resolve this ticket but if you think it's better that I go back and create a ProfileCompleted component, I can. Or I can create a new ticket for that specific request. I'll re-request for you to review this PR though. Just let me know your thoughts.

joshmorel:
> @joshmorel Sorry, I didn't see your updated post until later. The changes I made were based off your first suggestion. Just saw that you thought about a ProfileCompleted component. The second commit should resolve this ticket but if you think it's better that I go back and create a ProfileCompleted component, I can. Or I can create a new ticket for that specific request. I'll re-request for you to review this PR though. Just let me know your thoughts.

Nah let's go with this approach. I don't want to over complicate things for this PR/issue. If we get a new design for org created then we can revisit later.

I will review today.

client/src/pages/ProfileCompleted.js
```diff
@@ -1,6 +1,7 @@
-import React from "react";
-import { Link } from "react-router-dom";
+import React, { useEffect } from "react";
+import { Link, useLocation, withRouter } from "react-router-dom";
```
joshmorel: Hey @kterry1  - I looked through everything else and tried this and I'm good. 

But I forgot to mention the original uses `withRouter` - but we should use `useLocation` or `withRouter` not both. There is an indication `withRouter` might be retired - https://reacttraining.com/blog/react-router-v5-1/ - so let's use `useLocation`.


```suggestion
import { Link, useLocation } from "react-router-dom";
```

client/src/pages/ProfileCompleted.js
```diff
@@ -39,12 +52,18 @@ const ProfileCompleted = ({ user }) => {
...
     </ProfileCompletedWrapper>
   );
 };
 
-export default ProfileCompleted;
+const mapDispatchToProps = {
+  refetchUser,
+};
+
+export default connect(null, mapDispatchToProps)(withRouter(ProfileCompleted));
```
joshmorel: As per above - remove `withRouter`

```suggestion
export default connect(null, mapDispatchToProps)(ProfileCompleted);
```

joshmorel (changes requested):
Besides removing `withRouter` everything else looks good. Thanks @kterry1

kterry1:
@joshmorel Changed! Thanks for the help with this!

joshmorel (approved):
LGTM! Thanks!

idjevm (approved):

idjevm (approved):
LGTM! Thanks @kterry1 !

PR #1350, opened 2020-08-10
Mocha-integration API backend testing - new service

galeza:
Create new mocha API integration service

This service will perform API tests. 
We are going to use modules: should, supertest, chai, http-status 
Service waits for client and when http://backend-service:8000 proxy is created we start tests.
After mocha-integration service cypresse2e starts tests.
Mocha-integration works in the same manner as cypresse2e service and it was build similarly.

It's only service structure, two tests are only as an example. Test framework/structure will be added.

Screen Shot 2020-08-10 at 21 20 24


galeza:
Tests runs:
Screen Shot 2020-08-14 at 19 14 03

For QAs in case of problems with building backend service:

> docker system prune -a (enter y)
> docker system prune --volumes (enter y again)
> finally docker-compose up --build

mocha/package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "mocha",
+  "version": "1.0.0",
+  "description": "Mocha API integration tests",
+  "scripts": {
+    "test": "mocha './mocha/test/**/*.js' --recursive --timeout 60000 --exit"
```
jessica-lau: I think the test path should be changed to the following instead: 
"test": "mocha './test/**/*.js' --recursive --timeout 60000 --exit"

I ran the test on my local machine while in the parent folder using the original test path provided and it triggered an error because no package.json file exists in the parent folder. I ran the test while in the mocha folder using what I suggested above and it worked. 

image

mocha/package.json
```diff
@@ -0,0 +1,12 @@
...
+  },
+  "private": true,
+  "devDependencies": {
+    "mocha": "^8.1.0"
+  }
```
jessica-lau, line 15: Missing dependencies, need to run npm install --save-dev should supertest chai http-status

jessica-lau (changes requested):
The test itself looks great! Just have some comments on the mocha/package.json file :)

mocha/package.json
galeza: ok. I'll add. dependencies

mocha/package.json
galeza: @jessica-lau 
How do you want to change the test path?

jessica-lau: @galeza currently the test path in the package.json is this:
"test": "mocha './mocha/test/**/*.js' --recursive --timeout 60000 --exit"

but it should be changed to this:
"test": "mocha './test/**/*.js' --recursive --timeout 60000 --exit"

galeza: @jessica-lau 

I added devDependencies to package.json and modified also docker-compose to start tests using :_'./test/**/*.js'_
Please verify.

jessica-lau (approved):
LGTM 👍

mannykary (approved):
LGTM.

PR #1415, opened 2020-08-24
Refetched User to update Org name in nav dropdown after saving changes

### LINKED ISSUE #1409: Organisation name not updated in menu after update
Org name is not updated in menu list after editing.

org-name-not-updated

That is until page refresh. This is a React state management issue.

This can be resolved by re-implementing changes in this commit for `EditOrganisationAccount.js` file 33ec75c

kterry1:
After saving changes for the org name, the Org name will now update in the nav dropdown without needing to manually refresh the page: https://share.getcloudapp.com/WnuJkemq
Resolves: #1409 

- [X] I have read the latest [**README**](README.md) and understand the development workflow.
- [X] The name of this branch is: **`feature/<branch_name>`**, or **`hotfix/<branch_name>`** if this is a hotfix. The branch is created in a cloned version of this repo, **not a forked repo**.
- [X] This branch is rebased/merged with the **latest staging** (or the **latest production** if it is a hotfix).

### DIFF
--- a/client/src/pages/EditOrganisationAccount.js
+++ b/client/src/pages/EditOrganisationAccount.js
@@ -5,11 +5,13 @@ import Checkbox from "components/Input/Checkbox";
 import { WhiteSpace } from "antd-mobile";
 import FormInput from "components/Input/FormInput";
 import { Alert } from "antd";
+import { connect } from 'react-redux';
 import SuccessAlert from "components/Alert/SuccessAlert";
 import { ORANGE_RED, WHITE } from "../constants/colors";
 import { Link } from "react-router-dom";
 import InputLabel from "components/Input/Label";
 import orgData from "../assets/data/createOrganisationProfile";
+import { refetchUser } from "actions/authActions";
 import {
   EditLayout,
   TitlePictureWrapper,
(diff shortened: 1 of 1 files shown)


soniaroh (approved):
tested the build url, looks good!

galeza (approved):

galeza:
LGTM. Tested on build URL - on desktop and mobile

idjevm (changes requested):
Isn't something similar being done with Context? I think we should stick to context if it is possible to avoid having both context and redux in our app. @joshmorel didn't we do something similar to this with Context? Lmk what you think

joshmorel:
@idjevm this was the approach I had suggested for the issue. Because `sessionReducer` is using `redux` we need to use it here as well as `useContext` outside of a larger refactor.

The larger refactor still lacks consensus and direction. If we replace `redux` with hooks we need to come up with some kind of better guidance for devs to follow then what we currently have.

Alternatively we can further with `redux` and use it better - also adopting other libraries `redux-thunk` or `redux-sagas` etc. Since we are working so much with calls to the backend API this might actually be the better/easier approach for our project. 

But I'm not sure... If we can find a frontend engineering someone interested in taking on either with a proposal I'd support them! Then we can implement after the wide launch.

joshmorel (approved):
LGTM

idjevm (approved):
Sweet! Sounds good then! Thanks Josh and Kevin!

sheeba1:
Tested on staging. Editing the Org name reflected on profile-drop down org name and existing post

rabbi08:
Tested this new feature on staging, it works fine, the new EDITed name is being updated for existing older post too along with the nav dropdown. Thanks

Issue #1387, opened 2020-08-17
Add meta (counts) to GET /posts


joshmorel:
This is for chatbot pagination but would be useful elsewhere. Currently we are returning `[...]` we should still default to this & include a parameter to include a meta object (common among REST APIs) that we can add other stuff to. 

The result would be:

```json
{
  "meta": 
     { "total": 456 }
  },
  "data": [{}, {}, {}, {}, {}, {}]
}
```

Where `"total"` is the number in the result set based on filters EXCEPT for pagination parameters (skip/limit).

An alternate might be return an overall total as well as a filtered count. 

```json
{
  "meta": 
     { "total": 3478, "filtered": 456 }
  },
  "data": [{}, {}, {}, {}, {}, {}]
}
```

We could even include a count & skip/limit in meta so client doesn't necessarily need to keep track of these but this may be overkill..

```json
{
  "meta": 
     { "total": 3478, "filtered": 456, "count": 6, "skip": 450, "limit": 10 }
  },
  "data": [{}, {}, {}, {}, {}, {}]
}
```

joshmorel:
Resolved by #1446

PR #1446, opened 2020-08-26
Add meta (counts) to GET /posts #1387

lmpeixoto:
Created includeMeta query parameter (set default to false) that if true sends a response within an envelope. The envelope format is:

{
  "meta": 
     { "total": 456 }
  },
  "data": [{}, {}, {}, {}, {}, {}]
}

### DIFF
--- a/backend/lib/endpoints/posts.js
+++ b/backend/lib/endpoints/posts.js
@@ -228,13 +228,10 @@ async function routes(app) {
       // Get the total results without pagination steps but with filtering aplyed - totalResults
       /* eslint-disable sort-keys */
       const totalResultsAggregationPipeline = await Post.aggregate([
-        ...sortAndFilterSteps,
-        ...lookupSteps,
+        { $match: { $and: filters } },
         { $group: { _id: null, count: { $sum: 1 } } },
       ]);
       /* eslint-enable sort-keys */
-
-      const totalResults = totalResultsAggregationPipeline[0].count;
 
       const [postsErr, posts] = await app.to(
         Post.aggregate(aggregationPipelineResults),
@@ -242,13 +239,16 @@ async function routes(app) {
 
       const responseHandler = (response) => {
         if (!includeMeta) {
-          return response;
-        } else {
-          const responseToReturn = { meta: { total: null }, data: null };
-          responseToReturn.data = response;
-          responseToReturn.meta.total = totalResults;
-          return responseToReturn;
+          return posts;
         }
+        return {
+          meta: {
+            total: totalResultsAggregationPipeline.length
+              ? totalResultsAggregationPipeline[0].count
+              : 0,
+          },
+          data: response,
+        };
       };
 
       if (postsErr) {
--- a/backend/lib/endpoints/posts.js
+++ b/backend/lib/endpoints/posts.js
@@ -46,6 +46,7 @@ async function routes(app) {
         limit,
         objective,
         skip,
+        includeMeta,
       } = query;
       const queryFilters = filter ? JSON.parse(decodeURIComponent(filter)) : {};
       let user;
@@ -132,6 +133,7 @@ async function routes(app) {
 
       // _id starts with seconds timestamp so newer posts will sort together first
       // then in a determinate order (required for proper pagination)
+      /* eslint-disable sort-keys */
       const sortAndFilterSteps = location
         ? [
             {
@@ -150,7 +152,8 @@ async function routes(app) {
             { $sort: { distance: 1, _id: -1 } },
           ]
         : [{ $match: { $and: filters } }, { $sort: { _id: -1 } }];
-
+      /* eslint-enable sort-keys */
+      /* eslint-disable sort-keys */
       const paginationSteps =
         limit === -1
           ? [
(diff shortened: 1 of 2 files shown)


joshmorel:
Thanks @lmpeixoto  for this work.

I realize now I've probably complicated the issue a bit with all the various options, but it wasn't quite clear until now. 

Looking back at [the Laravel example](https://github.com/marcelgwerder/laravel-api-handler#include-meta-information) I cited, the `meta-total-count` is the total minus filters in query params (skip, limit, others...) not the total in the database (e.g. in our case we would never want to count `meta-filter-count`). `meta-filtered-count` is the total minus skip/limit params only.

Really for our pagination use case right now we only care about what the result would be if there was no skip or limit (but all other query params & backend-only filters were applied) - which is like `meta-filtered-count` in example above - but I'm not sure if that's what I want to call it.

If you look at https://jsonapi.org/examples/ - there is no standard for meaning/naming of elements in `"meta"` so it's up to us: 

> Note: Putting a property like "totalPages" in "meta" can be a convenient way to indicate to clients the total number of pages in a collection (as opposed to the "last" link, which simply gives the URI of the last page). However, all "meta" values are implementation-specific, so you can call this member whatever you like ("total", "count", etc.) or not use it at all.

For our use case now all we need is the one count (with other filters but minus skip/limit) - we can add more later. But I don't like calling it "filtered" like above.

Personally, I doubt, because we are following social media model, we'll ever show something like "Filtered 400 of 5989 results". So maybe just call it "total" or "totalCount" for pagination use case? Thoughts @robinv85 , @Naraujo13 

E.g. 

```
"meta": {
  "total": 346
},
"data": [{},{},{},{},{}]
```

Anyways @lmpeixoto  once we make a decision you can make modifications - but chime in if you have you're own ideas.

joshmorel:
Also my suggestion of including limit/skip/count in meta, I think that was more than we need since the frontend knows about them already.

lmpeixoto:
Thanks for your review Josh. Commited the changes.

backend/lib/endpoints/posts.js
```diff
@@ -207,19 +216,49 @@ async function routes(app) {
...
+        ...paginationSteps,
+        ...lookupSteps,
+        ...projectionSteps,
+      ];
+
+      // Get the total results without pagination steps but with filtering aplyed - totalResults
+      /* eslint-disable sort-keys */
+      const totalResultsAggregationPipeline = await Post.aggregate([
+        ...sortAndFilterSteps,
+        ...lookupSteps,
```
joshmorel:
```suggestion
```

We don't need to include lookupSteps for total - we really just need the filters. The sort is also unnecessary - and at scale but create a drag on performance. Not sure if MongoDB is smart enough to ignore them since they are not necessary for the result but I doubt it. 

We really just need `$match: { $and: filters } }` plus the grouping.

backend/lib/endpoints/posts.js
```diff
@@ -207,19 +216,49 @@ async function routes(app) {
...
+        { $group: { _id: null, count: { $sum: 1 } } },
+      ]);
+      /* eslint-enable sort-keys */
+
+      const totalResults = totalResultsAggregationPipeline[0].count;
```
joshmorel:
```suggestion
```


This will throw an error if there are no results - []

```json
{
    "statusCode": 500,
    "error": "Internal Server Error",
    "message": "Cannot read property 'count' of undefined"
}
```

My suggestion would be to skip this line & return something like...

```javascript
return {
  meta: { total: totalResultsAggregationPipeline.length ? totalResultsAggregationPipeline[0].count : 0 },
  data: response
}
```

joshmorel (changes requested):
Thanks @lmpeixoto this is the right direction definitely now that I see it vs the other option.

Eventually we should do some kind of middleware but let's keep it simple for now.

Besides the 2 other comment - another idea for improvement would be to just return the `posts` if `!includeMeta`. If not then follow with `totalResultsAggregationPipeline` etc. But this can also be left to when implementing middleware. 

FYI when returning from a function in an if block you don't need the else. This is purely a stylistic difference but I tend to find skipping the else "cleaner" and easier when reading through code, e.g. :

```javascript
if (something) {
   return X;
}
// other logic
return Y;
```

Tends to look cleaner than:

```javascript
if (something) {
   return X;
} else {
  // other logic
  return Y;
}
```

backend/lib/endpoints/posts.js
lmpeixoto: Thanks again for your review Josh. Commited the changes.

joshmorel (approved):
Thanks @lmpeixoto I'm happy with it. This will meet our need.

We can probably make some further improvements but with a middleware/decorator later. 

I look especially from @Naraujo13 for a 2nd review if I missed anything major.

joshmorel:
I've gone ahead and merged - @OzkanAbdullahoglu you can use this for the FE pagination & @CamilaArias the chatbot team can use this as well.

We can make improvements with a plugin/middleware later

Naraujo13 (commented):
LGTM

A bit late to the party, but seems okay

PR #1452, opened 2020-08-27
Implemented message option to socials of posts

### LINKED ISSUE #1433: Messaging: Add "Message" Button to Post
As a user, I want to send a message to the Author of a Post in order to connect with them to figure out how to help or how to get help without other users seeing my message.

This story is to add the message button on the post within the Help Board and to check for authentication when clicked.

https://www.figma.com/file/4cV84H1jyMo0rJEAl3HNnJ/FightPandemics-landing-page?node-id=10956%3A15744 

image

Users trying to click on this button need to be logged in.
If users are not logged in, they should be redirected to login.

AC:
Given user is looking at a post on the help board,
Then they will see the "Message" icon

Given user clicks on the "Message" icon and is logged in
Then user can proceed to the message submission page (future story)

Given user clicks on the "Message" icon and is not logged in,
Then user will be redirected to log in.

xreic:
Added button to allow users to message the authors of posts.
Currently, doesn't have message sending functionality.

Note that the assets only contain a blue mail icon.

Closes #1433 


client/src/components/Feed/PostSocial.js
```diff
@@ -198,6 +204,56 @@ const PostSocial = ({
           <StyledSpan>Share</StyledSpan>
         </div>
       </div>
+
+      <span></span>
+      {postId ? (
+        // NOTE: May need new function to trigger message sending feature on click
+        <div id={gtmTag("message", GTM.post.prefix)} className="social-icon">
```
idjevm: Please update these to the right comment formatting.

client/src/components/Feed/PostSocial.js
```diff
@@ -198,6 +204,56 @@ const PostSocial = ({
...
+          {renderMessageIcon()}
+          {renderLabels("Message", numComments)}
+        </div>
+      ) : (
+        <>
+          {isAuthenticated ? (
+            // NOTE: Update pathname and state when message submission page story is ready
+            <Link
```
idjevm: same as above

client/src/components/Feed/PostSocial.js
```diff
@@ -82,6 +83,11 @@ const PostSocial = ({
     );
   };
 
+  const renderMessageIcon = () => (
+    // NOTE: No gray mail icon within assets
+    <StyledSvg src={mail} className="social-icon-svg" />
```
idjevm, line 89: 👀 Some here too I believe.

idjevm (changes requested):
I left a couple comments and also I don't think we will want to show this -> 
image

This is not my post either but it is showing on the post's page.

client/src/components/Feed/PostSocial.js
xreic: Thank you for pointing these out!

idjevm (approved):
LGTM! As per the scope of the issue looks good to go!

rabbi08:
@emtay 
This feature is similar to the 'comment' already existing for the post, as if user want to communicate or share their viewpoint, comments serve the same purpose, adding additional duplicate feature with a difference that now that email will send to the author of the post. 

It will be more functional and user friendly if we connect user email with the 'comment' feature. So he should be notified.

Adding 'mail' will reduce the need or create confusion to end user for the 'comment' feature. 

Also I would like to highlight that this feature will interfere with the email functionality for the FP organisation 'Sourced by Fightpandemics' as currently I can send email without login to the org but here for the specific post it will ask me to login.
So email is same to connect to same org but if I choose path as:

View profile (sourced by Fightpandemics) -> click on email icon ---> user is able to send email without login


View profile (sourced by Fightpandemics) -> click on email icon existing on the post ---> user is NOT able to send email without login

I would like to suggest, that if we are implementing email feature to connect directly with the author of the post, in that case if we add email functionality at the profile level, rather than to each post. And let comments serve the purpose of communication for the post level only. 

Kindly share your thoughts. 

Also like if we see on linked we have comments, like or share for the posts shared and when you click on profile of the user, then we can directly send the message/ email.

CamilaArias:
@rabbi08 this "Message" option is not for sending emails, but rather for the private messaging feature, it's not the same as comments, those are public, these messages would live in a "chat" between the author and the user, in the "Private Messages" section.

We agree that it would be best to be able to message the author from their profile, but because this is a feature critical for launch, and this solution is more feasible because we're building on top of the comments structure that is already working, this is as far as we know the most effective way to make this work in time. This and the email notifications feature (different from this) aim to close the communication loop between users in time for launch.

Hope that helps!

rabbi08:
> @rabbi08 this "Message" option is not for sending emails, but rather for the private messaging feature, it's not the same as comments, those are public, these messages would live in a "chat" between the author and the user, in the "Private Messages" section.
> 
> We agree that it would be best to be able to message the author from their profile, but because this is a feature critical for launch, and this solution is more feasible because we're building on top of the comments structure that is already working, this is as far as we know the most effective way to make this work in time. This and the email notifications feature (different from this) aim to close the communication loop between users in time for launch.
> 
> Hope that helps!

@CamilaArias 
We do have 'Email functionality at profile level' working currently in FP. All the post from 'Sourced by FightPandemics' have email functionality at profile level. Please see below screen recording for the same. Can we copy and enhance that?

[RPReplay_Final1598674403.zip](https://github.com/FightPandemics/FightPandemics/files/5144771/RPReplay_Final1598674403.zip)

But in case, if that might time extra time than expected then, as discussed we can have this functionality, but can we create this feature as low priority user story? So that it can be tracked in future. Kindly update.

Thanks

rabbi08 (changes requested):
@xreic 
Non-Logged In user -> navigate to Help board -> Click on 'EMAIL' icon -> 
User is navigate to login screen -> PASS
Enter valid username/pwd and login.

Issue-1: User is navigate to POST, ideally compose-email-pop-up,  with associated author email, pre-filled should be displayed

Click on 'EMAIL' icon again to send email to the author of the post
Issue-2: Unable to click, nothing happen when user clicks on 'EMAIL' icon.

Hence, currently user is not able to send email using the email icon.

Thanks.

joshmorel:
Thanks @xreic . I'll take a full review tomorrow but:

1. We do want to add that css class to make it invisible until the whole feature is done as suggested in today's meeting. Maybe `hide-feature`? Define the class globally and we will use it ongoing for similar things.
2. For Google Tag Manager we should add something.  I'd confirm with BI team but I'd suggest we propose  `_MS` and use until we hear otherwise https://github.com/FightPandemics/FightPandemics/blob/1e4944b1db27826ca1accb07798dd622b51577e4/client/src/constants/gtm-tags.js#L35

emtay:
Perhaps I’m confused about this story. I didn’t think it was for email but instead for private message we’re building.

It will need to get connected to right backend once that is built out

manualzuru:
The idea of the envelope icon is to work as a private message.

However, I get your confusion @rabbi08 because we also have an envelope on the individual and organization profile. Maybe we should change that one for a "@" instead as in the future the envelope will might be the message as well.

Obviously, no matter if is a comment or private message, the idea is to still send the user an email notification.

joshmorel:
We could use this one like Instagram does:

image

We just need to replace it within the write comment box with something else.... (maybe just button w "Comment" txt) 

Most social sites seem to use a similar icon to what we are using for comments...

Or @manualzuru suggestion, use the envelope for messaging and on profile page use @ symbol for email.

xreic:
> Thanks @xreic . I'll take a full review tomorrow but:
> 
> 1. We do want to add that css class to make it invisible until the whole feature is done as suggested in today's meeting. Maybe `hide-feature`? Define the class globally and we will use it ongoing for similar things.
> 2. For Google Tag Manager we should add something.  I'd confirm with BI team but I'd suggest we propose  `_MS` and use until we hear otherwise https://github.com/FightPandemics/FightPandemics/blob/1e4944b1db27826ca1accb07798dd622b51577e4/client/src/constants/gtm-tags.js#L35

@joshmorel 
I added the `hide-feature` class and applied it to the button.

joshmorel:
Thanks @xreic I think there's going to have to be some way we make it easy to toggle outside of local development. 

I have some ideas - I'll be back on this soon.

We need to resolve @rabbi08 issue's anyways - but this requires consensus and a new design so this one is a bit on hold for now (@kterry1 - you're going ahead fine with just text for now right?).

joshmorel:
@xreic - about the global class I think it needs to be a wrapper component instead that is hidden on prod but visible in review & dev.

A global class is okay for local dev, but not QA etc.

It's probably simplest just turn it off in prod & staging. The wrapper component probably will do something like - if prod return `null`, otherwise return `<>{children}</>`. 

To check if prod I think for now, `/fightpandemics\.(com|work)$/.test(window.location.hostname)` will be okay. We'd need to add REACT_APP_ env vars otherwise I think to distinguish between review/prod/staging (also an option). 

Once the feature is ready for prod we create a new ticket to remove the wrapper class then it could be tested in staging. Or @rabbi08 @emtay should this hidden features be hidden in prod & staging, or only prod.

rabbi08:
> @xreic - about the global class I think it needs to be a wrapper component instead that is hidden on prod but visible in review & dev.
> 
> A global class is okay for local dev, but not QA etc.
> 
> It's probably simplest just turn it off in prod & staging. The wrapper component probably will do something like - if prod return `null`, otherwise return `<>{children}</>`.
> 
> To check if prod I think for now, `/fightpandemics\.(com|work)$/.test(window.location.hostname)` will be okay. We'd need to add REACT_APP_ env vars otherwise I think to distinguish between review/prod/staging (also an option).
> 
> Once the feature is ready for prod we create a new ticket to remove the wrapper class then it could be tested in staging. Or @rabbi08 @emtay should this hidden features be hidden in prod & staging, or only prod.

As per me I would say lets keep it hidden from both STG and PROD till it is fully functional because of following points:
- STG should be close match with PROD
- To avoid confusion with other features.
- We might need to do multiple changes/fixes, to this feature till it fully functional, in that way, it will be more fast and independent to be build on feature build and merge to STG when ready.

xreic:
@joshmorel 
My bad for misunderstanding I was still thinking about CSS.
I made a simple component that reads the env variables and renders based on that. 
Do you want me to switch it to the URL parsing method?

joshmorel:
Thanks @xreic  the resolution looks good (for dev/staging/prod at least --> see below).

@rabbi08 - that confirms what I thought for staging. So definitely hide there. But what about review? The problem is it will show in other review branches after it's been merged --- there's no easy way to hide/show for a specific review branch I don't think. 

The alternate to this whole approach is having long running "feature" branches with multiple PRs into that feature branch for different parts required for the feature. But there's risks there I think as well in general branch management complexity.

Thoughts @emtay ?

rabbi08:
'Messaging' functionality have already been implemented and working. Hence, closing this as of now, please open a new ticket if this issue re-occur (#1433 is already complemented and closed). Thanks.

PR #1458, opened 2020-08-28
ability to delete user account

### LINKED ISSUE #1148: Let users deactivate/delete their accounts
Task should process account deletion requests where all posts and account-related information is deleted.

Zaydme:

Ability for users to delete their account, this will include all their organizations, posts and comments.
### important notes: 
- when I try to delete a user account from auth0 i get: 
**`Insufficient scope, expected any of: delete:users,delete:current_user`**
so I commented the part that deletes Auth0 profile, until someone with access to Auth0 adds one of the needed scopes,
then all you have to do is un-commenting that line.
- This needs a front-end re-work, as I am not very familiar with React, and now it's now very mobile friendly.
### how does it look: (not so good)
preview
closes #1148 

### DIFF
--- a/backend/lib/endpoints/users.js
+++ b/backend/lib/endpoints/users.js
@@ -12,6 +12,7 @@ const {
 async function routes(app) {
   const Comment = app.mongo.model("Comment");
   const User = app.mongo.model("IndividualUser");
+  const Organisation = app.mongo.model("OrganisationUser");
   const Post = app.mongo.model("Post");
 
   app.get("/current", { preValidation: [app.authenticate] }, async (req) => {
@@ -190,6 +191,26 @@ async function routes(app) {
       return new User(userData).save();
     },
   );
+
+  app.delete("/", { preValidation: [app.authenticate] }, async (req) => {
+    const { userId } = req;
+    // Insufficient scope to delete users on Auth0
+    // expected any of: delete:users,delete:current_user
+    // Please add any of the needed scopes and uncomment the line below
+    // await Auth0.deleteUser(getCookieToken(req), userId);
+    await Promise.all([
+      User.deleteMany({ _id: userId }),
+      Organisation.deleteMany({ ownerId: userId }),
+      Comment.deleteMany({ "author.id": userId }),
+      Post.deleteMany({ "author.id": userId }),
+    ]).catch((deleteErr) => {
+      req.log.error(deleteErr, "Failed deleting user's data");
+      throw app.httpErrors.internalServerError();
+    });
+    return {
+      success: true,
+    };
+  });
 }
 
 module.exports = routes;
--- a/backend/lib/endpoints/users.js
+++ b/backend/lib/endpoints/users.js
@@ -200,9 +200,34 @@ async function routes(app) {
     // await Auth0.deleteUser(getCookieToken(req), userId);
     await Promise.all([
       User.deleteMany({ _id: userId }),
-      Organisation.deleteMany({ ownerId: userId }),
       Comment.deleteMany({ "author.id": userId }),
-      Post.deleteMany({ "author.id": userId }),
+      Post.find({ "author.id": userId }, async (err, allUserPosts) => {
+        if (err) throw err;
+        for (const userPost of allUserPosts) {
+          await Promise.all([
+            Comment.deleteMany({ postId: userPost._id }),
+            userPost.delete(),
+          ]);
+        }
+      }),
+      Organisation.find({ ownerId: userId }, async (err, allUserOrgs) => {
+        if (err) throw err;
+        await allUserOrgs.forEach(async (org) => {
+          await Promise.all([
+            Comment.deleteMany({ "author.id": org._id }),
+            Post.find({ "author.id": org._id }, async (errP, allOrgPosts) => {
+              if (errP) throw errP;
+              for (const orgPost of allOrgPosts) {
+                await Promise.all([
+                  Comment.deleteMany({ postId: orgPost._id }),
+                  orgPost.delete(),
+                ]);
+              }
+            }),
+            org.delete(),
+          ]);
+        });
+      }),
     ]).catch((deleteErr) => {
       req.log.error(deleteErr, "Failed deleting user's data");
       throw app.httpErrors.internalServerError();
--- a/backend/lib/components/Auth0.js
+++ b/backend/lib/components/Auth0.js
@@ -73,6 +73,18 @@ const createUser = async (token, payload) => {
   }
 };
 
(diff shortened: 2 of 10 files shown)


backend/lib/endpoints/users.js
```diff
@@ -190,6 +191,26 @@ async function routes(app) {
       return new User(userData).save();
     },
   );
+
+  app.delete("/", { preValidation: [app.authenticate] }, async (req) => {
+    const { userId } = req;
+    // Insufficient scope to delete users on Auth0
+    // expected any of: delete:users,delete:current_user
+    // Please add any of the needed scopes and uncomment the line below
+    // await Auth0.deleteUser(getCookieToken(req), userId);
```
idjevm, line 200: This seems like something that @joshmorel  will be able to help with, and I am sure it has to do with Auth0.

idjevm (changes requested):
I like this implementation of deleting an org! I think someone else had submitted a version of it, but this seems more neater. Not sure what other's thoughts are so let's see. 

Marking it as Request changes now but it looks really great!

backend/lib/endpoints/users.js
joshmorel: For sure @idjevm + thanks @Zaydme . I'll take a review tomorrow and look into this other aspect.

Many sites give you like some time period to recover if you delete and regret it... But I think for a now a full delete is okay

idjevm: Yeah, but I think this is more clearer as of now though. We would have to use the same design for the organisation and user profile so updates will be needed if we do move forward with this one.

Zaydme: Any updates about the needed auth0 scopes `delete:users,delete:current_user` ?
As documented [here](https://auth0.com/docs/api/management/v2#!/Users/delete_users_by_id), the `
delete /api/v2/users/{id}` endpoint requires the individual user to have **delete:users** scope, but it's a little bit dangerous, so **delete:current_user** should be good.

Zaydme:
@idjevm I updated it and now it will delete:
- User
- All user's posts, AND all comments on those posts (even if by other users, since the post is gone)
- All user's comments
- All user's orgs
- All posts by all the orgs, AND all comments on those posts.
- All comments by all the user's orgs (there is no comments by orgs now, but they might be added soon)

backend/lib/endpoints/users.js
```diff
@@ -190,6 +191,51 @@ async function routes(app) {
...
+      Comment.deleteMany({ "author.id": userId }),
+      Post.find({ "author.id": userId }, async (err, allUserPosts) => {
+        if (err) throw err;
+        for (const userPost of allUserPosts) {
+          await Promise.all([
+            Comment.deleteMany({ postId: userPost._id }),
+            userPost.delete(),
+          ]);
+        }
+      }),
```
Naraujo13, line 212: I think this could be changed to a single delete for all the comments instead of a for loop (opening fewer connections to the DB and avoiding that overhead). 

Instead of iterating through the `postIds` I think a single delete with the `$in` operator will do the trick, something like:
 `Comment.deleteMany({postId: {'$in': userPostIds}})`

backend/lib/endpoints/users.js
```diff
@@ -190,6 +191,51 @@ async function routes(app) {
...
+        await allUserOrgs.forEach(async (org) => {
+          await Promise.all([
+            Comment.deleteMany({ "author.id": org._id }),
+            Post.find({ "author.id": org._id }, async (errP, allOrgPosts) => {
+              if (errP) throw errP;
+              for (const orgPost of allOrgPosts) {
+                await Promise.all([
+                  Comment.deleteMany({ postId: orgPost._id }),
+                  orgPost.delete(),
+                ]);
```
Naraujo13, line 224: Same here as above, could be changed to single delete op

backend/lib/endpoints/users.js
```diff
@@ -190,6 +191,51 @@ async function routes(app) {
...
+
+  app.delete("/", { preValidation: [app.authenticate] }, async (req) => {
+    const { userId } = req;
+    // Insufficient scope to delete users on Auth0
+    // expected any of: delete:users,delete:current_user
+    // Please add any of the needed scopes and uncomment the line below
+    // await Auth0.deleteUser(getCookieToken(req), userId);
+    await Promise.all([
+      User.deleteMany({ _id: userId }),
+      Comment.deleteMany({ "author.id": userId }),
```
Naraujo13, line 203: I was thinking about this line and whether it might leave orphan comments...

Since we can have nested comments, the comments deleted here could be parents of other comments made by a different user. As it stands, this would delete the parent comment (made by the user being deleted), but keep the child comments, leaving them orphan.

Not sure what is expected here, but we could delete those by doing an extra `Comments.deleteMany({parent_id: {'$in': deletedCommentIds}})`

Naraujo13 (changes requested):
Reviewed only the backend and looks good!

Left one small comment regarding performance and one more relevant about the possibility of leaving orphan comments from other users.

backend/lib/endpoints/users.js
Naraujo13: Also, this query above will work only for up to 1 level of nesting, otherwise this would need to be recursive...
If we're sticking to 1 level this should be fine

emtay:
@idjevm @Zaydme - great work! one callout. i believe in future, there is a plan for an organization to have more than one user. i don't believe that is the case today though.

as we build that out. how do we make sure that delete of user doesn't delete organization if organization has more than one user? anything we can do now to protect from this scenario in future? (even if just comment in code?) thanks!

Zaydme:
@emtay 
> as we build that out. how do we make sure that delete of user doesn't delete organization if organization has more than one user? anything we can do now to protect from this scenario in future? (even if just comment in code?) thanks!

yet we have no idea how that is going to be implemented, so we cannot prevent it at this moment, but we can do it once we know how the additional owners are going to be stored in the database.

rabbi08:
@joshmorel  @Zayd
We need to fix the issue if there is hard delete of user from DB to avoid such issue:

I have another observation, though this might be data issue-> as user's account might be deleted, can we check on this. also. please le me know if we need issue in github for this, thanks:
In STG -> while clicking on the profile of following Individuals, Loading error is displayed:
"Failed loading profile, reason: Not Found"
But I can comment or like the post of those users.
RK - Rajbir K
JM - Josh Morel
JS - Jas S

rabbi08:
Currently closing this PR as low priority, in case the issue is observed again, please, feel free to reopen or raise a new ticket for the same. Thanks

PR #1477, opened 2020-08-30
Feature/1474

### LINKED ISSUE #1474: Show 404 page for not found profile/organisation
Noted in #1230 - cc @sheeba1 @jpmitri @vsan626 @emtay 

This would be a good first step until we make a decision about the other issue.

Currently if you visit profile/:id or organisation/:id and it is not found you get this ugly error message:
image

Let's instead redirect them to 404 page:
image

vsan626:
re-routes the user to 404 page if user profile does not exist

### DIFF
--- a/client/src/pages/OrganisationProfile.js
+++ b/client/src/pages/OrganisationProfile.js
@@ -1,7 +1,8 @@
 import { WhiteSpace } from "antd-mobile";
+import { message } from "antd";
 import axios from "axios";
 import React, { useState, useEffect, useContext, useReducer } from "react";
-import { Link } from "react-router-dom";
+import { Link, useHistory } from "react-router-dom";
 
 // ICONS
 import createPost from "assets/icons/create-post.svg";
@@ -91,6 +92,7 @@ const URLS = {
 const getHref = (url) => (url.startsWith("http") ? url : `//${url}`);
 
 const OrganisationProfile = () => {
+  const history = useHistory();
   let url = window.location.pathname.split("/");
   const organisationId = url[url.length - 1];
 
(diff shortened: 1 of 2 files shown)


idjevm (changes requested):
We need to add this for the organisations as well. I will double check once that change is done.

vsan626:
> We need to add this for the organisations as well. I will double check once that change is done.

@idjevm code has been updated for the organisations as well

joshmorel (changes requested):
Thanks @vsan626 

I don't think it was exactly fair to call this a "good first issue" - any time you are dealing with react hooks it's not that case!

Anyways I don't want to re-route I want to show - so we still see the actual URL, e.g.:

image

We can't return the page from `useEffect` but we can set a variable and use that to render it.

E.g.:

```javascript
      } catch (err) {
        if (err?.response?.status === 404) {
          setNotFound(true);
        }
      }
    })();
  }, [setNotFound, pathUserId, userProfileDispatch]);
```

And later

```javascript
  if (notFound) return <NotFoundPage />;
```
 

This will work but we need to use it in many places. There is an Axios interceptor we might be able to use for all 404 errors https://css-tricks.com/stay-dry-using-axios-for-api-requests/? @jpmitri is this what you were referring to in slack about listeners?

But that might be more complicated than it's worth right now.

Currently the most common errors we'll get is:

* 404 - not found
* 500 - server error but this would be because you visited fightpandemics.com/profile/joshs1234 (by accident I guess?) - because it is not a valid Mongoose ObjectId - but I think we'll switch to slugs eventually for more user friendly URLs...

Throwing a 401 on the backend (with `authentication` hook), I don't think that'll work for non logged user when visiting this page as it will cause many headaches - I have a different solution.

rabbi08:
Closing this issue as low priority, please reopen if the priority change

PR #1481, opened 2020-08-31
Feature/1437

### LINKED ISSUE #1437: Email Validation: Extend email to more than 64 (max supported by Auth0) characters and consolidate email validation
As a user, I want email validation to be consistent across my individual and organization profiles and for email length to be 200 characters so that I can be sure my email will fit.

Within this story, all email validation should be in one configuration so that we ensure consistency.

Email validations should be:

Empty field: “Email is required”
Not a valid email address: “Email is invalid”
Over 200 chars: “Email must not exceed 200 characters”
Invalid domain: “Email domain is invalid”

image

AC Character limit:
Given user is adding their email to create an individual profile
When their email is less than 200 characters
Then the system should accept the email and process the profile creation.

Given user is adding their email to create an organization profile
When their email is less than 200 characters
Then the system should accept the email and process the profile creation.

Given user is adding their email to create an individual or organization profile
When their email is longer than 200 characters
Then the system should given an error that there is a max of 200 characters and email is not valid.

Note we considered allowing same email address between indiv and org, but will keep those unique. so one more constraint is that the email is unique across other users (any type of user), so this should continue:
Error for signing up with same email:
image

paulms1:
This PR resolves #1437

Changes Maximum char limit for emails to 200, for both organisations and individual users.


joshmorel (changes requested):
Thanks @paulms1 

This will work but we need to also check in `EditOrganisationAccount.js`. Ideally we'd have some common EmailInput component but we've been inconsistent across our forms and I haven't figured out a good way to do this with react-hook-form (you need to use a [Controller](https://react-hook-form.com/api/#Controller) but I've found this library hard to work to be honest).

Anyways I would suggest:

Add maxLength to `EditOrganisationAccount.js` .

I think we should also add the custom validate function (`validate: (email)`...) since we're looking at this to the org email inputs. If we could add max length to validate function and return a specific message about length (besides "Invalid email" that would be an option as well - but don't worry about that).

paulms1:
Changes addressed - I have added maxLength to EditOrganisationAccount.js and the custom validate function to the create org input (edit org input already had it) I have also added a max length to the validate email function.

joshmorel (changes requested):
To clarify @paulms1  I guess I was suggesting maybe we could check length in the validator function if we could return a custom message - but in such case we wouldn't need the `maxLength` properly.

I don't think that will work easily however. I would say:

* revert `validator.js` (remove length check there). 
* just make sure every email input has maxLength 200 as well as validator function running.

joshmorel:
We should also check on the backend just in case but there's some weird stuff with the current state so instead of fixing here I can create another "good first issue" for backend.

sheeba1:
Please do the email length changes to both create/edit org profiles . Thanks! 
Now, Create Org allows 50 chars while edit Org allows 30.

joshmorel:
. @emtay you wanted 200 right?

emtay:
@sheeba1 good callout that email create and validation should all be the same characters.

@joshmorel yes i suggested 200 characters so that we don't have to touch it again!

paulms1:
Changes made - creation of new common email validator and validator implemented in all instances of email validation across site. 

Validations are:

Empty field: “Email is required”  
Not a valid email address: “Email is invalid”  
Over 200 chars: “Email must not exceed 200 characters”  
Invalid domain: “Email domain is invalid”

joshmorel (approved):
LGTM -thansk @paulms1

There was some formatting issues (mostly extra spaces) that I fixed up as well as an unused function & unnecessary comment.

galeza:
LGTM
@sheeba1 please test :)

galeza (approved):

paulms1:
@joshmorel Thanks!

idjevm (approved):
LGTM!

sheeba1:
For Sign-up, 200 character email is giving "Internal server error" as below:

email used: abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefgh@gmail.com



image

idjevm:
Oops, my bad, I should've actually requested changes. Turns out that Auth0 only supports emails that are not longer than 64chars. So please we should update this. The email should not be longer than 64chars length.

sheeba1:
Tested Create and Edit Org -worked fine .

PR #1501, opened 2020-09-08
Private Message models

### LINKED ISSUE #1432: Messaging: Create Private Message Object in DB
As a user, I want to be able to have a private message between me and the author of a post regarding a post so that I can share private information securely. 

This story should create the DB objects to allow for the "private message."
This can leverage the comment data model, so this story is to extend Backend to allow for next level of "nested comment."

Requirements:
- These should be Identified as "Private".
- These should be linked to the Post.
- These should be linked between Author of Post and User adding the first private.
- These need an indicator for tracking whether the private message has been "Read" by the recipient. This will be used on the front-end to bold or unbold in the inbox: https://github.com/FightPandemics/FightPandemics/issues/1435
- New "private messages" will need to trigger notifications, so will need an identifier separate from comments so notification knows it is a "private message" instead of a comment. (maybe just the "private" attribute?) > notification not in scope for this story.
- All "private messages" (these new objects) should be linked to a given user so that on the front end, we can create an "inbox" showing all threads of messages across all posts to which the user has sent a "private message."

Note: please review with Front End developer necessary to make sure they can leverage for displaying a "Message" button under Post, to display the private message exchange between Post Author and private messenger, and to create an "inbox." https://github.com/FightPandemics/FightPandemics/issues/1435

Spike for reference: https://www.notion.so/fightpandemics/Spike-Architecture-and-Database-Design-for-Direct-messaging-17def264151844278671bae00ec2ebec#a0ba11188cb243e8b6fb3e23b6254fcc

Naraujo13:

- Adds `Thread` mongoose model to represent a conversation between two users
- Adds `Message` mongoose model to represent a message sent by a user in a `Thread`

Closes #1432 

- [X] I have read the latest [**README**](README.md) and understand the development workflow.
- [X] The name of this branch is: **`feature/<branch_name>`**, or **`hotfix/<branch_name>`** if this is a hotfix. The branch is created in a cloned version of this repo, **not a forked repo**.
- [X] This branch is rebased/merged with the **latest staging** (or the **latest production** if it is a hotfix).

### DIFF
--- a/backend/lib/models/Thread.js
+++ b/backend/lib/models/Thread.js
@@ -52,6 +52,9 @@ const threadSchema = new Schema(
 
 // -- Indexes
 /* eslint-disable */
+// Index to ensure unique threads between users
+threadSchema.index({ "participants.id": 1 }, { "unique": true });
+
 // Index to get inbox sorted by most recent
 threadSchema.index({
   "participants.id": 1,
--- a/backend/lib/models/Thread.js
+++ b/backend/lib/models/Thread.js
@@ -14,7 +14,7 @@ const participantSchema = new Schema(
       required: true,
       type: ObjectId,
     },
-    lastAcess: Date,
+    lastAccess: Date,
     name: {
       required: true,
       type: String,
--- a/backend/lib/models/Thread.js
+++ b/backend/lib/models/Thread.js
@@ -1,0 +1,70 @@
+// -- Imports
+const { Schema, model, ObjectId } = require("mongoose");
+
+const CONVERSATION_STATUS_OPTIONS = [
+  "accepted", "archived", "blocked", "pending"
+];
+
+// -- Schema
+const participantSchema = new Schema(
+  {
+    id: {
+      ref: "User",
+      required: true,
+      type: ObjectId,
+    },
+    location: locationSchema,
+    name: {
+      required: true,
+      type: String,
+    },
+    newMessages: {
+      default: false,
+      required: true,
+      type: Boolean
+    },
+    photo: String,
+    type: {
+      enum: USER_TYPES,
+      required: true,
+      trim: true,
+      type: String,
+    },
+  }
+)
+
+const threadSchema = new Schema(
+  {
+    participants: {
+      type: [participantSchema]
+    },
+    status: {
+      default: "accepted",
+      enum: CONVERSATION_STATUS_OPTIONS,
+      lowercase: true,
+      required: true,
+      trim: true,
+      type: String,
+    },
+  },
+  { collection: "threads", timestamps: true },
+);
+
+// -- Indexes
+/* eslint-disable */
+// Index to get inbox sorted by most recent
+threadSchema.index({
+  "participants.id": 1,
+  "status": 1,
+  "updatedAt": -1
+});
+/* eslint-enable */
+
+// -- Model
+const Thread = model("Thread", threadSchema);
+
+module.exports = {
+  CONVERSATION_STATUS_OPTIONS,
+  model: Thread,
+  schema: threadSchema,
+};
--- a/backend/lib/models/Thread.js
+++ b/backend/lib/models/Thread.js
@@ -1,5 +1,6 @@
 // -- Imports
 const { Schema, model, ObjectId } = require("mongoose");
+const { USER_TYPES } = require("./Author");
 
 const CONVERSATION_STATUS_OPTIONS = [
   "accepted", "archived", "blocked", "pending"
@@ -13,7 +14,6 @@ const participantSchema = new Schema(
       required: true,
       type: ObjectId,
     },
-    location: locationSchema,
     name: {
       required: true,
       type: String,
--- a/backend/lib/models/Thread.js
+++ b/backend/lib/models/Thread.js
@@ -14,6 +14,7 @@ const participantSchema = new Schema(
       required: true,
       type: ObjectId,
     },
+    lastAcess: Date,
     name: {
       required: true,
       type: String,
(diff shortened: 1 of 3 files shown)


joaofnds (approved):

backend/lib/models/Thread.js
```diff
@@ -0,0 +1,70 @@
...
+
+// -- Schema
+const participantSchema = new Schema(
+  {
+    id: {
+      ref: "User",
+      required: true,
+      type: ObjectId,
+    },
+    location: locationSchema,
```
joshmorel: Looking at the designs I don't see us needing participant's `location` and I doubt we will ever need to reference it within the messaging feature itself. So I would omit here.

joshmorel:
Thanks @Naraujo13 . So this does come closest to the LinkedIn approach - you can "Send" a post to someone which either starts a thread/conversation, or adds to that thread if it already exists. The uniqueness of the thread is based on the participants (so no multiples between the same participants).

What we need to figure out - @kterry1 @lmpeixoto (cc @emtay ) - is on the routes and what I'm particularly unsure of is the logic for creating / appending to a thread from the modal. Originally we'd be creating a new thread always a POST (create thread) but in this case we need to:

1. Query for the thread based the participant's id. Maybe GET /threads?participantIds=5ef8d4b77211201600aaec3e%265f5645c59a4a0706dea6f47f ?
2. If it doesn't exist, create the thread. POST /threads 
3. Then create the message (append to thread), e.g. POST /threads/:threadId/message

3 is the straightforward part. 1/2 I'm not sure about. And if that's the right way then do we call 2 upon Submit?

lmpeixoto:
Thanks for your great job @Naraujo13! I agree with @joshmorel about the location. I can't see utility right now, but I'd like to hear your opinion @Naraujo13.
About the creation of a new thread @joshmorel, I think that's the way to go as the thread is supposed to be unique.

emtay:
@joshmorel good callouts. @Naraujo13

The threadIds will be the combination of the 2 users' ids? Is there a way to ensure that the "lower" number is always first so it will be a unique key between 2 users?

What would happen if you always try to create a new thread even if one is already created- will it overwrite it or can it merge?

Naraujo13:
> Thanks @Naraujo13 . So this does come closest to the LinkedIn approach - you can "Send" a post to someone which either starts a thread/conversation, or adds to that thread if it already exists. The uniqueness of the thread is based on the participants (so no multiples between the same participants).
> 
> What we need to figure out - @kterry1 @lmpeixoto (cc @emtay ) - is on the routes and what I'm particularly unsure of is the logic for creating / appending to a thread from the modal. Originally we'd be creating a new thread always a POST (create thread) but in this case we need to:
> 
> 1. Query for the thread based the participant's id. Maybe GET /threads?participantIds=5ef8d4b77211201600aaec3e%265f5645c59a4a0706dea6f47f ?
> 2. If it doesn't exist, create the thread. POST /threads
> 3. Then create the message (append to thread), e.g. POST /threads/:threadId/message
> 
> 3 is the straightforward part. 1/2 I'm not sure about. And if that's the right way then do we call 2 upon Submit?


I've added as a reflex from the author doc we already use at the post and comment model, but I agree that for the `Thread` and message models it makes more sense to remove it since we don't need to display the user's location anywhere.

I'll remove that and update it here. 

The embedded participant docs at `Thread` improve our read performance, but while reading the reviews on the PRs I've started to question if we need the embedded author doc at the `Message` collection, maybe just the user id would be better there.
The messages will always be retrieved in a context where we queried the Thread before, so we will already have all the user's info and just the id would be enough to know what to display. That would avoid copying the user info for every sent message (a possible problem if we want to keep that in sync and storage wise). 

What do you guys think? @joshmorel @lmpeixoto @gullwings13 @emtay

backend/lib/models/Thread.js
Naraujo13: I agree, I think we can omit this for all the user references here.

Naraujo13:
> @joshmorel good callouts. @Naraujo13
> 
> The threadIds will be the combination of the 2 users' ids? Is there a way to ensure that the "lower" number is always first so it will be a unique key between 2 users?
> 
> What would happen if you always try to create a new thread even if one is already created- will it overwrite it or can it merge?

@emtay The `threadId` is actually the `_id` in the `Thread` model, which is an `ObjectId` and not a combination of the users' ids.
We can enforce uniqueness  (one single thread for each combination of participants) by adding a unique index in the `participants.id` array. That at DB level would avoid the creation of a new document (I think it will be a write error), but I think it is best to handle this at the application level with the logic Josh mentioned.

We can add an index with `participants._id`, so the first step you mentioned @joshmorel will be a simple/performant check.

emtay:
@Naraujo13 - makes sense! 

Proposal by @joshmorel seemed good if doable!

Great work guys!

joshmorel:
> The embedded participant docs at Thread improve our read performance, but while reading the reviews on the PRs I've started to question if we need the embedded author doc at the Message collection, maybe just the user id would be better there.
The messages will always be retrieved in a context where we queried the Thread before, so we will already have all the user's info and just the id would be enough to know what to display. That would avoid copying the user info for every sent message (a possible problem if we want to keep that in sync and storage wise).

:100: I think I was going to ask about this. We only need the id and the FE can reference the thread participants.

lmpeixoto (approved):
IMHO it's done @Naraujo13, nice work!

joshmorel:
Okay I think we can merge this and start work on other issue (routes). If we need to iterate on model in that its okay

joshmorel (approved):

PR #1504, opened 2020-09-09
Feature/1462

### LINKED ISSUE #1462: Improve Create Account experience Indiv/Org
### First Issue
_Mainly on mobile_

User hasn't selected an address, and nothing says it's mandatory (and because it doesn't look like it is, they don't choose one, because they might not want to give an address).
They only need to agree to Privacy Policy and T&C for the "Create Profile" button to become active. They click it and it turns white. Nothing happens, no feedback.

image

Button turns light-blue again when clicked, now inactive.
Unless they scroll up, they won't see the error message below address selection.

### Second Issue
_Mobile/desktop_

User is trying to sign up with an email that's already in use.
After realising they have to enter an address (issue 1), they click "Create Profile". Nothing happens again. No feedback again. The button is now outlined and whenever it's clicked again it says "Creating Profile..." for a second.

image

Only if they scroll up, they'll see the error message, at the top of the screen, saying their email needs to be unique. No prompts to do anything. Also, instead of checking for both address value an email uniqueness, we check one first and then the other, we should at least tell them everything that's wrong, especially when email being wrong is more important than not selecting an address.

### Suggestions

Add a required * hint for address to minimize friction. We do it for other fields so it's odd we don't do it for this.

Display any errors on top of the "Create Profile" button, which is where the user is looking. Feedback is important and needs to be instantly visible.

--- This will be a different ticket---
When we retrieve the email from the account they're trying to sign up with, check that is unique.
- If it is -> create account
- If it's not, tell them there's an account with that email -> sign in

Don't let them go through a process deemed to fail. Also if we do this, we don't need to show an error for expecting unique email because that situation would no longer be possible.

cjohlmacher:

Closes #1462 
- By disabling the Submit button if Address is not filled in AND
- By adding an asterisk * next to required fields (Address, Privacy Policy, Terms and Conditions) AND
- By moving error message displaying errors that happen onSubmit to appear above the submit button

- [X ] I have checked that no one else is working on similar changes.
- [X ] I have read the latest [**README**](README.md) and understand the development workflow.
- [X ] The name of this branch is: **`feature/<branch_name>`**, or **`hotfix/<branch_name>`** if this is a hotfix. The branch is created in a cloned version of this repo, **not a forked repo**.
- [X ] This branch is rebased/merged with the **latest staging** (or the **latest production** if it is a hotfix).
- [ X] There are no merge conflicts.
- [X ] There are no console warnings and errors.


joshmorel (approved):
LGTM

idjevm (approved):
LGTM!

idjevm (changes requested):
@joshmorel Should we close this one? Loooks like it has become outdated as @jyclam changed a few things when implementing lokalise... Will mark it as Request changes for now, but we could close it or do it differently ?

rabbi08:
The 'Submit' button is not enable until user add all the required details. Hence, closing this issue, please open a new ticket, if issue is observed again. Thanks

PR #1544, opened 2020-09-23
Cypress test for Help Board Filter Panel

jessica-lau:

Automated cypress test for Filter side panel on the Help Board. This test includes all filtering aspects within the panel.


TimeaKovacs86 (commented):
Hi Jessica! @jessica-lau 
I have reviewed the code, I have some remarks and suggestions: 
 - postCategory function wouldn't work for the other situations where we are testing the visibility and clickability (where it's relevant)? It might be more consistent 
 - it might need a different name for the function like elementIsVisibleAndClickable
 
 - In the case of Filtering posts by provider, type and location on Help Board,
   if the context is "only" before and not beforeEach like in the previous tests, wouldn't it make too many dipendencies for the following tests? 
   Wouldn't it cause lots of false negative tests if something is wrong at the first step?

What do you think about these?

jessica-lau:
> Hi Jessica! @jessica-lau
> I have reviewed the code, I have some remarks and suggestions:
> 
> * postCategory function wouldn't work for the other situations where we are testing the visibility and clickability (where it's relevant)? It might be more consistent
> * it might need a different name for the function like elementIsVisibleAndClickable
> * In the case of Filtering posts by provider, type and location on Help Board,
>   if the context is "only" before and not beforeEach like in the previous tests, wouldn't it make too many dipendencies for the following tests?
>   Wouldn't it cause lots of false negative tests if something is wrong at the first step?
> 
> What do you think about these?

@TimeaKovacs86 good point on the first two suggestions! I went ahead and edited the function name and reused it for other tests in this spec that are testing the same situation. 

As for using the before() vs beforeEach() hook, I used before() to visit the page and click on the filters button once since the following tests will not work unless the user clicks the filters button to open the second panel. 

There are two other ways I can think of:
1. Use beforeEach() but within each test case under "Filtering posts by provider, type and location on Help Board" I would need to add the function seeAndClickElement(helpBoardFilter.getFilterButton()); in each of the test cases so that it triggers the second panel to open and run the tests. I was trying to avoid repeating the function multiple times so that's why I opted to use before().

2. Add the function of clicking the filters button right after the visit method and move the test case for checking and clicking the filters button to the first test suite.
 
`before(() => {
            helpBoardFilter.visit();
            seeAndClickElement(helpBoardFilter.getFilterButton());
        });`

I went ahead and tweak the code to reflect the 2nd option. Take a look and let me know what you think.

TimeaKovacs86 (commented):
The new function is great, looks good. :D I can see now what confused me with the before and beforeEach hooks... absolutely makes sense. The code you wrote works perfectly,
 the only thing I am concerned about that, this way in every step in the work flow is a separate test and there are dependencies, so in case future debugging if something goes wrong,
 it might be difficult to find the proper steps (what depends on what). What I want to say is, if I am understanding good, that these tests supposed to cover end-to-end actions then. 
 What do you think about to put together the separate tests in one, when you are testing one part of the filter?
 I mean for example: This would be a scenario under the name something general like: 'User can add his/her location by typing it in or have the option to share it' =  'Location filter is expandable and collapsible' + 'Type in location field and see the list of locations' + 'Location subtext is visible' + 'Share location link is visible and clickable' together would cover the workflow of the location filter from end-to-end (using Cypress as a end-to-end tool). I will PM you with my questions or ideas.

jessica-lau:
@TimeaKovacs86 I see where you're getting at. I went ahead and grouped the test cases as one where applicable so that they can be run independently from each other if needed. Thank you!

galeza:
Hi @jessica-lau 

I run tests on docker-compose. All pass:
Screen Shot 2020-10-15 at 11 28 20
I like how tests were grouped.

i have a question about cookieBanner in the POM HelpBoardFilter. Maybe if it  independent from the Help Board, we could move it to the separate POM and then be able to use it in other tests?

TimeaKovacs86 (approved):

jessica-lau:
> Hi @jessica-lau
> 
> I run tests on docker-compose. All pass:
> Screen Shot 2020-10-15 at 11 28 20
> I like how tests were grouped.
> 
> i have a question about cookieBanner in the POM HelpBoardFilter. Maybe if it independent from the Help Board, we could move it to the separate POM and then be able to use it in other tests?

Hey @galeza, I'm thinking about going over all the tests that have been written so far on the weekend and create files where repetitive things are being used across and putting them in their respective central file(s). I'll also move the error messages to fixtures. This will clean up the tests and reduce repetition but for now let's approve this PR if everything else looks good and I'll open another PR to make the changes I just mentioned. Thank you!

galeza (approved):

galeza:
@jessica-lau 
Good idea. Approved

galeza:
@idjevm 

Please merge :)

idjevm (approved):
Alrighty @galeza ! Merging now! Thanks for all the work Jessica! :)

PR #1566, opened 2020-10-01
Feature/1437 use validator is email

paulms1:
This PR resolves feature/1437-use-validator-isEmail

Implements the same validation rules as Auth0. Valid email format is:

(64max)@+(63max)+‘.’+(tld)

Valid top level domains should be any that appear in this list:

https://data.iana.org/TLD/tlds-alpha-by-domain.txt


client/src/utils/validators.js
```diff
@@ -11,16 +12,14 @@ const validateTopLevelDomain = (string) => {
...
-    errorMessage = "emailInvalid";
-  } else if (!validateTopLevelDomain(emailInput)) {
-    errorMessage = "emailDomainInvalid";
+    errorMessage = "Email address is required";
+  } else if (email.length > 254) {
+    errorMessage = "Email address must not exceed 254 characters";
+  } else if (!_isEmail(email)) {
+    errorMessage = "Email address is invalid";
+  } else if (!validateTopLevelDomain(email)) {
+    errorMessage = "Email address is invalid";
```
joshmorel:
```suggestion
  } else if (!_isEmail(email) || !validateTopLevelDomain(email)) {
    errorMessage = "Email address is invalid";
```

Let's combine this since the message is the same

joshmorel (changes requested):
One minor change but we need to also ask @jyclam about how do we integrate translation properly.

client/src/utils/validators.js
joshmorel: Okay I resolved this but will ask a few questions.

client/src/locales/translations/es_ES.json
```diff
@@ -224,7 +224,7 @@
         "linkedin": "URL de LinkedIn",
         "website": "Página web"
       },
-      "emailMaxLength": "La dirección de correo electrónico no debe exceder los 200 caracteres",
+      "emailMaxLength": "La dirección de correo electrónico no debe exceder los 254 caracteres",
```
jyclam, line 227: Just a note, there's no need to touch these other translation files manually. These changes won't be captured in Lokalise and the next time these languages are exported, the previous translation (200 characters) will show up again. 

Only the changes in `en_US` will get picked up by Lokalise and from there, translators will have to change the message of all the other languages. And then we can export the changes. That's the proper way to do it and there will be a lag / desync between the languages. But technically we can make the changes here and on Lokalise so there isn't a desync. 

So keeping these changes are fine, just wanted to give you a heads up!

We should have implemented this differently so you can inject the number of characters into the translation.

client/src/utils/validators.js
```diff
@@ -11,15 +12,13 @@ const validateTopLevelDomain = (string) => {
 
 export const validateEmail = (email) => {
   let errorMessage = "";
-  const emailInput = String(email).toLowerCase();
-  const re = /^(([^<>()\]\\.,;:\s@"]+(\.[^<>()\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;
   if (!email) {
     errorMessage = "emailRequired";
-  } else if (email.length > 200) {
+  } else if (email.length > 254) {
```
joshmorel: Now that I look more into this validator.js is already checking length so I think we can remove that and also the related email keys. So we only have 2 keys things to translate for each language - emailRequired & emailInvalid (I added back `emailDomainInvalid` but we can use the same message. 

Let me know if you agree @paulms1 

https://github.com/validatorjs/validator.js/blob/685c3d2edef67d68c27193d28db84d08c0f4534a/lib/isEmail.js#L48

joshmorel:
> Just a note, there's no need to touch these other translation files manually. These changes won't be captured in Lokalise and the next time these languages are exported, the previous translation (200 characters) will show up again.

Ah thanks @jyclam makes total sense.

In any case I think we will likely limit to only 2 error messages for frontend (required/invalid). So in this case if we remove them from `en_US` only would that be propagated to other languages? If not no big deal as long as it's not added to net new languages as that's extra work to be avoided.

jyclam:
Technically, yes removing keys in `en_US` should trigger a clean up and delete the key and translations on Lokalise but I've found it to be a bit hit or miss, sometimes it picks up changes as I expect them to, sometimes it doesn't.

joshmorel (approved):
LGTM! Thanks @paulms1 we'll get this merged

PR #1575, opened 2020-10-02
Lokalise: Translations update - adding German

jyclam:
This pull request was initiated by Lokalise (user Do Hee Kim) at 2020-10-03 00:18:00


joshmorel:
@jyclam do you know why the translation file ended up at root (`locale/translations/de.json`) instead of within client/ like the rest (e.g. `client/src/locales/translations/pl.json`)

I guess we also need to uncomment `client/src/locales/index.js`

joshmorel (approved):
Thanks @doheekim93 

@jyclam I went ahead and fixed the file location but it would be good to know why it ended up at root with the auto PR.

doheekim93:
@joshmorel that was my b -- I wrote "locale" not "locales". I'm to blame!!

idjevm (approved):
LGTM! Are we doing to have a JSON file for every language that our platform uses?

jyclam:
@joshmorel good catch! Culprit has confessed. 

I see that you've enabled the language and added it to the menu on this PR too. We were actually thinking it might be better to do it separately because:

1. Current workflow only builds on `feature/` or `hotfix/` branches. So QA can't do a quick click through unless we change our flow.
2. Subjective, but I think it's cleaner to keep the Lokalise PR strictly to make changes to the language JSONs

jyclam:
@idjevm 
> LGTM! Are we doing to have a JSON file for every language that our platform uses?

Yes, every language will have its own JSON that will be managed by Lokalise

joshmorel:
Ah okay should have asked @jyclam . Going forward this makes total sense. 

Should I undo it for this one? I did test it locally and it seems to be working fine like the other options. You decide.

jyclam:
@joshmorel I am going to close this PR and initiate a new one from Lokalise just to make sure that it was only the typo that threw it off. Sound good?

PR #1598, opened 2020-10-06
#1594 - add email and password validator in backend (reuse frontend logic)

### LINKED ISSUE #1594: Sign Up/Sign In Backend Server Validation Messages and Status Codes
When running backend tests on staging regarding signup/sign in there are:

1. return of some validation messages from the server that need to be rephrased 
2. incorrect HTTP status code/error being returned from server

The below tests are based on the valid email format (64max)@+(63max)+‘.’+(tld) within the max char limit of 254 implemented by Auth0.

**Scenario 1: _**Sign Up**_**
When I enter more than 64 characters before '@' for the email, the error message returns 500 internal server error when it should return an "Email address is invalid" related message (400 Bad Request status code) .

image

**Scenario 2: **_Sign Up & Sign In_****
Password fields left empty on signup/sign in triggers 500 internal server error when it should return a 400 Bad Request code with message "Password is required".

image

**Scenario 3: **_Sign Up & Sign In_****
There are a couple of tests pertaining to when an invalid email format is entered and the server returns an error message of 'body.email should match format "email"'. Such messages in this format should be rephrased to "Email address is invalid" or “Email address is required” instead. Below are some examples:

Domain exceeding 63 characters on signup/sign in -
image

Email field left empty on signup/sign in -
image

Unspecified email domain on signup/sign in -
image

**Scenario 4:** 

While keeping the local part and domain part of an email within their character limits then using an invalid tld (ones that don't exist on the [tld list](https://data.iana.org/TLD/tlds-alpha-by-domain.txt)), I received two different outcomes.

- 62 char for local + @ + 5 char for domain + . + 125 char for tld = 194 total. Received error message "body.email should match format \"email\"

image

- 62 char for local + @ + 5 char for domain + . + 4 char for tld = 73 total. Received no error message.

image

**Scenario 5:** 

After running the login test continuously 10+ times, the two tests in the screenshot failed because they're expecting a 429 status error (too many requests) instead of a 401 status error (unauthorized). The message we expect the server to return is 401 unauthorized user. Apparently, the 429 error appears due to rate limits being set to protect the server from intentional abuse. X-RateLimit-* headers can be set with information about the limits. How can we set this and what limits can we use?

https://blog.bearer.sh/error-429-too-many-requests-what-to-do-when-youve-been-rate-limited/

image

ddveloper:
Apply the front-end email and password validation code to the backend  
**(Need a thorough test before merging, This draft PR aims for code review)**

Fixes #1594 


backend/lib/utils.js
```diff
@@ -17,8 +19,37 @@ const bool = (env) => env == "true";
...
-const emailRegEx = /^\w+([.-]?\w+)*@\w+([.-]?\w+)*(\.\w{2,3})+$/;
-const isValidEmail = (email) => emailRegEx.test(email);
+// email validation code from front-end 'src\utils\validators'
+const validateTopLevelDomain = (string) => {
+  for (const tld of tlds) {
+    if (string.endsWith("." + tld)) {
+      return true;
+    }
+  }
+  return false;
```
joaofnds: What do you think of applying the [Replace Loop with Pipeline Refactoring](https://refactoring.com/catalog/replaceLoopWithPipeline.html)?

```suggestion
tlds.some(tld => string.endsWith("." + tld))
```

joaofnds (approved):

backend/lib/utils.js
```diff
@@ -17,8 +19,37 @@ const bool = (env) => env == "true";
 
 const getCookieToken = (req) => req.cookies.token;
 
-const emailRegEx = /^\w+([.-]?\w+)*@\w+([.-]?\w+)*(\.\w{2,3})+$/;
-const isValidEmail = (email) => emailRegEx.test(email);
+// email validation code from front-end 'src\utils\validators'
+const validateTopLevelDomain = (string) => {
```
joaofnds: Maybe change the name to have the prefix `isValid`, to match the other validators

```suggestion
const isValidTopLevelDomain = (string) => {
```

galeza:
Hi @ddveloper 
Is this ready for a review? Let me know.

FYI @jessica-lau

ddveloper:
Yes @galeza the error codes were fixed and confirmed with @jessica-lau , we also agreed on using the translation key as the error strings, I think this is good for now, if there is further Mocha test failure please let me know, I will get back to it then.

jessica-lau:
@ddveloper does this PR also fix the rate limit mentioned in scenario 5 of PR#[1594 ](https://github.com/FightPandemics/FightPandemics/issues/1594)or that's still being looked into? If not, should I create a separate issue for that?

ddveloper:
yes please @galeza , this does not fix the rate issue, please create a separate issue

galeza:
@jessica-lau 
I think that we will have to discuss the rate limit with dev. We can set it in our tests (mocha/apiHelper.js). The question is - is it ok/acceptable and what this limit should be?

jessica-lau:
@galeza @ddveloper ok I'll create a separate issue for rate limit.

ddveloper:
Thanks @jessica-lau , I will follow that one also

mannykary (approved):
LGTM

backend/lib/utils.js
```diff
@@ -17,8 +19,32 @@ const bool = (env) => env == "true";
...
+
+const isValidPassword = (password) => {
+  const results = [
+    containsLowerCase(password),
+    containsUpperCase(password),
+    containsNumber(password),
+    containsSpecialChars(password),
+  ];
+
+  return results.filter((res) => res).length >= 3;
```
idjevm: could we rename the `results` to something else, also the flag `res` to something else? It is not as straight forward. Also do we want the password only to meet 3 out of 4 of those validators?

backend/package.json
```diff
@@ -50,7 +50,9 @@
     "mongoose": "^5.9.13",
     "mongoose-unique-validator": "^2.0.3",
     "node-cache": "^5.1.0",
-    "pino-papertrail": "^2.0.0"
+    "pino-papertrail": "^2.0.0",
+    "tlds": "^1.207.0",
+    "validator": "^13.1.17"
```
idjevm, line 57: Mmmm, if two new dependencies were added, we should update the package-lock.json... any reason why was not that committed?

idjevm (changes requested):

backend/package.json
ddveloper: Thanks JV, I will commit the package-lock.json

backend/lib/utils.js
ddveloper: how about `validCombo` and `valid`, while this code was copied from our front end code, may need the same update there. I am not sure how strict we want the password to be, this applies to both our F/E and B/E, @joshmorel has a ticket talking about if we should keep validation on both side #1641

jessica-lau:
hey @ddveloper @idjevm, just wanted to check in on this. Is this good to go or still waiting on some changes to be made?

backend/lib/utils.js
ddveloper: Hi @idjevm I renamed it to be `checks` and `check`, let me know if that works : )

backend/lib/utils.js
```diff
@@ -17,8 +19,32 @@ const bool = (env) => env == "true";
...
+
+const isValidPassword = (password) => {
+  const checks = [
+    containsLowerCase(password),
+    containsUpperCase(password),
+    containsNumber(password),
+    containsSpecialChars(password),
+  ];
+
+  return checks.filter((check) => check).length >= 3;
```
idjevm:
```suggestion
  const passwordChecks = [
    containsLowerCase(password),
    containsUpperCase(password),
    containsNumber(password),
    containsSpecialChars(password),
  ];

  return passwordChecks.filter(isValid => isValid).length >= 3;
```

idjevm (changes requested):
Let me know if you think this is a good suggestion for naming. Then it should be good to go ~

backend/lib/utils.js
ddveloper: Thanks JV, that's better, I commit it : )

jessica-lau:
hey @idjevm just checking if this one's good to be merged now? :)

idjevm (approved):
LGTM! 

Should we merge it @ddveloper ?

ddveloper:
> LGTM!
> 
> Should we merge it @ddveloper ?

I will double-confirm with @Zaydme to see if there will be potential conflicts, please take a look @Zaydme

idjevm:
@ddveloper With which PR you think there might be any conflicts? I could double check too. So far it looks good to go but maybe I am unaware of another PR that has changes related to this as well.

Zaydme:
@ddveloper @idjevm I think you are talking about #1767? I think there shouldn't be any conflicts. LGTM

jessica-lau:
Thanks everyone!

PR #1600, opened 2020-10-06
#786 - Fix eslint/prettier/husky config

### LINKED ISSUE #786: Run husky in docker & local

camba95:

Resolves https://github.com/FightPandemics/FightPandemics/issues/786

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,8 @@ If, at any point, you are having trouble setting up the app, please use Slack's 
 
 If, for some reason, you are unable to use Docker, you can still set up the app locally. Note that you can set up each of these three services separately, but some functionality may not work. For example, if only the client is running, none of the calls to the backend will work.
 
+**Note:** running `npm install` from project root, will install sub projects dependencies (client and backend).
+
 #### MongoDB
 
 Follow the MongoDB [installation instructions](https://docs.mongodb.com/manual/installation/) for your operating system.
@@ -207,3 +209,11 @@ When a pull request is merged to staging, it will automatically be deployed to t
 ### Production
 
 When a pull request is merged to production, it will automatically be deployed to the production environment. You can view the build logs in [GitHub Actions](https://github.com/FightPandemics/FightPandemics/actions). After the build successfully completes, wait a few minutes for the changes to be reflected, and then access the production app at https://fightpandemics.com.
+
+## Know Issues
+
+### VSCode Source Control, Nvm, Husky Issue
+
+If you prefer to use nvm for managing multiples Node versions, and you use control source integreated VSCode tool to commit changes. Husky, one of our our dependencies, may raise an error related to Node version not being met, as Husky does not use nvm automatically. Please see fllow this link for the fix and more information
+
+https://github.com/typicode/husky/issues/438#issuecomment-700647847
--- a/package.json
+++ b/package.json
@@ -5,6 +5,9 @@
   "version": "1.0.0",
   "description": "Help to fight global pandemics",
   "scripts": {
+    "install:client": "cd client && npm install",
+    "install:backend": "cd backend && npm install",
+    "postinstall": "npm-run-all install:*",
     "precommit:client": "cd client && npm run precommit",
     "precommit:backend": "cd backend && npm run precommit",
     "precommit": "npm-run-all precommit:*"
--- a/package.json
+++ b/package.json
@@ -1,0 +1,25 @@
+{
+  "name": "fightpandemics",
+  "author": "Fight Pandemics",
+  "license": "ISC",
+  "version": "1.0.0",
+  "description": "Help to fight global pandemics",
+  "scripts": {
+    "precommit:client": "cd client && npm run precommit",
+    "precommit:backend": "cd backend && npm run precommit",
+    "precommit": "npm-run-all precommit:*"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git@github.com:FightPandemics/FightPandemics.git"
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "npm run precommit"
+    }
+  },
+  "devDependencies": {
+    "husky": "^4.3.0",
+    "npm-run-all": "^4.1.5"
+  }
+}
--- a/backend/package-lock.json
+++ b/backend/package-lock.json
@@ -1074,12 +1074,6 @@
       "integrity": "sha512-NOKm8xhkzAjzFx8B2v5OAHT+u5pRQc2UCa2Vq9jYL/31o2wi9mxBA7LIFs3sV5VSC49z6pEhfbMULvShKj26WA==",
       "dev": true
     },
-    "compare-versions": {
-      "version": "3.6.0",
-      "resolved": "https://registry.npmjs.org/compare-versions/-/compare-versions-3.6.0.tgz",
-      "integrity": "sha512-W6Af2Iw1z4CB7q4uU4hv646dW9GQuBM+YpC0UvUCWSD8w90SJjp+ujJuXaEMtAXBtSqGfMPuFOVn4/+FlaqfBA==",
-      "dev": true
-    },
     "component-emitter": {
       "version": "1.3.0",
       "resolved": "https://registry.npmjs.org/component-emitter/-/component-emitter-1.3.0.tgz",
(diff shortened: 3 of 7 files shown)


joshmorel:
Thanks @camba95 . I'll work through a few scenarios - both w docker & w/o docker and provide by feedback soon.

README.md
```diff
@@ -39,6 +39,8 @@ If, at any point, you are having trouble setting up the app, please use Slack's
 
 If, for some reason, you are unable to use Docker, you can still set up the app locally. Note that you can set up each of these three services separately, but some functionality may not work. For example, if only the client is running, none of the calls to the backend will work.
 
+**Note:** running `npm install` from project root, will install sub projects dependencies (client and backend).
```
joshmorel, line 42: If you follow typical docker set-up, `npm install` will be run inside containers and then we have `node_modules` owned by root. If I want to run `npm install` at root later to run husky I will get:

image

We should have a `npm install:all` command for local set-up that does this, but `npm install` at root only installs root stuff.

joshmorel (changes requested):
Thanks @camba95 

Now that I'm actually looking at this https://github.com/sudo-suhas/lint-staged-multi-pkg#manually-call-precommit-in-each-package - I realize there is one major limitation with calling lint-staged in each sub-module if we are using Docker.dev files as is.

We won't have easy eslint/prettier integration with the code editor because eslint is still in 

I was going to say move eslint/prettier/lint-staged to root but I don't think there's any way for eslint to work if it's not part of the same package.json as the actual project with all it's dependencies.

What I think we need to do, is not run `npm install` in the `Dockerfile` but instead as a container command as described here maybe - https://stackoverflow.com/a/45894288/4582054

Thoughts also @mannykary . Or maybe I'm missing something?

package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "fightpandemics",
+  "author": "Fight Pandemics",
+  "license": "ISC",
+  "version": "1.0.0",
+  "description": "Help to fight global pandemics",
+  "scripts": {
+    "install:client": "cd client && npm install",
```
mannykary, line 8: Not sure if it will work in Windows, but it's probably better to use `pushd`, followed by `popd` to return to the root directory: `pushd client && npm install && popd`. Alternatively, I think this would work in Windows as well as *nix systems: `cd client && npm install && cd ..`

mannykary:
> Thanks @camba95
> 
> Now that I'm actually looking at this https://github.com/sudo-suhas/lint-staged-multi-pkg#manually-call-precommit-in-each-package - I realize there is one major limitation with calling lint-staged in each sub-module if we are using Docker.dev files as is.
> 
> We won't have easy eslint/prettier integration with the code editor because eslint is still in
> 
> I was going to say move eslint/prettier/lint-staged to root but I don't think there's any way for eslint to work if it's not part of the same package.json as the actual project with all it's dependencies.
> 
> What I think we need to do, is not run `npm install` in the `Dockerfile` but instead as a container command as described here maybe - https://stackoverflow.com/a/45894288/4582054
> 
> Thoughts also @mannykary . Or maybe I'm missing something?

I'm trying to follow the accepted StackOverflow answer
* They are using `yarn` but I suppose that's specific to their use case, so I don't think we need to use `yarn` as well.
* They are not mounting an anonymous volume for `node_modules` which I guess is because they need the dependencies to be accessible on the host. I think that's okay, as long as developers are not mixing up dependency installs between the host and client. But I can foresee an issue when a developer is installing a new dependency by running `npm install --save <package_name>` on the host rather than in the container. So the proper way to install new dependencies would be `docker-compose run client npm install --save <package_name`. We should probably document that, but in my experience I think a lot of developers will either not see that or they may forget, so they will end up installing on the host anyway.
* They want to install the dependencies from within the container, hence the command to run `docker-compose run react yarn` prior to `docker-compose up -d`. I think the other way to do this is to add an override `command` to `backend-service` and `client` containers to the `docker-compose.yml` file:
```
...
backend-service:
    build:
      context: backend
      dockerfile: Dockerfile.dev
      args:
        NODE_ENV: dev
    command: bash -c "npm install && npm run dev"
...
  client:
    build:
      context: client
      dockerfile: Dockerfile.dev
    restart: always
    command: bash -c "npm install && npx concurrently npm:start"
...
```

Now, every time `docker-compose up` is run, the dependencies are installed. I think a similar effect can be achieved by updating the `Dockerfile.dev` themselves to add the `npm install` command, so then it won't be necessary to update `docker-compose.yml`

Now going back to my second point, maybe it would be a good idea to clean out the `node_modules` folder before each `npm install` so that there is no mixing of dependency installs. The downside is that it would be a lot slower to bring up containers. It would be nice if we could figure out a way to do this only if a new dependency is added to the `package.json`.

In any case, whatever route we decide to take, we will need to test this all out and see whether it would make life easier or harder for developers.

joshmorel:
@mannykary @camba95 I realize now another issue with this comment.

> I was going to say move eslint/prettier/lint-staged to root but I don't think there's any way for eslint to work if it's not part of the same package.json as the actual project with all it's dependencies.

For some editors, VS Code at least, if it's not at the root the eslint/prettier extensions won't work unless the node packages are installed globally. That's okay with prettier, but with eslint and all it's extension packages it can definitely get ugly. The resolution might be lerna but that seems again too much of a headache. With eslint, however, on VS Code we can open both folders separately and create a workplace. I don't want to get too caught up with making it work with _every_ editor. The root .prettier config becomes an issue then, but let's just remove it, we have only one default rule (who knows why?) - https://github.com/FightPandemics/FightPandemics/blob/6dd78920c708dff81a6843540818b9ef96d94892/.prettierrc.yml#L2 - so we don't need it. The default is fine.

Regarding node packages installed via docker vs host. I think if we go along with the npm-run-all install @camba95 has I think we are okay - we just need to tweak some docker lines & readme steps. It does feel wrong (containerization bad practice? - host node installing vs docker node running), but I think with Node >= 12 I can't imagine this causing issues for the most part (I had it working kind of like this for a bit actually).

I will make PR into this branch with tweaks with some ideas.

joshmorel:
>  It does feel wrong (containerization bad practice? - host node installing vs docker node running), but I think with Node >= 12 I can't imagine this causing issues for the most part (I had it working kind of like this for a bit actually).

Actually we should probably test this with Windows & Mac. Linux is no problem.

But any of these approaches under consideration - for docker installs - are relying on running one thing from the host (precommit hooks) and something else from the container (the actual program), regardless of what installs the node modules.

This `lefthook` library may be an interesting alternative to `husky` + `lint-staged` but still we have many of the same problems to consider:

https://dev.to/nitzano/linting-docker-containers-2lo6
https://github.com/Arkweid/lefthook

joshmorel:
@mannykary @camba95 I have 2 options to consider. Pros & cons are related to one another. These both seems like they will resolve the issue. There is still question of cross-platform issues, however, see below.

### [Install all from host](https://github.com/FightPandemics/FightPandemics/tree/786-fix-eslint-prettier-host-install)

`npm install` installs husky at root as well as backend/client node modules for both local & docker set-ups. Docker containers just run the programs.

Pros: Maybe slightly faster/simpler to get started up.
Cons: May run into cross-platform issues - installing all node modules on host.

### [Install in container at run time](https://github.com/FightPandemics/FightPandemics/tree/786-fix-eslint-prettier-install-on-run)

Docker containers install packages at run-time so available to the host for running eslint.

Pros: Unlikely to have cross platform issues - since docker is installing modules & only eslint is being run from the host. `npm install` on run no more questions when contributors run into errors due to a new package.
Cons: Might still have cross-platform issues? Need to test running pre-commit eslint on Mac & Windows. A little slower to get started up.

I prefer the 2nd option assuming tests with Mac/Windows work well. Totally fine on Linux. I can try Windows (hopefully, super old laptop) but not Mac. With the 1st option I would still have a concern going forward with cross-platform even if it works "for now".

> The downside is that it would be a lot slower to bring up containers. 

After the 1st run for a new container it isn't _that_ slow. 2s for backend & 9s for client to audit packages. For development no biggie IMHO.

camba95:
@joshmorel I tested it on a mac, np with that. What I can do it is to use one of this packages that executes the right command based on the OS, there may be some out there. Also I think that the manual and separated command to install dependencies on the host is OK, you may install it once every time you clone the project on a new machine. We gotta get used to the fact that FP will get bigger and so the starting up process.

joshmorel:
@camba95  I tested it on a mac, np with that. What I can do it is to use one of this packages that executes the right command based on the OS, there may be some out there. Also I think that the manual and separated command to install dependencies on the host is OK

Tested okay for both options - or specifically a) Install all from host or b) Install in container at run time .

@Zaydme or @idjevm wonder if you could take a look at the convo above?

rabbi08:
Closing this issue as low priority as of now. Please feel free to reopen if priority changes in future. thanks.

PR #1517, opened 2020-09-12
#1436 - Header: update profile to be logged in user or Org

shalinidc:
### What does this pull request aim to achieve? 💯
This PR is the FE for feature >> https://github.com/FightPandemics/FightPandemics/issues/1436.
_Please be concise and link any related issue(s):_

- [x] I have checked that no one else is working on similar changes.
- [x] I have read the latest [**README**](README.md) and understand the development workflow.
- [x] The name of this branch is: **`feature/<branch_name>`**, or **`hotfix/<branch_name>`** if this is a hotfix. The branch is created in a cloned version of this repo, **not a forked repo**.
- [x] This branch is rebased/merged with the **latest staging** (or the **latest production** if it is a hotfix).
- [x] There are no merge conflicts.
- [x] There are no console warnings and errors.
- [x] On the right hand side of this PR, I have linked the appropriate issue(s) under "Issues" and added a project progress under "Projects".

PR #1609, opened 2020-10-07
WIP authorize and set "author"

### LINKED ISSUE #1470: Add authorization middleware to check user has right to act as organisation
This is backend work to support #1493

See also comment: https://github.com/FightPandemics/FightPandemics/issues/1432#issuecomment-683336966

Code improvement so we're not relating code when checking that user is authorized to act as or update an organisation.

joshmorel:

Resolves #1470 

I thought we should get at least something for preliminary review from backend perspective. Originally I though merging all would be more ideal but it's turned out the existing design we are targeting has too many question marks and we need to clarify a few things.  

I was hoping there was some kind of very clear authentication policy option like we have with [Laravel](https://laravel.com/docs/8.x/authorization) but I couldn't find anything I was sure about. Here are a few I considered:

1. CASL - https://github.com/stalniy/casl. **Pros**: Isomorphic (could re-use policies on frontend React), integrated with mongoose. **Cons**: A bit confusion and I wasn't sure exactly how to integrate. ?Perhaps limited when it comes to checking role within team (as opposed to overall)? Not sure how well used/maintained it is
1. Casbin -   https://casbin.org/en/. **Pros** multi platform, could re-use policies on frontend React, maybe even native clients? **Cons**. More than a bit confusing.... ?Perhaps limited when it comes to checking role within team (as opposed to overall)?
1. gateman.js - https://github.com/NwangwuOsitadinma/gateman. **Pros**: Works with mongo data model. Might be worth considering later. **Cons**: only BE, doesn't seem well used/maintained. 

Anyways so far my code works with existing post as org/user and will be usable for FE when we find a way to comment/like as an org. But that's it for now (no viewing feed as if org).

@davidoduk @tejasm2 @emtay @CamilaArias 
**When it comes to alignment with product item ** https://www.notion.so/fightpandemics/0562fe39e5364561a0e12f76f19f8996?v=f9b57e5e8a3448e0b63bc4154b38ddf5&p=dd17a9eea26143bf8782ef2a73a19042

Currently it will not quite work with this but could with some modifications... As I raised in Slack recently I think if we are going with this "universal switch" one problem currently is it we need add designs (mobile - to see who is active user) in Figma as well as revamp  flows for some others (e.g. clicking on Organisation in profile menu, create a post flow, more) before I think we can confidently code this approach.

I'm not saying this is the way it has to be forever but maybe adding the user/org toggle in the PostSocial section would be a better option for launch until we clarify requirements and have a comprehensive design in Figma? This is similar to Facebook (personally I think it's also better UX for our use case as well especially considering future org "roles" but that's just me...).

As a reminder from Facebook (IU = Individual User, OU = Organisation User):

1. By default the PostSocial is in the context of yourself (even if during last visit you were the "page/business")
image
2. But you can toggle and the context is updated, then you can like/comment etc as the selected org
image

When it comes to messaging it's pretty straight forward... You visit the org page and then the context is switched you can message as that org (OR not if you don't have the permission later). 

Just ideas we can discuss further through some other forum
 

### DIFF
--- a/backend/lib/endpoints/posts.js
+++ b/backend/lib/endpoints/posts.js
@@ -266,42 +266,11 @@ async function routes(app) {
   app.post(
     "/",
     {
-      preValidation: [app.authenticate],
+      preValidation: [app.authenticate, app.setAuthor],
       schema: createPostSchema,
     },
     async (req, reply) => {
-      const { userId, body: postProps } = req;
-      // return postProps;
-      const [userErr, user] = await app.to(User.findById(userId));
-      if (userErr) {
-        req.log.error(userErr, "Failed retrieving user");
-        throw app.httpErrors.internalServerError();
-      } else if (user === null) {
-        req.log.error(userErr, "User does not exist");
-        throw app.httpErrors.forbidden();
-      }
-
-      // Author defaults to user unless organisationId set
-      let author = user;
-      const { organisationId } = postProps;
-      if (organisationId) {
-        const [orgErr, org] = await app.to(User.findById(organisationId));
-        if (orgErr) {
-          req.log.error(userErr, "Failed retrieving organisation");
-          throw app.httpErrors.internalServerError();
-        } else if (org === null) {
-          req.log.error(userErr, "Organisation does not exist");
-          throw app.httpErrors.forbidden();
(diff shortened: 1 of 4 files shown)


joshmorel:
Closing as (different but related) changes applied in #1517

Issue #1426, opened 2020-08-25
Notification Backend and DB [Desktop][Mobile] - for creating notifications


nfranklin9:
As a commentor/author of a post, I want to receive a notification for every comment posted, as well as when a post or comment is edited.

Current Scenario: When someone comments on my post, I don't know right away because I'm not sent an email or notification.

Expected results: As soon as someone comments on my post or edits it, I should be sent a notification to alert me.

This ticket is to create the database, middleware & queue for creating notifications and managing them on the queue.
middleware would be in node.js, queue would be SNS

Environment: node.js middleware, AWS SNS for queue
Expected results: Create notifications and manage them in the SNS queue
Actual results: Does not exist

mannykary:
Based on our meeting today, here is our updated plan for the backend:
* We agreed to not use Mongo Atlas for the MVP as it would add unnecessary overhead for CI/CD, and we already have it set up for AWS/Lambda
* We will add a “trigger” to the existing web backend which will create a notification when a like or comment is created
* Lambda will run on a cron job to query for new notifications, and then group the notifications to send an email
* We need to refine the model and make it simpler, as mentioned by Josh above
* We need to sync up with Zayd to see how the web notifications will be implemented, and also share the tweaked model to see if we need to add or modify it to make it work with web notifications
* HTML template provided by Cami above will be used by the Lambda to build the email
* Lambda will send the email through SES
* We need an algorithm for grouping notifications for daily, weekly, and bi-weekly emails

I recommend that tasks be broken up as follows:

1. Refine/finalize Notification data model
2. Add logic to web backend to create a new notification on new like/comment
3. Add a Cloudwatch rule to Terraform to run the Lambda every 5 minutes
4. Create basic code in Lambda to query notifications and send an email for any unread notifications
5. Come up with algorithm for grouping notifications for digest emails, and add this logic to the Lambda
6. Finalize HTML templates for the notification emails

Web notifications is covered in #1428 and #1531. We need to meet with Zayd to discuss this implementation

Unsubscribe from notifications is covered in #1507. We will need to generate unsubscribe links in the Notification Lambda, and add a new unsubscribe page to the web frontend (with corresponding logic in the web backend to toggle the user's notification preferences).

PR #1639, opened 2020-10-12
#1426 Notification model

mannykary:

Notification model (draft)

### DIFF
--- a/backend/lib/models/Notification.js
+++ b/backend/lib/models/Notification.js
@@ -1,0 +1,58 @@
+const { Schema, model } = require("mongoose");
+
+const notificationSchema = new Schema(
+  {
+    action: {
+      enum: ["like", "comment", "post"],
+      required: true,
+      type: String,
+    },
+    post: {
+      id: {
+        ref: "Post",
+        required: true,
+        type: Schema.Types.ObjectId,
+      },
+      title: String,
+    },
+    receiver: {
+      ref: "User",
+      required: true,
+      type: Schema.Types.ObjectId,
+    },
+    readAt: Date,
+    emailSentAt: Date,
+    triggeredBy: {
+      id: {
+        ref: "User",
+        required: true,
+        type: Schema.Types.ObjectId,
+      },
+      name: String,
+      photo: String,
+      type: String,
+    },
+  },
+  { collection: "notifications", timestamps: true },
+);
+
+/* eslint-disable sort-keys */
+notificationSchema.index({
+  createdAt: -1,
+}, {
+  expireAfterSeconds: 1296000,  // 15 days
+});
+notificationSchema.index({
+  readAt: -1,
+});
+notificationSchema.index({
+  emailSentAt: -1,
+});
+/* eslint-enable */
+
+const Notification = model("Notification", notificationSchema);
+
+module.exports = {
+  model: Notification,
+  schema: notificationSchema,
+};


backend/lib/models/Notification.js
```diff
@@ -0,0 +1,47 @@
...
+      title: String,
+    },
+    receiver: {
+      ref: "User",
+      required: true,
+      type: Schema.Types.ObjectId,
+    },
+    seenAt: Date,
+    sentAt: Date,
+    triggeredBy: {
```
Zaydme, line 25: Maybe we want the id too? To make a hyperlink to that user's profile

mannykary: Yes, good catch.

backend/lib/models/Notification.js
```diff
@@ -0,0 +1,47 @@
...
+      id: String,
+      title: String,
+    },
+    receiver: {
+      ref: "User",
+      required: true,
+      type: Schema.Types.ObjectId,
+    },
+    seenAt: Date,
+    sentAt: Date,
```
Zaydme: Maybe a boolean is enough?
For seenAt at least.
Maybe we can keep sentAt as a date if we want to send the same notification on daily + weekly digest etc.. and not only once.

mannykary: Yeah, I was thinking maybe `seenAt` could be useful for BI/analytics to see interaction of users (future-proofing) but if that's definitely not something that would happen, then we can change it to a boolean and rename it to `isRead`.

Zaydme (approved):
LGTM

mannykary:
Just added TTL of 15 days since we probably don't want to keep notifications around forever, but obviously if we want to keep some of this data for BI/analytics, we could increase the TTL to a longer timeframe.

backend/lib/models/Notification.js
```diff
@@ -0,0 +1,49 @@
...
+    },
+  },
+  { collection: "notifications", timestamps: true },
+);
+
+/* eslint-disable sort-keys */
+notificationSchema.index({
+  createdAt: -1,
+  expireAfterSeconds: 1296000,  // 15 days
+});
```
Zaydme: I am not sure but maybe this should be
```js
notificationSchema.index({
   createdAt: -1,
},
{ expireAfterSeconds: 1296000, }
);
```
As in [mongodb docs](https://docs.mongodb.com/manual/tutorial/expire-data/#expire-documents-after-a-specified-number-of-seconds)

mannykary: Yup, good catch.

backend/lib/models/Notification.js
joshmorel: > Yeah, I was thinking maybe seenAt could be useful for BI/analytics to see interaction of users (future-proofing) but if that's definitely not something that would happen, then we can change it to a boolean and rename it to isRead.

I think we should keep it as a date for future proofing. It feels like it should be `readAt` however (maybe because `sentAt` looks too similar). This would be when the user reads clicks on the notification (or marks all as read) in the web app?

The `sentAt` is then when the lambda sends the email for that same notification correct?

backend/lib/models/Notification.js
```diff
@@ -0,0 +1,50 @@
...
+    },
+    receiver: {
+      ref: "User",
+      required: true,
+      type: Schema.Types.ObjectId,
+    },
+    seenAt: Date,
+    sentAt: Date,
+    triggeredBy: {
+      id: String,
```
joshmorel:
```suggestion  id: {
    ref: "User",
    type: ObjectId,
  },
```

This should be a `ObjectId` `ref`

backend/lib/models/Notification.js
```diff
@@ -0,0 +1,50 @@
...
+
+const notificationSchema = new Schema(
+  {
+    action: {
+      enum: ["like", "comment", "post"],
+      required: true,
+      type: String,
+    },
+    post: {
+      id: String,
```
joshmorel, line 11: id: {
    ref: "Post",
    required: true,
    type: ObjectId,
  },

joshmorel:
```suggestion
    id: {
      ref: "Post",
      required: true,
      type: ObjectId,
    },
```

joshmorel (changes requested):
Thanks @mannykary. Besides the suggestions, if you agree I think we should keep `seenAt` as a date but rename to `readAt`.

backend/lib/models/Notification.js
mannykary: > The sentAt is then when the lambda sends the email for that same notification correct?

Yes. Maybe `sentAt` should be renamed to `emailSentAt`?

I will rename `seenAt` to `readAt`. And yes, this timestamp would be for when the notification is marked as read in the web app (and in the future, in the mobile native app).

joshmorel: Sounds good! @mannykary

backend/lib/models/Notification.js
ddveloper: it looks good to me

backend/lib/models/Notification.js
ddveloper: I am trying to follow the idea, so who's the responsibility to update the readAt and emailSentAt? I guess FE will call BE to update the readAt once the user read the notification, will redis update the emailSentAt? (sorry if that's a silly question)

ddveloper (approved):

backend/lib/models/Notification.js
Zaydme: Maybe we can avoid the additional queries required to return the referenced documents, since notifications have a 15 days TTL and are sent shortly, we shouldn't worry much about changes in post or triggeredBy, but I could be wrong.

backend/lib/models/Notification.js
mannykary: @ddveloper The Lambda will need to filter out any notifications for which it has already sent an email, and so it can set the `emailSentAt` field after successfully sending an email for the notification. So the next time in queries for notifications, it can filter out any that have `emailSentAt` set to a value. It will also filter out any notifications for which the `readAt` field is set as well.

backend/lib/models/Notification.js
mannykary: Are referenced documents always queried (I guess this is eager loading)? I am not familiar with Mongo or Mongoose. Is there a concept of "lazy loading" where the referenced documents are only fetched if the field is accessed?

mannykary: Ah, found the relevant Mongoose doc about eager loading: https://mongoosejs.com/docs/populate.html
By default, the related documents are not fetched unless you call `populate()`. But it is still an extra query if we use it. I guess there is no way around the N+1 query problem due to the nature of MongoDB.

Should we just include the ref as well as the other three fields? Like @Zaydme said, I don't think it matters too much that the name/photo changes for notifications.

mannykary:
@joshmorel @Zaydme @ddveloper I went ahead and made the changes per your suggestions. I also did research on whether there is a performance impact with regards to the document references incurring extra queries and I think it should be fine as long as we don't eager load using `populate()`, if I'm interpreting the Mongoose [doc](https://mongoosejs.com/docs/populate.html) correctly. @Naraujo13 can you confirm that this is correct?

backend/lib/models/Notification.js
```diff
@@ -0,0 +1,58 @@
+const { Schema, model } = require("mongoose");
+
+const notificationSchema = new Schema(
+  {
+    action: {
+      enum: ["like", "comment", "post"],
```
ddveloper, line 6:
```suggestion
      enum: ["message", "like", "comment", "post"],
```

Zaydme: @ddveloper messages notifications are not stored here, we can directly query the Threads collection with `newMessages: true`

ddveloper: Does that mean Lambda will query Threads also for sending email notifications? I thought Socket will generate these notifications in the same format and Lambda then will query Notifications only

Zaydme: We don't need to create a notification document for every message, all we need is to notify the user that they have X unread messages from Y. which is already stored on the Threads collection, so just using that makes more sense.

ddveloper: Thanks @Zaydme

backend/lib/models/Notification.js
Naraujo13: A bit late here but:

You're right it will only query the referenced doc if you call `populate()` on that field.
Populate should make a single query for all referenced docs, the N+1 problem is just if you want to call `populate()` for an array of docs or something similar. Then if that is something usual in this use case you can just create a custom method in the model that does a single custom lookup and get all the data with a single query.

Issue #1637, opened 2020-10-11
Web Notifications Backend


mannykary:
The scope of this ticket is to integrate socket.io with web notifications. As discussed in the meeting on 10/11, this will involve taking the existing socket.io Fastify plugin that is in the direct messages branch `feature/message-modal/1457` and updating it to support notifications. We also need to create the notifications in the database.

mannykary:
@Zaydme I created the branch `feature/1637-notifications-web-backend`. I was about to merge in the notification lambda code into this branch, but realized that web notifications does not depend on any of that code, so right now this is an exact copy of `feature/message-modal/1457`. I think you might need the code that is in the `feature/notification-drop-down/1531`, so you will probably have to merge that in.

Latha-AP:
on this staging link for web notifications:
https://feature1637-notifications-web-backend.fightpandemics.xyz/
1. we could create the post. 
2. we could like the post: **Web notification shown.**
3. **we could n't comment on the post!**
4. 2 times, today I could able to comment on own post, but not on other. 

![ezgif com-gif-maker (1)](https://user-images.githubusercontent.com/72217779/96402731-7e2df700-118b-11eb-8ec0-5e2fe7a81516.gif)

![comment notification disabled](https://user-images.githubusercontent.com/72217779/96420735-b5f86700-11aa-11eb-940d-7c1ca68cae3e.png)


****

claudiang1357:
I was in a post to attempt updating post comment (stay for quite some time - probably expired?), error was displayed as per screenshot -> Error message is overlapped with Help Board button.
![Overlapping Error](https://user-images.githubusercontent.com/72559826/96641113-3675af80-12f2-11eb-9302-baa87e414bc2.png)

Zaydme:
@Latha-FP [this](https://github.com/FightPandemics/FightPandemics/issues/1637#issuecomment-711550797) was fixed

shrutigupta1981:
I think second 2 is not required?
https://files.slack.com/files-pri/T010BHVS3EF-F01CWAC3K9C/image.png

shrutigupta1981:
Should we have scroll as notification nos increases?
https://files.slack.com/files-pri/T010BHVS3EF-F01D890UQ2W/image.png

shrutigupta1981:
Scenario: I created a post and left comment on it. I received comment from another user saying interested. I replied back to that comment. I myself shouldn't get notification in such a case? Right? No notification should be generated.
[...]

PR #1677, opened 2020-10-20
Feature/1427 notifications unsubscribe

### LINKED ISSUE #1427: Notification Settings [Desktop][Mobile] - for users to turn Notifications on or off
I want to be able to turn Notifications on or off, so I can receive Notifications when I want them (when I have an active post) and turn them off when I don't need them.

Expected results: A Notification Settings page where users can toggle Notifications to be either on or off. 
Actual results: Does not exist

To start (version 1), we can just have booleans on the user - on or off for all emailNotifications. Later we can add  webNotifications and controls by type. 

This can be broken up into 3 parts. 
a.	B/E update mongoose model to include booleans.
b.	F/E update form - to include Booleans. 
Initial design: https://fightpandemics.slack.com/archives/C018ZNAD62Y/p1598344009136700 
c.	We also need a special backend route for unsubscribing from email notifications via email - we'd need to authenticate with some other kind of token then what we use for auth

### LINKED ISSUE #1507: Notifications: Email Notification Unsubscribe link
I receive email notifications when users interact with my post. I need a link at the bottom of these emails to "Unsubscribe" me from these emails and access the Notification Settings page to change my email preferences. The "Unsubscribe" link may either link directly to the Notification Settings page, or unsubscribe the user in some other way (and a different link goes to the Notification Settings page).
Notification settings page: https://github.com/FightPandemics/FightPandemics/issues/1427

Acceptance Criteria:

- [ ] The Unsubscribe link is included in all Notification emails

- [ ] The Unsubscribe link is shared with other FightPandemics components that send emails (Messaging, Signup, etc)

- [ ] The Unsubscribe link allows users to stop receiving email notifications.

This unsubscribe button will call the backend to process the unsubscribe and update mongodb user to set them to opted out (or maybe keep track of opt outs in a separate model, with a reference to user)
https://fightpandemics.slack.com/archives/C018ZNAD62Y/p1599171215043400

I think it’s actually required that the unsubscribe page be accessible directly from the email per CAN-SPAM requirements. The CAN-SPAM Act requires that every email sent must contain a way to unsubscribe from that email list. In addition to offering the opt-out option, it must be easy to find and uncomplicated to do.

-The account update via web form should be relatively simple but we also need something that can be done securely from a link in the email
-We should include a token in the unsubscribe URL that the backend can validate
-We also need a special backend route for unsubscribing from email notifications via email - we'd need to authenticate with some other kind of token then what we use for auth. https://fightpandemics.slack.com/archives/C018ZNAD62Y/p1598202665129800
-We may need a backend route and store settings in the backend

ddveloper:

update User's model to have a notification preference field, as discussed in #1427 

- [ ] This branch is rebased/merged with the **latest staging** (or the **latest production** if it is a hotfix).


backend/lib/endpoints/schema/notifypreference.js
```diff
@@ -0,0 +1,45 @@
...
+  );
+
+const updateNotifyPrefsSchema = {
+  body: strictSchema()
+    .prop("notifyPrefs", notifyPreferenceSchema),
+};
+
+module.exports = {
+  updateNotifyPrefsSchema,
+};
```
Zaydme: Maybe this could be added to `/backend/lib/endpoints/schema/users.js` as part of `updateUserSchema` if you are going to use the same `PATCH /api/users/current` endpoint

backend/lib/models/NotifyPreference.js
```diff
@@ -0,0 +1,35 @@
...
+    daily: { default: true, type: Boolean },
+    weekly: { default: false, type: Boolean },
+    biweekly: { default: false, type: Boolean },
+  },
+});
+
+const NotifyPreference = model("NotifyPreference", notifyPreferenceSchema);
+
+module.exports = {
+  model: NotifyPreference,
```
Zaydme: You don't need to make it a model, just a schema. as it will be part of the `User` model

backend/lib/endpoints/schema/notifypreference.js
ddveloper: Right @Zaydme , I was thinking to use this for the `users/unsubscribe` endpoint, in which we don't have full User's data in the request, the user is not authenticated at F/E at that time, and we use a separate token to validate it's the correct user

backend/lib/models/NotifyPreference.js
ddveloper: Got it, thanks Zayd

backend/lib/endpoints/schema/notifypreference.js
Zaydme: `notifyPreferenceSchema` could be exported and re-used in `updateUserSchema` but also keeping everything in `/backend/lib/endpoints/schema/users.js` would be better

ddveloper: that makes sense, I will update it now 👍

backend/lib/endpoints/users.js
```diff
@@ -285,6 +291,38 @@ async function routes(app) {
       return new User(userData).save();
     },
   );
+
+  app.patch(
+    "/unsubscribe",
+    { schema: updateNotifyPrefsSchema },
+    async (req) => {
+      const { notifyPrefs } = req;
+      const token = getCookieToken(req);
```
ddveloper: to test this, I modified it to be `const { token, notifyPrefs } = req.body;` instead of getting from the cookie, Postman test ok, I do have a question here, should we put the token inside the cookie or inside the body?

ddveloper (commented):
Manny answered my question, leave token in headers, and notifyPrefs in body

backend/config.js
```diff
@@ -58,6 +59,7 @@ const config = {
     secretKey: configData.AUTH_SECRET_KEY,
     state: configData.AUTH_STATE,
   },
+  unsubscribeKey: configData.UNSUBSCRIBE_KEY,
```
mannykary: Minor: we actually have an ESLint rule to order keys in alphabetical order (which is actually supposed to run on a Husky post-hook commit but it doesn't work for every developer). I would still move this so that it is in alphabetical order in the object.

backend/lib/endpoints/schema/users.js
```diff
@@ -77,6 +77,44 @@ const getUserByIdSchema = {
...
+      .prop("biweekly", S.boolean().default(false)),
+  )
+  .prop(
+    "post",
+    S.object()
+      .prop("instant", S.boolean().default(false))
+      .prop("daily", S.boolean().default(true))
+      .prop("weekly", S.boolean().default(false))
+      .prop("biweekly", S.boolean().default(false)),
+  );
```
mannykary, line 82: We will be adding "share", right?

backend/lib/endpoints/users.js
```diff
@@ -476,6 +482,57 @@ async function routes(app) {
...
+      }
+      const [err, user] = await app.to(User.findById(userId));
+      if (err) {
+        req.log.error(err, `Failed retrieving user userId=${userId}`);
+        throw app.httpErrors.internalServerError();
+      } else if (user === null) {
+        throw app.httpErrors.notFound();
+      }
+
+      return user;
```
mannykary: Since this is an unauthenticated endpoint, we should only return the `user.notifyPrefs`. Otherwise it could be a security issue if we return everything stored in `user`, since the unsubscribe link is send in an email, and email is not really a secure medium of communication.

mannykary (changes requested):

backend/lib/endpoints/schema/users.js
ddveloper: I saw @Zaydme  updated the `post` in #1637 to be `share` I will do a similar change

backend/config.js
ddveloper: Got it, Thanks Manny 👍

backend/lib/endpoints/users.js
ddveloper: I was thinking to show more user information, like user name etc. but I agree there could be potential secure issue, I will return only the preference value as suggested

ddveloper (commented):
@Zaydme @mannykary my latest commit add one default initialization for the `notifyPrefs` field, please have a look and see if that's the right thing / way to do.

This was trying to address a case when creating a new user but no notification preference was given in the request, it's better to default the field with the default preferences. so later when querying or unsubscribing we have data to refer to, otherwise, it will be undefined

client/src/pages/Unsubscribe.js
```diff
@@ -0,0 +1,165 @@
...
+
+const Unsubscribe = (props) => {
+  const { control, formState, handleSubmit } = useForm({
+    mode: "change",
+  });
+  const { t } = useTranslation();
+  const queryParams = useQuery();
+  const tokenValue = queryParams.get("token");
+  const config = {
+    headers: { token: tokenValue },
```
Zaydme, line 88: Looks like `tokenValue` is an object, as you can see in the dev tool, **note that this is when visiting `/unsubscribe` without a `token` query.** but I don't think this is a problem, as the jwt decoder will just count it as a string ¯\_(ツ)_/¯
image

backend/lib/models/NotifyPreference.js
```diff
@@ -0,0 +1,39 @@
...
+    daily: { default: true, type: Boolean },
+    weekly: { default: false, type: Boolean },
+    biweekly: { default: false, type: Boolean },
+  },
+  share: {
+    instant: { default: false, type: Boolean },
+    daily: { default: true, type: Boolean },
+    weekly: { default: false, type: Boolean },
+    biweekly: { default: false, type: Boolean },
+  },
```
Zaydme, line 27: I think we should add an `enabled` field, this way we can preserve the old values when the user disables email notifications completely, and can just enable them later without having to recheck the boxes to where they want them.

ddveloper: that sounds like a user experience feature, not strongly relevant to our model design, I wonder if we have another way to do that without modifying our model ( ⚆ _ ⚆ )

client/src/pages/Unsubscribe.js
ddveloper: I tried to change it to `headers: { token: JSON.stringify(tokenValue) }` but that does not work, is that the way it supposes to be?

backend/config.js
```diff
@@ -40,7 +40,8 @@ const configData = envSchema({
     .prop("S3_CDN_BUCKET", S.string().default("fp-dev-cdn"))
     .prop("SENTRY_DSN", S.string())
     .prop("SENDGRID_API_KEY", S.string())
-    .prop("SENDGRID_CONTACTS_LIST_ID", S.string()),
+    .prop("SENDGRID_CONTACTS_LIST_ID", S.string())
+    .prop("UNSUBSCRIBE_KEY", S.string()),
```
Zaydme: can we use the same `AUTH_SECRET_KEY` we already have? if we did, we can get rid of the new "jwt-simple" dependency, and just use the existing `app.jwt.decode(token)`

client/src/pages/Unsubscribe.js
Zaydme: > I tried to change it to headers: { token: JSON.stringify(tokenValue) } but that does not work, is that the way it supposes to be?

you can just ignore it.

backend/config.js
ddveloper: @mannykary also suggested this on slack, I didn't find how to encode with "Fastify-jwt" at that time so I proposed to use "jwt-simple" instead, I will double-check this, (it seems the `sign` method can be used here instead)

Zaydme: yes the `sign` method is using for encoding, please check [this small example ](https://github.com/fastify/fastify-jwt#example-3)

backend/lib/endpoints/users.js
mannykary: In that case it would be good to selectively return only what we need, including user name.

backend/config.js
ddveloper: code updated with package `jsonwebtoken`, tested with Postman and F/E

backend/lib/endpoints/users.js
ddveloper: I don't show any user information right now at the F/E, could be a stretch goal : )

backend/config.js
ddveloper: @mannykary the only change at Lambda side will be changing the key to be `AUTH_SECRET_KEY`

backend/lib/endpoints/users.js
```diff
@@ -476,6 +484,74 @@ async function routes(app) {
...
+    let decoded = {};
+    jwt.verify(token, config.auth.secretKey, (err, payload) => {
+      if (err) {
+        req.log.error(err, `Invalid token for unsubscribing`);
+        throw app.httpErrors.badRequest("token is invalid");
+      }
+      decoded = payload;
+    })
+
+    const { userId, expireDate } = decoded;
```
mannykary: I think the format will be like `{ userId: 'abc', iat: 1603780773, exp: 1606372773 }`. Tried this in the Node console:
```
> const encodedToken = jwt.sign({ userId: 'abc' }, key, { expiresIn: '30d' });
undefined
> const decodedToken = jwt.verify(encodedToken, key);
undefined
> decodedToken
{ userId: 'abc', iat: 1603780773, exp: 1606372773 }
> new Date(decodedToken.exp*1000)
2020-11-26T06:39:33.000Z

Comment (ddveloper) [reply backend/lib/endpoints/users.js]: right, that's from package `fastify-jwt` and when I tried it, I got an error message likes:
```
    const decoded = app.jwt.verify(token, config.auth.secretKey);
    // error message: "verify must be called asynchronous if secret or 
    //                            public key is provided as a callback"
```

mannykary: Oh no that part is fine, what I wanted to highlight was that the `jsonwebtoken` library automatically adds the expiration in the `exp` key. This is how I'm signing the JWT in the Notification Lambda: https://github.com/FightPandemics/FightPandemics/blob/service/1426-notifications-lambda/services/notification/src/helpers/template-builder.js#L95

ddveloper: not sure if I imported the `jwt` wrong, how is your `jwt` got imported (or defined) @mannykary

ddveloper: Got it Manny, I will update my date checking accordingly

Zaydme:
@ddveloper @mannykary do I merge this with feature/1637-notifications-web-backend? or it's still WIP?

ddveloper:
> @ddveloper @mannykary do I merge this with feature/1637-notifications-web-backend? or it's still WIP?

Yes @Zaydme please merge it

Issue #840, opened 2020-06-26
DRY-up logic for returning own user from DB (low priority)


joshmorel:
As F/E is using `session.user` for a variety of things the fields included with user object returned from db might differ depending on if it is `POST /api/auth/oauth` `POST /api/auth/login` and `GET /users/current`. 

This should be put into some util function (maybe plugin?). 

Note: the main difference with respect to DB user for `GET /users/current` vs auth routes is for the former if user is null that's an error, while the latter there might not be a user in DB until user has created profile. 

At the very least extract desired fields from DB user object to return so they will always be the same e.g.:

```
    const {
      _id: id,
      about,
      email,
      firstName,
      hide,
      lastName,
      location,
      needs,
      objectives,
      organizations,
      urls,
    } = user;
```

ethanKocheril:
I'm going to try to work on this, this is for mongoDB, right?

PR #1716, opened 2020-10-28
Feature/comments 4 actor change

joshmorel:
This seems to work. Also addressed that issue with `organisations` is undefined.

We probably could DRY up the code a bit, in terms checking if the active user/org is the author of the comment. I may add that later today or tomorrow. Let me know your preferences on next steps @camba95

### DIFF
--- a/backend/lib/endpoints/users.js
+++ b/backend/lib/endpoints/users.js
@@ -230,8 +230,10 @@ async function routes(app) {
     "/",
     { preValidation: [app.authenticate], schema: createUserSchema },
     async (req) => {
-      const user = await Auth0.getUser(getCookieToken(req));
-      const { email, email_verified: emailVerified } = user;
+      const {
+        email,
+        email_verified: emailVerified,
+      } = await Auth0.getUser(getCookieToken(req));
       if (!emailVerified) {
         throw app.httpErrors.forbidden("emailUnverified");
       }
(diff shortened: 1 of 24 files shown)


backend/lib/endpoints/users.js
```diff
@@ -250,7 +252,11 @@ async function routes(app) {
         authId: req.user.sub,
         email,
       };
-      return new User(userData).save();
+      const user = await new User(userData).save();
+      return {
+        ...user.toObject(),
+        organisations: [],
```
camba95, line 258: why don't we save it as a default, `organisations: []`, instead of hard coding it

client/src/components/Feed/NestedComments.js
```diff
@@ -53,6 +48,7 @@ const NestedComments = ({
   const [showReply, setShowReply] = useState(false);
   const [editComment, setEditComment] = useState(false);
   const [editedComment, setEditedComment] = useState(comment.content);
+  const isAuthor = useSelector(selectIsAuthor);
```
camba95: is this a new pattern for redux? as much as I know selector return an specific sub object of the global state (state managed by redux) but a function. What you can do is return de actor Id (`selectActorId`) and then use that value to do the comparison.

joshmorel: I found something similar here - https://github.com/reduxjs/reselect/issues/140#issuecomment-227554080

But `selectActorId` will work well enough. I will make the change.

backend/lib/endpoints/users.js
joshmorel: `organisations` is a virtual based on a ref so this won't work. Hardcoding is less than ideal but this is what this #840 is about. This fix should address this in a consistent manner but let's work with this for now.

PR #1752, opened 2020-11-03
#840 - Defines logic for getting user from db

### LINKED ISSUE #840: DRY-up logic for returning own user from DB (low priority)

emanuelquerty:

This PR resolves #840 

This PR relates to #1749 and #1750


joshmorel:
Hey @emanuelquerty thank you. 

Now that I'm looking at the implementation with a util function, I think maybe a better option would be to leverage something via the ODM Mongoose. I would want it to work with the same results with calls to `.save()` https://github.com/FightPandemics/FightPandemics/blob/feature/840-logic-for-getting-user-from-db/backend/lib/endpoints/users.js#L405. 

This would be similar like the ORM Sequelize's scope - https://sequelize.org/master/manual/scopes.html - but I don't see any good equivalents in mongoose.

Maybe some combination of schema options - https://mongoosejs.com/docs/api/schema.html#schema_Schema - and something else. I know we will need toJSON: { virtuals: true }, but there are some gotchas... Needs research

emanuelquerty:
Hi @joshmorel so per your comments, I have the toObject return `id,
        about,
        email,
        firstName,
        hide,
        lastName,
        location,
        needs,
        objectives,
        organisations,
        urls,
        photo,` which eliminates the need for destructuring anywhere we need the user object with the fields bellow. I also updated the utility function to reflect that change. I left the utility function as it is still useful for returning the error in the case of calls to **GET /users/current** and for returning an object with a null user for auth routes. 

I've pushed the new changes to this branch. As per your other comments in the slack thread such as using scopes, although I did found libraries for using scopes with mongoose, I couldn't see the use case in the given tasks I'm working on as I was able to do it without it, at least the way I understood you envisioned this. Maybe I missed something else that you thought it could be done with scopes. Let me know after you see the changes and how it's working so far if there's anything else you think I missed or that I can do to have this work as you envisioned this.

rabbi08:
Closing this issue as low priority as of now. Please feel free to reopen if priority changes in future. thanks.

Issue #782, opened 2020-06-22
Auth0 user should be logged off from all tabs


galeza:
env: staging

Steps:
1. Open staging in the first tab and log in. Go to profile screen
2. Open staging in the second tab. You are already logged in and you can open your Profile page. 
3. In the second tab click Logout
4. Go to the first tab
In the first tab Navigation bar shows Profile (it's for a logged in user). When you click Profile you'll see error: Failed loading profile, reason: No Authorization was found in request.cookies

I would expect that my user will be logged of in both tabs. And in the first tab I'll see Login, Register buttons.

Desktop:
![Screen Shot 2020-06-22 at 20 01 13](https://user-images.githubusercontent.com/21987083/85320795-f299a780-b4c3-11ea-8757-518817aac6dd.png)

joshmorel:
Thanks @galeza. I guess we could run setInterval on `App.js` that checks the remember cookie every second or few seconds and if it's not there dispatches `AUTH_LOGOUT`. @robinv85 if you have another suggestion let me know. I'll start work on this later today.

joshmorel:
I checked Facebook, LinkedIn & Amazon and none will log you out of the other tab with something like `setInterval` doesn't mean it's not a bad idea but dunno what the best practice is. Facebook does log you out after OG graph API call returns error (not sure if they are timed or random).

No matter, clicking a LoggedIn only page (like Profile in example) when the `remember` cookie is not present should redirect to home. I can at least add that and we can consider `setInterval` later.

galeza:
Hi @joshmorel 

Ok, agree. So we can redirect user to Home screen or maybe Login screen?

joshmorel:
@galeza That's okay I think the title is fine.

tejasm2:
Hey @joshmorel should I add this to "done"?

PR #1760, opened 2020-11-04
Update user avatar on navigation sidebar

### LINKED ISSUE #1757: The user avatar does not update, if avatar is uploaded from another device, unless user logout and login again.
1. Login into STG 'https://staging.fightpandemics.work' using valid credentials from Desktop
2. Navigate to user profile and upload the avatar.
3. Open FP - STG on the mobile, and user is already logged in on its mobile device with same account.

Issue: You will observe that the user avatar is not updated on the left navigation pop up, though it is already updated on the profile and post. User needs to logout and login again to sync the navigation sidebar and profile avatar. i.e. the session need to be reset.
Refer below GIF

ezgif com-video-to-gif 5

This issue exist for users who keeps the session active multiple devices. mainly on mobile as we have to display user avatar on profile page and left navigation sidebar.

Zaydme:
Makes the NavigationLayout have it's own UserContext so it can update the username/avatar and not use the global user object.
This update will be performed on route change.

closes #1757 

### DIFF
--- a/client/src/templates/layouts/NavigationLayout.js
+++ b/client/src/templates/layouts/NavigationLayout.js
@@ -1,7 +1,7 @@
 import { Drawer, List, Button, WhiteSpace } from "antd-mobile";
 import { Typography, Menu, Dropdown } from "antd";
 import axios from "axios";
-import React, { useState, useReducer } from "react";
+import React, { useState, useReducer, useContext, useEffect } from "react";
 import { Link, useHistory } from "react-router-dom";
 import styled from "styled-components";
 import { useTranslation } from "react-i18next";
@@ -40,7 +40,8 @@ import Logo from "components/Logo";
 import logo from "assets/logo.svg";
 import GTM from "constants/gtm-tags";
 import ProfilePic from "../../components/Picture/ProfilePic";
-
+import { UserContext, withUserContext } from "context/UserContext";
+import { fetchUser, fetchUserSuccess } from "hooks/actions/userActions";
 const { royalBlue, tropicalBlue, white } = theme.colors;
 
 const drawerStyles = {
@@ -263,10 +264,18 @@ const StyledDrawer = styled(Drawer)`
 
 const NavigationLayout = (props) => {
   const { t } = useTranslation();
-  const { authLoading, mobiletabs, navSearch, tabIndex, isAuthenticated, user } = props;
+  const {
+    authLoading,
+    mobiletabs,
+    navSearch,
+    tabIndex,
+    isAuthenticated,
+  } = props;
   const history = useHistory();
   const [drawerOpened, setDrawerOpened] = useState(false);
   const [searchKeywords, setSearchKeywords] = useState(false);
+  const { userProfileState, userProfileDispatch } = useContext(UserContext);
+  const { user } = userProfileState;
 
   const handleSearchSubmit = (inputValue) => {
     setSearchKeywords(inputValue);
@@ -287,6 +296,20 @@ const NavigationLayout = (props) => {
     },
     { stateKey: "generalFeedback", label: t("feedback.otherFeedback") },
   ];
+
+  useEffect(() => {
+    if (!props.isAuthenticated || !props.user) return;
+    (async function initialProfileFetch() {
+      userProfileDispatch(fetchUser());
+      try {
+        const res = await axios.get(`/api/users/current`);
+        userProfileDispatch(fetchUserSuccess(res.data));
+      } catch (err) {
+        // fallback to the global user.
+        userProfileDispatch(fetchUserSuccess(props.user));
+      }
+    })();
+  }, [isAuthenticated, props.isAuthenticated, props.user, userProfileDispatch]);
 
   const displayAvatar = (user) => {
     if (user?.photo || (user?.firstName && user?.lastName)) {
@@ -745,4 +768,4 @@ const NavigationLayout = (props) => {
   return <>{renderNavigationBar()}</>;
 };
 
-export default NavigationLayout;
+export default withUserContext(NavigationLayout);
--- a/client/src/templates/layouts/NavigationLayout.js
+++ b/client/src/templates/layouts/NavigationLayout.js
@@ -298,7 +298,7 @@ const NavigationLayout = (props) => {
   ];
 
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


rabbi08 (approved):
Uploaded avatar from Desktop -> view on mobile, already opened application, just navigate to helpboard or so. the left navigation bar is updated along with profile pic with same avatar.

Uploaded avatar from Mobile-> view on Desktop, already opened application, just navigate to profile page or helpboard or other Org or so. the profile pic is updated as expected.

Thanks @Zaydme

Zaydme:
This solution will work but it's not ideal, because it fetches the current user data on every route change, I don't know if this is fine.
But I can't think of any other way since we can't predict if the user was updated or not, so we just have to fetch it to find out.
Any thoughts @joshmorel ?

galeza:
Hi,

I tested using: macOS and Android
Scenario 1: logged in on both devices
- upload avatar pic on desktop
- click Profile screen on Android - avatar was updated

Scenario 2: logged in on both devices
- upload new avatar on Android device
- click Help Board on ma macOS - avatar was updated for all my posts.

galeza (approved):
LGTM

client/src/templates/layouts/NavigationLayout.js
```diff
@@ -288,6 +297,20 @@ const NavigationLayout = (props) => {
     { stateKey: "generalFeedback", label: t("feedback.otherFeedback") },
   ];
 
+  useEffect(() => {
```
joshmorel, line 300: I can't recreate the issue in staging as described or I don't understand it...

But I agree with your point @Zaydme I don't think this is ideal. We shouldn't be fetching the user in this place just for this purpose...

In any case we will be changing navigation significantly with #1767 and then maybe again with some other redesign work so I don't think we should merge this until at least 1767 is done.

joshmorel (changes requested):

client/src/templates/layouts/NavigationLayout.js
joshmorel: I'm caught up on the issue. I understand now.

client/src/templates/layouts/NavigationLayout.js
```diff
@@ -263,10 +264,18 @@ const StyledDrawer = styled(Drawer)`
...
+    authLoading,
+    mobiletabs,
+    navSearch,
+    tabIndex,
+    isAuthenticated,
+  } = props;
   const history = useHistory();
   const [drawerOpened, setDrawerOpened] = useState(false);
   const [searchKeywords, setSearchKeywords] = useState(false);
+  const { userProfileState, userProfileDispatch } = useContext(UserContext);
```
joshmorel, line 277:
```suggestion
```

But for the solution we should stick with `user` which comes from https://github.com/FightPandemics/FightPandemics/blob/1a471543dca12ccb00a7b63367ab3ef58356e2de/client/src/reducers/sessionReducer.js#L47. 

`UserContext` here is very confusing name wise, it should be called ProfileContext and shouldn't be used outside of org/user profile pages. Not sure when/why we diverted on state management but we definitely should be aiming for something more clear & consistent. A goal for post wide launch...

client/src/templates/layouts/NavigationLayout.js
```diff
@@ -288,6 +297,20 @@ const NavigationLayout = (props) => {
     { stateKey: "generalFeedback", label: t("feedback.otherFeedback") },
   ];
 
+  useEffect(() => {
+    if (!isAuthenticated || !props.user) return;
+    (async function initialProfileFetch() {
```
joshmorel, line 302:
```suggestion
    refetchUser();
```

Instead I'd dispatch this action (which will update the global user) similar to https://github.com/FightPandemics/FightPandemics/blob/1a471543dca12ccb00a7b63367ab3ef58356e2de/client/src/pages/ProfileCompleted.js#L23

joshmorel (changes requested):
Now that I understand the problem (the videos caused confusion for me this time) I've added more specific suggestions.

However, another thing that could be solved here with some tweaks is to also "log them out" if they are no longer authenticated (e.g. logged out in other tab). Currently, the page needs to be refreshed for this to be clear. I think there might be an old issue related to that...

joshmorel:
This is the related issue #782 (which I am still assigned :sweat_smile:).

I think we should aim to solve both these issues together as the mechanism for one should work for the other.

Either:

1. Similar to this resolution when they navigate the user is updated
2. Some kind of "global" heartbeat that queries the backend /api/users/current every 30-60s.

We'd just need to update `refreshUser` so if there's a 401 error thyen `dispatch({ type: AUTH_LOGOUT });`.

Or maybe modify `initAuth` function so it works for all uses?

Zaydme:
> We'd just need to update refreshUser so if there's a 401 error thyen dispatch({ type: AUTH_LOGOUT });.

Or maybe not call the API at all if the auth cookie doesn't exist, and just dispatch logout.

Having an interval running every 30/60 seconds would probably make more requests than updating the user on route change, we just need to know how often the user changes the route, I am thinking that they will spend most of the time on the help board and casually checking some profiles, also when the tab is just open without any activity, maybe in the background, there will be no interval spamming the API. Maybe business intelligence can tell us how often the user changes the route on a single session.

Zaydme:
@joshmorel 
Or even better, we can update the user on route change => store the time of the update, on the next route change, if it was less than 30 seconds ago don't refetch the user.
This way there will be no interval spamming the API, and nor quickly switching routes. So the user will be updated a maximum 1 time per 30 seconds and only when the user is actively using the website and not idle or in the background.

joshmorel:
> Or maybe not call the API at all if the auth cookie doesn't exist, and just dispatch logout.

You are right. We could just this on a `setTimeout` loop and dispatch logout if it no longer exists: https://github.com/FightPandemics/FightPandemics/blob/431a96a3b8fcc8dce952ad132c5a962f1f40f85f/client/src/actions/authActions.js#L17

I've changed my mind about resolving them together I think we can solve them separately. The logout in other tab thing wouldn't really have an API usage implication then so maybe something that can be resolved without further discussion. 

The API usage implication but be worth it here but maybe not? I would see the need to login and logout to see updated avatar a bug, but not the need to reload the page (which seems to work fine). We can discuss further in Slack.

PR #1820, opened 2020-11-14
Add indexes and emailSent field

mannykary:

Adding indexes for Notification Lambda to query more efficiently, and `emailSent` field.

### DIFF
--- a/backend/lib/plugins/socket.js
+++ b/backend/lib/plugins/socket.js
@@ -115,6 +115,7 @@ function onSocketConnect(socket) {
           id: participant._id,
           lastAccess: participant == sender ? new Date() : null,
           name: participant.name,
+          emailSent: false,
           newMessages: participant == receiver ? 1 : 0,
           photo: participant.photo,
           status: participant == sender ? "accepted" : "pending",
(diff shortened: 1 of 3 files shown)


backend/lib/plugins/socket.js
```diff
@@ -307,6 +308,10 @@ function onSocketConnect(socket) {
             // set status to "pending again" if it was "ignored".
             updates.$set = { "participants.$[userToUpdate].status": "pending" };
             break;
+          case "accepted":
```
Zaydme, line 311: @mannykary you should also reset it in `case "archived"` since it gets unarchived on new message and just becomes "accepted".

(I would suggest adding this to the updates.$set object above and delete the object key in `case "ignored"`)

Zaydme (changes requested):

backend/lib/plugins/socket.js
mannykary: Do we also need to add `case "pending"` and then do `delete update.$set`?

We need to keep the key for `case "archived"` but it is setting `$set` to a new object, so I will update it to update the existing object instead of clobbering it: `updates.$set["participants.$[userToUpdate].status"] = "accepted";`

For `case "ignored"`, I think it's fine to leave it since it will set `$set` to a new object, thus clobbering the `$set` for `emailSentAt`

Zaydme: Yes we need to add `case "pending"` and delete the entire updates.$set

PR #1821, opened 2020-11-14
#1741 Email notifications for DMs

### LINKED ISSUE #1741: Instant Email Notifications for Direct Messages
@Zaydme to add details on how to query for new DMs to send emails for.

mannykary:

Adding email notifications for direct messages.

Closes #1741 

- [ ] The name of this branch is: **`feature/<branch_name>`**, or **`hotfix/<branch_name>`** if this is a hotfix, where `<branch_name>` briefly describes the issue(s) resolved and is prefixed with the issue number(s) (e.g. `1127-1130-update-git-branching-model`). The branch is created in a cloned version of this repo, **not a forked repo**. **NOTE: using nodeploy/ prefix since there is only one Notification Lambda in review and I don't want to clobber it**

### DIFF
--- a/services/notification/src/helpers/database-helper.js
+++ b/services/notification/src/helpers/database-helper.js
@@ -1,5 +1,6 @@
 const { DateHelper } = require("./date-helper");
 const { EmailFrequency } = require("../models/email-frequency");
+const { MessageThreadStatus } = require("../models/message-thread-status");
 const { MongoClient } = require("mongodb");
 const { NotificationAction } = require("../models/notification-action");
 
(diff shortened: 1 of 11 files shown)


services/notification/src/templates/instant/message-request.html
```diff
@@ -0,0 +1,3 @@
+<h1>{{senderName}} sent you a direct message request</h1>
```
Zaydme: Most first messages will have a postRef, I think that could give the user an idea why they are getting a DM request.

services/notification/src/helpers/template-builder.js
```diff
@@ -57,10 +67,61 @@ class TemplateBuilder {
...
+          {
+            senderName: sender.name,
+          },
+        );
+        const view = {
+          baseUrl: this.baseUrl,
+          senderName: sender.name,
+        };
+
+        if (templateKey === "message") {
```
Zaydme, line 118: Probably for messageRequest too, since that's where postRef is more useful to know the reason of the request.

services/notification/src/helpers/database-helper.js
```diff
@@ -15,19 +16,158 @@ class DatabaseHelper {
...
+              },
+            },
+          },
+        },
+      ])
+      .toArray();
+
+    const threadIds = threads.map((thread) => thread._id);
+
+    // Convert threads array to object keyed by thread ID for O(1) lookup
```
ddveloper, line 110: this seems will still be O(n), [ES Spec](https://tc39.es/ecma262/#sec-array.prototype.reduce) description says current value comes from previous result, so there might be no parallel happening

ddveloper (approved):
LGTM

services/notification/src/helpers/database-helper.js
```diff
@@ -15,19 +16,158 @@ class DatabaseHelper {
...
+                        new Date(),
+                        this.instantUnreadLookbackInterval,
+                      ),
+                    },
+                  },
+                  {
+                    "participants.lastAccess": {
+                      $gt: DateHelper.subtractMinutes(new Date(), 30),
+                    },
+                  },
```
Zaydme, line 64: Do we need to keep this? sice we already have 
```js
"participants.newMessages": { $gt: 0 },
"participants.emailSent": false,
```
So it will not send to old threads except if they had newMessages > 0, and it will only send once so it's good, since `emailSent` will be true after that

Zaydme (approved):
LGTM!
image

services/notification/src/helpers/database-helper.js
mannykary: Yes, it's `O(n)` when converting from an array to object (which makes sense since it has to process every item in the array), but the reason for converting to an object (or hash lookup) is for `O(1)` lookup when it finds the message by thread ID [here](https://github.com/FightPandemics/FightPandemics/pull/1821/files#diff-63ecab986d8af29fb81344382b42ee27cfb405e5bd945c4d5dc3b70f37d27983R138). So overall `O(n) + n*O(1) = 2*O(n) ~ O(n)`. If we left it as an array, yes we save `O(n)` time in converting to an object but the lookup will be `O(n)` since it has to look for the element in the array by iterating through every element. And it's done in a loop, so it becomes `n*O(n) ~ O(n^2)`. `O(n) << O(n^2)`

services/notification/src/helpers/database-helper.js
mannykary: We don't technically need it but I think we should leave it in as a safety, so that we don't send out emails for very old messages. I've developed a notification lambda at work where it started sending out emails for years-old messages due to a bug (because we were not limiting the lookback period in the query), and it was a fiasco. So it would be better to leave it in here. We could increase this lookback if we want though, but I probably wouldn't set this to more than a day.

services/notification/src/helpers/database-helper.js
ddveloper: that makes sense, I misunderstood it ...

PR #1862, opened 2020-11-19
Feature/backend tests for /api/posts/{postId} endpoint and dbHelper

galeza:

Add tests for /api/posts/{postId} and dbHelper


galeza:
Hi @jessica-lau 

If you have time, please look at this solution. Once you asked me about adding integration with DB for our tests.
I added dbHelper.js and I'm retrieving data from DB.
What you think about having more integration tests?

jessica-lau:
> Hi @jessica-lau
> 
> If you have time, please look at this solution. Once you asked me about adding integration with DB for our tests.
> I added dbHelper.js and I'm retrieving data from DB.
> What you think about having more integration tests?

hey @galeza I looked at the solution and I think the post.test.js should include code to set up a dataset in the local mongoDB server before the test cases so that they can run off on the local data. If we're using the live server, the tests can come out flaky since data will always be changing. What do you think about this? Yea, I agree it's necessary to have more integration tests.

Another question I have is how come a post.js file was created in the the mocha>models folder instead of importing the schema in Post.js from the backend directly like this:

I'm just curious since this can reduce the need to create additional files.

In post.test.js:

`const { model } = require("../backend/lib/models/Post")`

In docker-compose.yml:
```volumes:
      - ./mocha/test/:/app/test/
      - ./mocha/models/:/app/models/
      - ./mocha/utils/:/app/utils/
      - ./backend/lib/models/:/app/backend/lib/models

**Comment by galeza** [comment]
Hi,
I added dbHelper.js.
The flow is as follow:
Before - we connect to the DB
Before each - Insert a new document to the DB
Test - compare objects
After each - Delete a document
After - Close DB

When I compare object returned from DB and object returned by API I omit some values, API object contains additional data. But "author.id", "createdAt", "updatedAt", "_id" have different types. Let me know if you know how to bypass that. I was using JSON.stringify, but it doesn't work.

Two more tests are verifying endpoint with different values (random ObjectID and Int). Issue [API] - endpoint GET /api/posts/{postId} #2088 was created.

FYI. 
1. Two Login Tests fail + one cypress Login test. I think that they should be corrected in a different PR
2. Another thought - we need refactor validators.js. Maybe we can make it generic. Pass map with key value and compare them. If you agree I add a new task.

**Comment by TimeaKovacs86** [comment]
I got the same result running the tests:  Two Login Tests fail + one cypress Login test as you said @galeza 
But somehow I couldn't run mocha integration: keep getting the same error message : Error: Cannot find module 'lodash', do you have any idea why is that? What did I do wrong?

**Comment by galeza** [comment]
> I got the same result running the tests: Two Login Tests fail + one cypress Login test as you said @galeza
> But somehow I couldn't run mocha integration: keep getting the same error message : Error: Cannot find module 'lodash', do you have any idea why is that? What did I do wrong?

@TimeaKovacs86 
I think that docker-compose up --build is necessary as I changed package and package-lock

Yes that was the issue! Thank you! Ran the tests again, got the same result as you and @jessica-lau  also.  Login test fails which, I agree, should be handled by a separate PR. Other than this... I don't know how you can bypass datatype differences. Sorry. :( Seems like I learn more than I can help...

File: mocha/utils/apiEndpoints.js [hunk 1]
```diff
@@ -3,3 +3,6 @@ exports.feedbackEndpoint = "/api/feedback";
 exports.signUpApiEndpoint = "/api/auth/signup";
 exports.organisationsEndpoint = "/api/organisations";
 exports.postsEndpoint = "/api/posts";
+exports.postApiEndpoint = "/api/posts/{postId}";
+exports.getPostApiEndpoint = "/api/posts/";
```
jessica-lau: Is this a duplicate? If I'm understanding correctly, I think line 8 "/api/posts" is being used in the test but not line 7 "/api/posts/".

mocha/utils/validators.js
```diff
@@ -39,3 +39,22 @@ exports.validateStatusCodeResponseBody = function (
     }),
   );
 };
+
+exports.validateStatusCodeAndMessage = function (response, statusCode, status) {
+  expect(response).to.have.a.property("statusCode").to.be.equal(statusCode);
+  expect(response).to.have.a.property("status").to.be.equal(status);
+  expect(response).to.have.a.property("ok").to.be.equal(true);
+};
```
jessica-lau, line 18: I think checking either the statusCode or status in the response is enough since both of them are the same but just stated differently, so no need for both. 200 is httpStatus.OK

mocha/utils/validators.js
```diff
@@ -39,3 +39,22 @@ exports.validateStatusCodeResponseBody = function (
...
+
+exports.validateInitialAndResponseObject = function (
+  initial,
+  initialPropToOmit,
+  response,
+  responsPropToOmit,
```
jessica-lau: Just a minor typo "responsPropToOmit" should be "responsePropToOmit"

mocha/utils/validators.js
```diff
@@ -39,3 +39,22 @@ exports.validateStatusCodeResponseBody = function (
...
+) {
+  var result = _.isEqual(
+    _.omit(initial, initialPropToOmit),
+    _.omit(response, responsPropToOmit),
```
jessica-lau: Just a minor typo "responsPropToOmit" should be "responsePropToOmit"

mocha/test/post.test.js
```diff
@@ -0,0 +1,124 @@
...
+    after(function () {
+      dbHelper.disconnect();
+    });
+
+    it("Success - user gets a post by ID", async function () {
+      let response = await apiHelper.sendGETRequest(
+        APP_URL,
+        getPostApiEndpoint + "/" + postID,
+      );
+      validator.validateStatusCodeAndMessage(response, httpStatus.OK, 200);
```
jessica-lau: I think checking either the statusCode or status in the response is enough since both of them are the same but just stated differently, so no need for both. 200 is httpStatus.OK

mocha/test/post.test.js
```diff
@@ -0,0 +1,124 @@
...
+  language: ["English"],
+  likes: [],
+  objective: "offer",
+  title: "Testers South FL Mutual Aid",
+  types: ["Information"],
+  updatedAt: new Date(),
+  visibility: "worldwide",
+};
+describe("GET /api/posts/{postId} endpoint - for a user that is NOT signed in", function () {
+  describe("GET /api/posts/{postId} endpoint - record is added to the DB and then retrived using API", function () {
```
jessica-lau: "retrieved"

mocha/utils/apiEndpoints.js
```diff
@@ -3,3 +3,6 @@ exports.feedbackEndpoint = "/api/feedback";
 exports.signUpApiEndpoint = "/api/auth/signup";
 exports.organisationsEndpoint = "/api/organisations";
 exports.postsEndpoint = "/api/posts";
+exports.postApiEndpoint = "/api/posts/{postId}";
```
jessica-lau: Don't think "/api/posts/{postId}" endpoint is being used. Remove?

jessica-lau (changes requested):
This looks great @galeza . I only got the login test fails which, I agree, should be handled by a separate PR. Also, your suggestion on refactoring the validators.js where applicable makes sense to simplify things. 

I don't know a way to bypass the different datatype for ["author.location", "author.id", "createdAt", "updatedAt", "_id"] but I can look into this when I have some time and see if I can find some sort of solution. 

One other thing -  could we possibly rename the test file from "post.test.js" to something like "postId.test.js"? There's another test called "posts.test.js" that checks posts exists in the /api/posts endpoint and the names are so close and can be confused with each other. Thanks!

mocha/utils/apiEndpoints.js
galeza: Corrected

mocha/utils/validators.js
galeza: Corrected.

mocha/utils/validators.js
galeza: Corrected.

mocha/utils/validators.js
galeza: Corrected.

mocha/test/post.test.js
galeza: Corrected.

mocha/test/post.test.js
galeza: Corrected.

mocha/utils/apiEndpoints.js
galeza: Removed

galeza:
@jessica-lau @TimeaKovacs86 
@jessica-lau I implemented your suggestions, please check.
I added also a generic validator that takes dict and validate kay-value. I removed all other methods. Please see and let me know if I miss anything.
All tests should pass.

JayashreeSamantasinghar (commented):
@galeza 
All mocha integration tests passed

TimeaKovacs86:
@galeza  LGTM! Mocha tests pass! And as we discussed before there is only one cypress test failing, and gonna be fixed in a future PR! Great job! 

backend test post endpoint mocha integration tests pass

TimeaKovacs86 (dismissed):

jessica-lau (dismissed):
LGTM @galeza 

image

JayashreeSamantasinghar:
@galeza 
I was trying other methods to compare both files without omitting some keys. 
In validator.js----> exports.validateInitialAndResponseObject = function (initial, response) {
  _.isEqual(initial, response);
};
In postId.test.js------>  validator.validateInitialAndResponseObject(
        insertResult[0],
        response.body.post
);
Can we try in this way?

galeza:
> @galeza
> I was trying other methods to compare both files without omitting some keys.
> In validator.js----> exports.validateInitialAndResponseObject = function (initial, response) {
> _.isEqual(initial, response);
> };
> In postId.test.js------> validator.validateInitialAndResponseObject(
> insertResult[0],
> response.body.post
> );
> Can we try in this way?

Hi @JayashreeSamantasinghar 
The problem is that we have more data in the response.body.post that's why it was necessary to omit them (we have additional data like status, counters).

JayashreeSamantasinghar (dismissed):

adorapham:
LGTM, i get the same result 
image

adorapham (dismissed):

PR #1871, opened 2020-11-21
#1806 - Set Maximum Character Limit User is Allowed to Enter in Feedback Responses

### LINKED ISSUE #1806: Set Maximum Character Limit User is Allowed to Enter in Feedback Responses
**Environment:** https://staging.fightpandemics.work/

**Steps to reproduce:**

- Click the feedback icon on the top right of the homepage when user is not signed in.

- Answer question on screen 1. When user reaches screen 2, enter as many characters in the response fields as the user wants. When user reaches screen 3, enter as many numbers in the age response field as the user wants. Submit feedback.

**Actual results:**

When user is on screen 2, he/she can enter as many characters in the response fields with no character limit.
When user is on screen 3, he/she can enter as many numbers in the response field with no integer limit.
Feedback can be submitted successfully.

**Expected Results:**

A max character limit should be set for the response fields on screen 2 (e.g. 1000 character limit).
A max 3 number limit (assuming there are people over age 99) should be set for the age response field on screen 3.

image

image

shidoro:

- [x] The name of this branch is: **`feature/1806-set-maximum-character-limit`**. The branch is created in a cloned version of this repo.

---

This PR fixes #1806

### DIFF
--- a/client/src/hooks/reducers/feedbackReducers.js
+++ b/client/src/hooks/reducers/feedbackReducers.js
@@ -22,7 +22,13 @@ export const initialState = {
 };
 
 export const feedbackReducer = (oldState, action) => {
-  const { type, key, value } = action;
+
+  // this is only for type number inputs
+  if (action.limit && action.value < action.limit.min) action.value = action.limit.min;
+  if (action.limit && action.value > action.limit.max) action.value = action.limit.max;
+
+  const { type, key, value} = action;
+  
   switch (type) {
     case TOGGLE_STATE:
       return { ...oldState, [key]: !oldState[key] };
@@ -35,6 +41,7 @@ export const feedbackReducer = (oldState, action) => {
 
 export const feedbackFormReducer = (state, action) => {
   const { type, ...payload } = action;
+
   switch (type) {
     case FEEDBACK_FORM_SUBMIT:
       return { ...state, loading: true, error: null };
--- a/client/src/components/Input/FormInput.js
+++ b/client/src/components/Input/FormInput.js
@@ -8,7 +8,11 @@ import Label from "./Label";
 
 const { colors } = theme;
 
-const FormInput = styled.input`
+const FormInput = styled.input.attrs(({maxLength, min, max}) => ({
+  maxLength: maxLength || Number.MAX_SAFE_INTEGER,
+  min: min || Number.MIN_SAFE_INTEGER,
+  max: max || Number.MAX_SAFE_INTEGER
+}))`
   border: none;
   box-shadow: none;
   color: ${colors.black};
--- a/client/src/templates/layouts/NavigationLayout.js
+++ b/client/src/templates/layouts/NavigationLayout.js
@@ -106,12 +106,17 @@ const NavigationLayout = (props) => {
     {
       stateKey: "mostValuableFeature",
       label: t("feedback.mostValuable"),
+      limit: 1000,
     },
     {
       stateKey: "whatWouldChange",
       label: t("feedback.oneChange"),
+      limit: 1000,
     },
-    { stateKey: "generalFeedback", label: t("feedback.otherFeedback") },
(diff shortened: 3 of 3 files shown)


rabbi08 (changes requested):
- The character limit is set as 1000: PASS
- age limit character is 3 numbers : PASS
- Not able to enter letters for age: PASS

I have following Observation: 
- The feedback provided should be reset every time user open Feedback, currently the previous provided/submitted feedback is displayed when user clicks again to provide feedback.
- Can we display the character count, as this will guide the user that how many characters left rather than just keeping them block. Refer below GIF.

ezgif com-gif-maker (66)

jessica-lau (changes requested):
-I was able to enter only up to 3 digits for the age as expected.

Issues:

-When I was entering a 3 digit age and as soon as I entered a fourth digit, the entire age switches to "999".

Screen Recording 2020-11-21 at 03 11 56 74 PM

-I was allowed to enter over 1000 characters in the feedback responses. However, the feedback form only displays the successful submission message when I enter up to 1000 characters. 

-I agree with @rabbi on adding some type of message to inform the user how many characters are left for each response. It can also be a message that appears underneath each field informing the user had reached the 1000 character limit when this limit is hit - i.e "1000 character limit reached." Also, the user shouldn't able to enter more than 1000 characters if she/he is trying to.

-Adding a message underneath the age field when the user tries to enter more than 3 integers would be helpful.

client/src/hooks/reducers/feedbackReducers.js
```diff
@@ -22,7 +22,13 @@ export const initialState = {
 };
 
 export const feedbackReducer = (oldState, action) => {
-  const { type, key, value } = action;
+
+  // this is only for type number inputs
+  if (action.limit && action.value < action.limit.min) action.value = action.limit.min;
+  if (action.limit && action.value > action.limit.max) action.value = action.limit.max;
+
```
Zaydme, line 29: Have you tried just using vanilla `max` and `maxlength` attributes for inputs without any of this? and adding type attributes for text/number inputs, or there was any issues with using that alone?

shidoro: Yes I tried, but there's a problem with the `max` attribute as it does not prevent users from typing any number, and the form is still being submitted because the submit button is in fact just an anchor tag. HTML would prevent it if it was a type submit button wrapped in an actual form element but that's not the case and given the way everything is abstracted in the code, I wouldn't know how to make such change without breaking stuff xD

shidoro:
@jessica-lau @rabbi08 about the character count, great idea but the spacing and the alignment when designing the form did not take that into account.. where do I put it so that it looks good? (making good design is not really my area if you leave it up to me it will look like... probably bad xD)

client/src/hooks/reducers/feedbackReducers.js
Zaydme: Yeah then probably just "substring" numbers lol (sounds weird), so they don't turn into 999.

Zaydme:
For the errors, just something similar to this:
image

you can turn it into a from to benefit from the validation instead of making your own.
Take a look at some of the already existing ones, like in `client/src/pages/EditProfile.js` 
```js
  const { errors, formState, register, handleSubmit } = useForm({
    mode: "change",
  });
....
            <FormInput
              inputTitle={t("profile.individual.intro")}
              name="about"
              type="text"
              defaultValue={about}
              error={errors.about}
              ref={register({
                maxLength: {
                  value: ABOUT_MAX_LENGTH,
                  message: t("profile.common.maxCharacters", {
                    maxNum: ABOUT_MAX_LENGTH,
                  }),
                },
              })}
            />
```

rabbi08:
This issue is resolved in PR #2040 
Hence, closing this please re-open if issue is observed again. Thanks

Issue #1722, opened 2020-10-30
Enable Multiple Auth Providers


joshmorel:
We need an ability to use multiple authentication providers - Auth0, Silkey, mock for local, maybe more.

We need to decouple some of our user/authentication logic from Auth0.

joshmorel:
FYI @ddveloper there is also this which is related #1495 all as part of this work we want to:

1) Support multiple auth providers (this ticket)
2) Create a mock dev provider for local dev & testing
3) Seed some dummy data for ease of development

We can also split up some of this work.

ddveloper:
sorry @joshmorel I commented in the PR #1872  , here is the latest status from my side:
```
fastify-passport: need to upgrade fastify to 3.0+, causing many other package issue to fix (haven't got a point to fix all and stopped there)
Oauth protocol with Nodejs: able to login with Facebook
fastify-oauth2: a potential package we can use helping us to save some efforts, to be experimented
```
all the experiment I did were authentication related only, no database schema work yet
will be a little slow on this progress next week due to personal appointments

PR #1872, opened 2020-11-21
#1722 - Support Multiple Auth Providers

joshmorel:
This is very much a draft for now as so far I have only updated the User model. I guess you will take the next steps @ddveloper ?

We are no longer relying for auth0 on generating the mongo_id on auth0 side and then using that as the `_id` of the user in our DB. 

Instead the `authId` is a compound of `provider|idFromProvider`.

We anticipate 3 providers in the near future but only 2 as part of this work:

* `auth0`: Our current provider. The `idFromProvider` id will be `sub` from JWT
* `silkey`: To do later. The `idFromProvider` will be a wallet address but might be made available via `sub` similar to `auth0`. TBD
* `email` or `local` or `mock`: Email & password for mock auth (testing & local development). But we do have an admin@fightpandemics which may need to use this as well. `password` is new option field in `User` to support this. There might however, be a different/better way to support these so this one is open to a different approach.

We have to change all the `auth.js` files to not be so deeply coupled with Auth0. How we determine the provider will be part of this work.

Since there is a [fastify passport library](https://github.com/fastify/fastify-passport ) now maybe we can leverage this to reduce amount of code we need to write. 

There are two strategies to consider:

* http://www.passportjs.org/packages/passport-local/
* http://www.passportjs.org/packages/passport-auth0/

@Naraujo13 I wouldn't mind feedback on the update to the User model and index. It is a unique index for quick look up of user, but it requires to be part of it `_id` as `authId` will be undefined for organizations.


Naraujo13:
Hi @joshmorel,

> @Naraujo13 I wouldn't mind feedback on the update to the User model and index. It is a unique index for quick look up of user, but it requires to be part of it _id as authId will be undefined for organizations.

I do agree with the approach to `authId` being now `provider|idFromProvider` and generating the ObjectId directly in Mongo (we probably should have done that since the beginning tbh). 

With that in mind, I get that we may need an index in `authId`, probably with a unique setup, for looking up the user based on their auth. I didn't understand about including the `_id` in the same index. 
Could you elaborate a bit more on this part?

joshmorel:
> With that in mind, I get that we may need an index in authId, probably with a unique setup, for looking up the user based on their auth. I didn't understand about including the _id in the same index.
> Could you elaborate a bit more on this part?

Thanks @Naraujo13 . This is because of the differentiators, and that the organization is using the same collection but will not have an `authId`. Or at least a real one... We could for now copy `_id` so there's something unique there on creation if that would be better than the compound index?

Naraujo13:
> > With that in mind, I get that we may need an index in authId, probably with a unique setup, for looking up the user based on their auth. I didn't understand about including the _id in the same index.
> > Could you elaborate a bit more on this part?
> 
> Thanks @Naraujo13 . This is because of the differentiators, and that the organization is using the same collection but will not have an `authId`. Or at least a real one... We could for now copy `_id` so there's something unique there on creation if that would be better than the compound index?

Okay, but why is not having `auth_id` a problem for orgs? If we create it and some docs don't have it, they just won't be present in the index (what is not a problem for returning the rest of them).
Is there a case we want to return organizations when searching by it?

I mean, having a `auth_id` index, for cases where we want to get the user authenticated, and a separate `_id` index (which is there by default, since it is the PK), for regular cases. Wouldn't this be enough for the use case?
I think I'm not getting the right use case here.

joshmorel:
@Naraujo13 
We can make it a non-unique index. I was thinking the compound might be worthwhile as a fail safe against duplicate users for the same auth id. But we can handle that in the backend logic.

Naraujo13:
> @Naraujo13
> We can make it a non-unique index. I was thinking the compound might be worthwhile as a fail safe against duplicate users for the same auth id. But we can handle that in the backend logic.

I think a unique index on only `auth_id` should work even with the null values and serve as safeguard agains duplicate users with the same auth.

ddveloper:
some thoughts: 1) there seems could be [duplicates of auth0 ID](https://community.auth0.com/t/users-with-duplicate-auth0-ids/50149/3), 2) our frontend is heavily depending on the mongoId currently I believe, 3) frontend doesn't use authId, we should prevent it from sending back to frontend

about design: do we renew the whole _id mechanism or trying to be compatible with current records?

ddveloper:
Sorry for my slow progress on this, I will be around and assist if any subtask you want me to work on

ddveloper:
@joshmorel when I try the [fastify passport library](https://github.com/fastify/fastify-passport) , it needs to upgrade our `fastify` library to be 3.0 + (currently is 2.15.1), and once I upgraded fastify, there will be other libraries need to be upgrade afterward (I didn't get a point to fully resolve all the errors yet, and I feel maybe this way causing too much changes?)

for another issue #2068 , instead of talking to `Auth0` or `passport` I experimented with social accounts API directly following Oauth2.0 flow, currently using Facebook API I can get access_token and user info after that (I believe that solution can be applied to this ticket also)

PR #1883, opened 2020-11-22
refactor geo facilities backend & fix 2 issues

joshmorel:
@maskowe these are the improvements I was suggesting. Many relate with code style but there are some other issues I found.

1. This was there from before for `location-reverse-geocode` as well (mea culpa :sweat_smile: ). We're not accepting latitude for the entire southern hemisphere!
2. The prior code wasn't handling the case where a destination in distancematrix returns with no `distance` (instead status: "ZERO_RESULTS"). Below is an example prior output for  `lat=12.704977&lng=106.127937&type=hospital`
3. This one impacts performance only. I changed the function `getFacilityDetails` to only call the GMaps API once per place. This way we can then use `Promise.all` for the `places` array, which is more performant than looping through the places and `await`ing each details result call. On my comp at least I had ~500-800ms instead of ~1500-200ms.

The rest of the changes relate to code style & organization. I applied the prettier/eslint but also:

1. Removed function Place, as discussed it wasn't adding much except adding `type` attribute which I did a different way 
2. Consistently use async/await (no `then`). 
3. Handle passing of params consistently (`getNearbyPlacesByType` was different then the rest, I also removed the `maxPlaces` param since it is also going to be the constant `MAX_FACILITIES`).
4. `FACILITY_RADIUS_BY_TYPE` as object instead of `FACILITY_RADIUS` array of objects, so to get the radius it's just a key look-up (not `find`).


maskowe:
Thanks for these changes, Josh!   I have no problem with any of them, so should I click the "Confirm merge" or should you do that?  Will that merge go into Staging env?

PR #1899, opened 2020-11-23
Post Moderation: Internal dashboard

### LINKED ISSUE #1882: Post Moderation: back-end
**Post moderation backend engineering**

Feature Summary: https://www.notion.so/fightpandemics/0562fe39e5364561a0e12f76f19f8996?v=f9b57e5e8a3448e0b63bc4154b38ddf5&p=66eab44f36204ebfbe39d7414eb11422

We want to ensure the help board contains relevant and helpful posts and comments. In instances where a post or comment is created on the help board that is in violation of our terms, contains inappropriate content, is irrelevant or considered spam, etc., we want to empower users to report these posts or comments for removal from the help board.

We also want the ability to monitor and moderate what gets reported, and this will require us to implement some moderator tools.

Figma Designs for post moderation: https://www.figma.com/file/xRDJLUsuMg3sqXA2tQEADv/Post-Moderation?node-id=5%3A0

_**Keep in mind this feature is for Web, Responsive, and Native.**_

1# Data:
Updating the User model to include:
+ permissions. (using bitfields, or just booleans)
Updating the Post model to include
+ status: String, enum "public", "reported", "removed".
+ reportedBy: Array of "mini" user objects with the reason of report (username: String, id: ObjectId ref: "User", reason: String), used to notify users of the action taken, and to know how much the post was reported, and who reported, and to hide the post from those users feed.
A post with status "public" will be visible for everyone, a post with status "reported" will be hidden from reportedBy but visible to everyone else.
When a report is rejected, the post status will be back to "public", if the report was accepted, the post status will become "removed" and will not be in the feed for everyone.
After the moderator takes a decision, all users reportedBy will receive a notification.
posts with status "public" but with reportedBy != null fall into the category of "Rejected reports"
I suggest that we have a new model, for the audit log. all actions taken by moderators will be logged in the audit log.
+ postId: ObjectId, ref "Post"
+ moderatorId: ObjectId, ref "User"
+ action: String, type of action, "accepted", "rejected"
+ justification: String, the moderator's justification for the action. (to be sent in the notification to reported and the reporter users.) and as future proof.
+ timestamps.
The audit log will give us a lot of metrics, how many actions were taken per each moderator, and the type of those actions.

2# backend:
permission management system, each fastify req will have a permission level, which we get from the User model.
moderation endpoints will require a certain perm level, and admin endpoints will require a higher perm level.
not logged in: perm level 0
logged in: perm level 1
moderator: perm level 2
Admin: perm level 3
notifications should be simple since notifications are soon merged into STG
The audit logger should be similar to the notifier fastify plugin we made for notifications.

3# endpoints:
POST api/reports/:postId , accessible to perm level 1 (loggedin user). create a new report
PATCH api/reports/:postId/:action , accessible to perm level 2 (mod). action can be "accept" or "deny",
GET api/reports/ , accessible to perm level 2 (mod). to list reports on the dashboard, this can have filters/search queries
GET api/reports/log/ , accessible to perm level 3 (Admin). to list the audit log, this also can have filters/search queries. (READ ONLY)

### LINKED ISSUE #1943: Post moderation: Internal dashboard
User story:
As a moderator, I should be able to:
- View reported posts in dashboard, prioritized by number of reports.
- Have the option to remove reported post or keep post.

Designs:
https://www.figma.com/file/EMKcg0eTnOlr7bRZGFdnZu/Web-app-product-spec?node-id=88%3A2
image
image
image

Success criteria:
After I click remove post/keep post, corresponding action is performed and post snippet is moved to closed reports section. 

Note: 
This feature is only FP internal at the moment - UI can be just functional. 
Designers will add more mockups in the meanwhile to depict other actions.
Ensure that you set the base branch of your pull requests to feature/postmoderation

Zaydme:
Adds internal moderation dashboard.

closes #1882
closes #1943


rabbi08:
@Zaydme @OzkanAbdullahoglu 

Observed few more issues while testing for Post -moderation:
tested on mobile/iPhone, working fine no UI issue observed.

1. UI issue on profile page (same as it was on help board) Refer below screenshot - Not fixed
Screen Shot 2020-11-29 at 11 07 57 PM

2. The button name in 'Delete' pop up is changed, it should be 'Delete' only, currently it is displayed as 'Delete your post' refer below screenshot.
Screen Shot 2020-11-29 at 11 18 21 PM

3. The  names for Edit and Delete options needs to be updated as suggested in below screenshot.

image

OzkanAbdullahoglu:
Hi @rabbi08 ,

I fixed the texts as below, 

delete2
deletePost1

I couldn't reproduce the first issue, can you elaborate?

OzkanAbdullahoglu:
Okay. Can you try again? @rabbi08 

chrome_2020-11-30_16-26-15

rabbi08:
@OzkanAbdullahoglu 
Verified the issues mentioned above. They are working fine, Thanks. 
Please see below GIF from mobile and screenshot.

ezgif com-video-to-gif 2
Screen Shot 2020-11-30 at 1 08 41 PM

rabbi08:
@Zaydme I have following observations for during initial testing of post moderation dashboard:

1. After we decide to 'Keep Post' -> the post is moved to 'Kept Post' 
Issue: It should also be visible to the users like normal post on the Help Board.
Currently it is NOT visible to the user who reported it even after 'Keep Post' action.

Refer below GIF.
ezgif com-video-to-gif

2. When user try to login using social media account -> the page direct to create profile page and then automatically direct the user to FP home page without login. I tried this scenario in STG, it is working fine but not here.
Please see GIF.
ezgif com-video-to-gif-2

While using -> Google -> it navigate to create profile page -> clear the profile page -> do nothing
When I click on FP logo, and I land on FP page, I am logged in using google login

3. when there is NO search results to display, the message is displayed as "There are no "Pending" reports."
It should be something like "There are currently no matching posts available." as we have used word 'Post' with our labels in the dashboard.
image

4. Navigate to 'Audit Log' -> click on 'View Post' for any post -> Click on 'BAck' button 
Issue: Users lands on Help Board page, it should be navigated back to the page from where he navigate i.e. 'Audit log' page.

5. On Dashboard -> user is able to view the post by click on the post description, even if they are read only.

As a 'Reader' ->
1. Following options are not displayed: PASS
- Audit Logs: Not displayed
- Manager Users: Not displayed

2. Can Only view cannot perform 'Keep' or 'Remove' actions: PASS

Note: Reader can have post view rights.

As a 'Moderator' ->
1. Following options are not displayed: PASS
- Audit Logs: Not displayed
- Manager Users: Not displayed

2. Can 'Keep post'
3. Can 'Remove' and 'Restore' post
4. After 'Restore' from 'Remove' the post is moved from "Removed -> Kept" : PASS

As a 'Inspector' ->
1. Following options are not displayed: PASS
- Manager Users: Not displayed

2. Can NOT perform 'Keep post', 'Remove' and 'Restore' post: PASS
3. Can View 'Audit Logs'.

As a 'Administrator' 
- Can view all options and perform all the actions
- Cannot edit other 'Admin' users: PASS

Zaydme:
Thanks @rabbi08 

**1**: We agreed that if a user reported a post, they should never see it again, even if a moderator kept it. If a user reported a post, they definitely don't want to see it again, they either think it is offensive or something.

**2/3/4** I will look into them.

**5**: is expected behavior, anyone with access to the dashboard ("reader" or above) has the right to see reported and removed posts, even if they are "Read only". "Read only" means that you can't take action, but you can always "Read"

backend/lib/endpoints/reports.js
```diff
@@ -40,15 +48,599 @@ async function routes(app) {
...
+        throw app.httpErrors.internalServerError();
+      }
+
+      if (actionProps.action === "accept") {
+        // action, post, actorId (moderator), authUserId (moderator), details (justification)
+        app.notifier.notify(
+          "report",
+          updatedPost,
+          actionProps.moderatorId,
+          actionProps.moderatorId,
```
ddveloper, line 227: just want to make sure the 226, 227 are same is not a typo

backend/lib/endpoints/reports.js
```diff
@@ -40,15 +48,599 @@ async function routes(app) {
...
+        },
+        { $group: { _id: null, count: { $sum: 1 } } },
+      ]);
+      /* eslint-enable sort-keys */
+
+      const [postsErr, posts] = await app.to(
+        Post.aggregate(aggregationPipelineResults).then((posts) => {
+          posts.forEach((post) => {
+            post.elapsedTimeText = {
+              created: translateISOtoRelativeTime(post.createdAt),
```
ddveloper, line 143: might not be a relevant comment, this `translateISOtoRelativeTime` seems obsoleted in the Frontend (I don't see any usage other than the definition), I think keeping backend with utc time and leave the Frontend to do the translate is a better idea 🤷‍♂️

backend/lib/endpoints/reports.js
```diff
@@ -40,15 +48,599 @@ async function routes(app) {
...
+                  _id: null,
+                  count: { $sum: 1 },
+                },
+              },
+            ],
+            dailyNewReportsCounts: [
+              {
+                $match: {
+                  status: "flagged",
+                  createdAt: { $gt: new Date(Date.now() - 2.678e9) },
```
ddveloper: need a comment for the number

backend/lib/endpoints/reports.js
```diff
@@ -40,15 +48,599 @@ async function routes(app) {
...
+                $match: {
+                  status: "public",
+                  reportedBy: {
+                    $exists: true,
+                    $not: { $size: 0 },
+                  },
+                },
+              },
+              {
+                $unwind: "$reportedBy",
```
ddveloper, line 421: would this (and following `unwind`) be expensive? @Naraujo13 ?

backend/lib/endpoints/schema/reports.js
```diff
@@ -1,15 +1,38 @@
 const S = require("fluent-schema");
 const { strictSchema } = require("./utils");
+const { POST_STATUS } = require("../../models/Post");
+const { MODERATOR_ACTIONS } = require("../../models/Audit");
 
 const createPostReportSchema = {
-  body: strictSchema().prop("reason", S.string().required()),
+  body: strictSchema().prop("reason", S.string().required().maxLength(500)),
+  params: strictSchema().prop("postId", S.string().required()),
 };
```
ddveloper, line 9: it's better to add `description` and `response` codes for new endpoints, refer to @jessica-lau 's [commit for example](https://github.com/FightPandemics/FightPandemics/pull/2049/commits/131184bb19b490a8781b4bae69133dfa44a5fad7) 😁

backend/lib/endpoints/schema/users.js
```diff
@@ -78,6 +78,11 @@ const getUserByIdSchema = {
   params: strictSchema().prop("userId", S.string().required()),
 };
 
+const setUserPermissionsSchema = {
+  body: strictSchema().prop("role", S.string().maxLength(100)),
+  params: strictSchema().prop("userId", S.string().required()),
+};
+
```
ddveloper, line 86: same as above comment for `description` and `response` codes

client/src/components/DashBoard/AdminDashboard.js
```diff
@@ -0,0 +1,137 @@
...
+
+const { Option } = Select;
+const { Search } = Input;
+
+const AdDashboard = styled(Table)`
+  padding: 1rem;
+  border-radius: 0.2rem;
+  .delete {
+    font-weight: 700;
+    color: ${RED};
```
ddveloper: minor: the styling seems not consistent in our project, e.g. do a global search for `color:`

client/src/components/DashBoard/AdminDashboard.js
```diff
@@ -0,0 +1,137 @@
...
+    } catch (e) {
+      console.log(e);
+      alert("Something went wrong.");
+    }
+  };
+
+  //we get the id from the input here
+  const onSearch = async (value) => {
+    if (!value) return;
+    if (!/^(?=[a-f\d]{24}$)(\d+[a-f]|[a-f]+\d)/i.test(value))
```
ddveloper, line 49: add comment for regex?

client/src/components/DashBoard/AuditLog.js
```diff
@@ -0,0 +1,101 @@
...
+        {justification
+          .replace(/[^|]*$/, "")
+          .split("|")
+          .filter((e) => e)
+          .map((reason, idx) => (
+            <FilterTag key={idx} disabled={true} selected={false}>
+              {reason}
+            </FilterTag>
+          ))}
+        <p>{(justification.match(/([^\|]+$)/) || [])[0]}</p>
```
ddveloper, line 55: add comment for regex?

client/src/components/DashBoard/PostCard.js
```diff
@@ -0,0 +1,237 @@
...
+      }
+    }
+
+    &::before {
+      content: normal !important;
+    }
+
+    .highlighted {
+      font-weight: bold;
+      color: #425af2;
```
ddveloper: warning 🚨 hardcoded color 🎨

client/src/components/DashBoard/Stats.js
```diff
@@ -0,0 +1,231 @@
...
+            data={{
+              datasets: [
+                {
+                  data: [
+                    stats.pendingPostsCount[0].count,
+                    stats.keptPostsCount[0].count,
+                    stats.removedPostsCount[0].count,
+                  ],
+                  backgroundColor: ["#2888d155", "#ff220055", "#11ff0055"],
+                  borderColor: ["#2888d1", "#ff2200", "#11ff00"],
```
ddveloper, line 69: 🎨 ( minor )

client/src/pages/DashBoard.js
```diff
@@ -0,0 +1,484 @@
...
+      } else {
+        return Promise.resolve();
+      }
+    },
+    [feedPosts.length, isLoading, loadMore], // eslint-disable-line react-hooks/exhaustive-deps
+  );
+
+  useEffect(() => {
+    setItemCount(loadMore ? feedPosts.length + 1 : feedPosts.length);
+  }, [feedPosts.length, loadMore]);
```
ddveloper, line 367: miss one? `// eslint-disable-line react-hooks/exhaustive-deps`

ddveloper (commented):
I didn't review css part, GREAT WORK!!

backend/lib/endpoints/reports.js
ddveloper: maybe simply remove the frontend duplicated code if that is not used

client/src/components/DashBoard/AdminDashboard.js
OzkanAbdullahoglu: yeap minor; should be pulled from theme

client/src/pages/DashBoard.js
OzkanAbdullahoglu: no, it doesn't need that. Dependency array is already included all dependants.

client/src/components/DashBoard/AdminDashboard.js
OzkanAbdullahoglu: I think this is a refactoring case, we have duplicate colors in front-end.  We have already a theme but at the same time, we have colors as constants. Whilst this specific issue would be fixed we should get rid of the duplicate folder and adjust its related components back to the theme.

backend/lib/endpoints/reports.js
Zaydme: Not a typo :sweat_smile: actorId (moderator) === authUserId (moderator) in this case, and the notifications plugin needs both, (normally it uses it to check if an org interacted with owner's post and vice versa)

backend/lib/endpoints/reports.js
Zaydme: Comments and posts get the time converted from the backend. but when making messaging, I had to generate the time on the front-end (to be relative to the client timezone), so the front-end code is used for messaging and notifications only, and I agree that it should be used for everything else, because it is relative to the client timezone, and not the server's one.
but since this includes changing it for comments and posts not only on the dashboard, I think it should be handled as a separate issue, to not make this PR larger than it is

backend/lib/endpoints/reports.js
Zaydme: the entire stats aggregation pipeline is very expensive, (it takes 2 seconds!) but since it is limited to a veryyyy small number of users, and will only be called once in a while, I don't think it is a concern.

backend/lib/endpoints/schema/reports.js
Zaydme: She already did in that commit, for `createPostReportSchema` https://github.com/FightPandemics/FightPandemics/pull/2049/commits/131184bb19b490a8781b4bae69133dfa44a5fad7#diff-3fbedfbc2f534dc85376a9c486d68b39bf969d0766b6914e409607ebd84d134d
the other endpoints are internal and should not be documented (?)
`getPostReportsSchema`, `getAuditLogSchema`, `moderatorActionSchema`

backend/lib/endpoints/schema/users.js
Zaydme: yeah I am not sure if this endpoint should be documented. it is for internal dashboard use.

ddveloper (dismissed):
LGTM

OzkanAbdullahoglu (commented):
LGTM!

OzkanAbdullahoglu (dismissed):

backend/lib/endpoints/reports.js
Naraujo13: Yeah, this pipeline is expensive because of the facet (since it is basically multiple pipelines to compose this one), so the `unwind` step is not something that will impact that much performance here.
Ideally, we could switch this facet to a single pipeline, I'm not that familiar with the data being used here, but it seems that if we used a two-stage grouping we would be able to get most if not all stats without the facet (first grouping by status and reported_by, getting the number of posts in each status for each reporter, and then grouping the results by statuses only again to get the total sum of each status, but keep the individual counts needed too). 

But this may be too much work if this is not going to scale that much.
Just creating the right indexes to support the queries, especially the initial matches, will make this drop a lot already.
I'd suggest the following indexes:

- For the stats aggregation:
      - `{ status: 1, created_at: 1 }`
      - `{ status: 1, reported_by: 1 }`
- For the new filter in the feeds posts: add the status to the index that already exists in the Post model to support this filtering (with the exact same order as the fields in the match, since the ordering is sensitive).

backend/lib/models/Notification.js
```diff
@@ -3,7 +3,7 @@ const { Schema, model } = require("mongoose");
 const notificationSchema = new Schema(
   {
     action: {
-      enum: ["like", "comment", "share"],
+      enum: ["like", "comment", "share", "report"],
```
mannykary, line 6: Was the Notification Lambda updated to support report? If not, we'll either need to add support for it, or filter it out if it's not a requirement. Otherwise, the Lambda will break. In the interest of time, if there is no requirement for email notifications, I would simply update the queries in `services/notification/src/helpers/database-helper.js` to filter for only supported actions. So update `_findInstantNotifications()` and `_findDigestNotifications()` to add a filter for notification actions in the `NotificationAction` enum (can probably just add `$in: Object.values(NotificationAction)`)

client/src/components/CreateReport/CreateReport.js
```diff
@@ -8,7 +8,20 @@ import { useTranslation } from "react-i18next";
...
+const CreateReport = ({
+  postId,
+  currentPost,
+  setCallReport,
+  callReport,
+  fromPage,
+  forModerator,
+  changeType,
+}) => {
+  let post;
```
mannykary: Minor: Could do `const post = currentPost || undefined;`

client/src/components/DashBoard/AdminDashboard.js
```diff
@@ -0,0 +1,139 @@
...
+          >
+            &times;
+          </a>
+        );
+      },
+    },
+  ];
+
+  return (
+    <div className="AdDashboard">
```
mannykary: Minor nitpick, I probably wouldn't shorten `Admin` to `Ad` as `Ad` is shorthand for `Advertisement`. `Admin` is already shorthand for `Administrator` so I don't see why it needs to be shortened even further.

backend/lib/endpoints/reports.js
```diff
@@ -40,15 +48,599 @@ async function routes(app) {
...
+              },
+              {
+                $set: {
+                  reasonSplitted: { $split: ["$reportedBy.reason", "|"] },
```
mannykary: I'm probably being a grammar Nazi, but the past tense of split is...[split](https://pasttenses.com/split-past-tense) (go figure).

client/src/components/DashBoard/AuditLog.js
```diff
@@ -0,0 +1,105 @@
...
+          .replace(/[^|]*$/, "")
+          .split("|")
+          .filter((e) => e)
+          .map((reason, idx) => (
+            <FilterTag key={idx} disabled={true} selected={false}>
+              {reason}
+            </FilterTag>
+          ))}
+        {
+          // Justification (and reasons) are joined in a single string by "|" as a delimiter
```
mannykary, line 52: Is there a reason for why we are storing justifications/reasons as a single string rather than an array?

client/src/components/DashBoard/Posts.js
```diff
@@ -0,0 +1,176 @@
...
+// Constants
+import { mq } from "constants/theme";
+
+const cellMeasurerCache = new CellMeasurerCache({
+  fixedWidth: true,
+  defaultHeight: 380,
+});
+
+const HorizontalRule = styled.hr`
+  display: none;
```
mannykary, line 26: Sorry, this is probably out of scope and more of a nitpick of the overall React code structure/design pattern rather than the PR specifically, but I've noticed that we are basically inlining/embedding CSS styles within the JavaScript code, but I feel like CSS code should be in their own separate files, and then loaded in. Assuming we name the CSS file the same as the component, it would be easier to separate styles from component code, and make it more maintainable and reusable.

client/src/components/DashBoard/Stats.js
```diff
@@ -0,0 +1,231 @@
...
+    return (
+      dataObject.find((e) => {
+        return e._id === moment().format("D/M");
+      })?.count || 0
+    );
+  };
+
+  const LineChartScaleMaxMin = Math.max(
+    ...getLast30DaysStats(stats.dailyNewReportsCounts),
+    ...getLast30DaysStats(stats.auditStats.rejectedReports,),
```
mannykary:
```suggestion
    ...getLast30DaysStats(stats.auditStats.rejectedReports),
```

client/src/constants/permissions.js
```diff
@@ -0,0 +1,22 @@
+const SCOPES = {
```
mannykary, line 1: Minor: we might want to think about a good way to share constants that are used by both the frontend and backend. Out of scope for this ticket obviously, but it's not very DRY nor maintainable to have the same constants defined in both places.

client/src/locales/translations/en_US.json
```diff
@@ -1038,6 +1039,18 @@
...
+    "buttonKeep": "Keep",
+    "removePostTitle": "Remove Post Confirmation",
+    "keepPostTitle": "Keep Post Confirmation",
+    "reportDetails": "Additional details provided by users:",
+    "userReported": "Users reported this post for:",
+    "removeReason": "Reason(s) for removing:",
+    "keepPostConfirm": "Are you sure you want to keep this post?",
+    "reportAcceptedTitle": "Report is accepted!",
+    "reportRejectedTitle": "Report is rejected!",
+    "reportAcceptedBody": "This post is succesfully removed!",
```
mannykary:
```suggestion
    "reportAcceptedBody": "This post was successfully removed!",
```

client/src/locales/translations/en_US.json
```diff
@@ -1038,6 +1039,18 @@
...
+    "reportRejectedBody": "This post is succesfully kept"
```
mannykary: I don't think this copy for a report being rejected makes sense. I don't have enough context for where this copy is being displayed so I'm not sure how this should be rephrased, but maybe it can just be as simple as `This post was not removed`

mannykary:
```suggestion
    "reportRejectedBody": "This post was successfully kept"
```

mannykary (changes requested):
Overall, great work! I have a few comments.

client/src/locales/translations/en_US.json
```diff
@@ -1038,6 +1039,18 @@
...
+    "youReported": "You have reported this post. It will no longer be visible on your feed",
+    "buttonRemove": "Remove",
+    "buttonKeep": "Keep",
+    "removePostTitle": "Remove Post Confirmation",
+    "keepPostTitle": "Keep Post Confirmation",
+    "reportDetails": "Additional details provided by users:",
+    "userReported": "Users reported this post for:",
+    "removeReason": "Reason(s) for removing:",
+    "keepPostConfirm": "Are you sure you want to keep this post?",
+    "reportAcceptedTitle": "Report is accepted!",
```
mannykary:
```suggestion
    "reportAcceptedTitle": "Report was accepted!",
```

client/src/locales/translations/en_US.json
```diff
@@ -1038,6 +1039,18 @@
...
+    "reportRejectedTitle": "Report is rejected!",
```
mannykary:
```suggestion
    "reportRejectedTitle": "Report was rejected!",
```

client/src/locales/translations/en_US.json
OzkanAbdullahoglu: When I wrote this phrase I felt that there should be some better phrase. But I couldn't find too. 😏

client/src/components/DashBoard/Posts.js
OzkanAbdullahoglu: Can you elaborate this a little more? We are using styled-components in front-end consistently in every component. I believe this is a decision to choose styled-components or to go with separated CSS files. And I assume that the decision is given at the very beginning of the project.

OzkanAbdullahoglu: Can you elaborate this a little more? We are using styled-components in front-end consistently in every component. I believe this is a decision to choose styled-components or to go with separated CSS files. And I assume that the decision is given at the very beginning of the project.

mannykary: ~Yeah, I'm just not seeing how that's better than a separate CSS file that can be loaded in and be able to reuse styles across multiple components. Maybe the rationale was to avoid leaking styles across components and make it easier to debug. But take my comment with a grain of salt; I am not a frontend expert.~

Actually I take that back. After reading [this](https://medium.com/@stephanieworkman/styling-in-react-best-practices-777f6fc5ed69), it seems like this is the accepted pattern for React. So feel free to ignore me 😂

backend/lib/models/Notification.js
OzkanAbdullahoglu: @Zaydme Can you take a look?

client/src/components/DashBoard/AuditLog.js
OzkanAbdullahoglu: actually, they are all in an array in the front end, when we are sending to backend we join them with a delimiter and converted to a string. @Zaydme should have a better answer I think.

OzkanAbdullahoglu (dismissed):

Zaydme:
I apologize for not being active on slack/github lately. I am just a little bit busy with school 😅😅
I will be back as soon as I can!

Zaydme:
Will be looking at this soon. Let's get this done!

OzkanAbdullahoglu:
> Will be looking at this soon. Let's get this done!

rebase or merge the branch to the staging seems like a good idea 😁

backend/lib/models/Notification.js
ddveloper: I will look into this and make some update today 😃

ddveloper: @mannykary would you take a look at my commit, to see if that's the way to filter the report out? thanks : )

maskowe (dismissed):
Found no issues.  Looks like most conversations were resolved.

OzkanAbdullahoglu (dismissed):

ddveloper:
tested with `service/postmoderation-test` [review build](https://github.com/FightPandemics/FightPandemics/pull/2158), email notification works fine, steps I took:
```
comment on a post, report the post
after 5 minutes, the author was able to receive the email notification
```
image

PR #1964, opened 2020-11-27
Added clear notifications feature and fixed mobile UI issues with notification dropdown

### LINKED ISSUE #1925: Notification: No option to 'Clear' my notifications and other UI/UX issues for notification dropdown
1. Open FP app and login using valid username and pwd.
2. Click on 'Notification' bell icon. 

Issue: You will observe that user can navigate to notification preferences but cannot CLEAR the notifications.
User should be able to clear the notification as per general user experience.

image

rz-0605:
Added clear notifications feature and fixed mobile UI issues with notification dropdown.
image

_Please be concise and link any related issue(s):_ resolves issue #1925 

Also LMK if UI needs to be changed, I based it off how I would do it because I couldn't find the dropdown in figma.


rabbi08:
@rz-0605 Thanks a lot for quick fix. This is looking great.

Verified the fix on the feature build:
I would like to request a change, that we should have 'clear' option to clear all notification at once. 

- Currently as per the new implementation, we are able to click 'x' to clear the notification one-by-one: PASS
- The scroll bar is grey: PASS
- The notifications are displayed as dropdown properly without covering the whole screen: PASS
Refer below screenshot.
httpsfeature1925-clear-notifications fightpandemics xyz

Refer below GIF
ezgif com-gif-maker (3)

client/src/components/Notifications/NotificationDropDown.js
```diff
@@ -184,6 +198,17 @@ const MenuItem = ({
   t,
   gtmId,
 }) => {
+  const { clearNotification } = useContext(WebSocketContext); 
+  const dispatch = useDispatch();
+
+  const clearLocalNotification = (notification_id) => {
```
mannykary: Since camelCase is the standard convention for JavaScript, I would change this, and associated references, to `notificationId`

client/src/components/Notifications/NotificationDropDown.js
```diff
@@ -231,8 +265,10 @@ const menu = (notifications, organisationId, t) => {
       </Menu.Item>
       <div className="notifications-container">
         <div>
-          {notifications.map((each) => (
+          {notifications.map((each, i) => (
```
mannykary: Rather than use a one letter variable name, which is hard to read, I would change this to `index`

client/src/components/Notifications/NotificationDropDown.js
```diff
@@ -284,18 +320,21 @@ export const NotificationDropDown = ({
...
-    path: `/post/${n.post.id}`,
-    actionAvatar: notificationTypes[n.action].icon,
-    createdAt: getRelativeTime(n.createdAt),
-    avatar: n.triggeredBy.photo,
-    sharedVia: n.sharedVia,
-    unread: !n.readAt,
-    gtmId: notificationTypes[n.action].gtmId,
-  }));
+  const mappedNotifications = notifications
+    .filter((n) => !n.isCleared)
```
mannykary: Like my previous comment, I would change this from `n` to `notification`

client/src/context/WebsocketContext.js
```diff
@@ -291,6 +291,10 @@ export default class SocketManager extends React.Component {
     this.socket.emit("MARK_NOTIFICATIONS_AS_READ");
   };
 
+  clearNotification = (notification_id) => {
```
mannykary: Similar to my previous comment, I would change this to `notificationId`

client/src/reducers/wsReducers.js
```diff
@@ -200,6 +201,17 @@ function wsReducer(state = initialState, action) {
           ...state.notifications.map((n) => ({ ...n, readAt: new Date() })),
         ],
       };
+    case LOCAL_NOTIFICATIONS_MARK_AS_CLEARED:
+      return {
+        ...state,
+        notifications: [
+          ...state.notifications.map((n) =>
```
mannykary: Similar to my previous comment, I would change `n` to `notification`

mannykary (changes requested):
Thanks for taking on this ticket! It looks good overall; just a few minor comments.

rz-0605:
@rabbi08 @mannykary just completed both tasks. Also, for clearing notifications I just iterate over them and individually clear them. I don't know if there is any performance improvement if we batch the updates, but let me know if you want me to change it.

mannykary:
> @rabbi08 @mannykary just completed both tasks. Also, for clearing notifications I just iterate over them and individually clear them. I don't know if there is any performance improvement if we batch the updates, but let me know if you want me to change it.

There will be a performance hit if there are a lot of notifications - the round trip for each API call to the backend to clear the notification, and also for each query to write to the database. I would add another Socket.IO listener to clear all notifications, and then use  `Notification.updateMany` to update all the user's notifications in one query. So instead of N API calls and N queries (`O(N)`), we will just have 1 API call and 1 query (`O(1)`)

mannykary:
@rz-0605 Just noticed one thing - we should add an index to the Notification model for `receiver` so that the updateMany operation is faster (since we would be updating based on the receiving user's ID). I am not sure why this was not added in the first place; it was probably an oversight on our part. (We query notifications by receiver in order to show the receiving user's notifications in the dropdown, so that query would also benefit from this index)

Also, we could potentially add an index on `isCleared` so that the query for notifications only return those set to `false`. Then we wouldn't need to filter on the client side.

@Zaydme do you have any additional thoughts on this?

client/src/components/Notifications/NotificationDropDown.js
```diff
@@ -202,14 +234,22 @@ const MenuItem = ({
           })}
         </div>
       </Content>
+      <CloseNotificationBtn
+        onClick={(e) => {
+          e.preventDefault();
+          clearNotification(_id);
+        }}
+      >
+        &times;
```
ddveloper, line 259: 👍

backend/lib/plugins/socket.js
```diff
@@ -437,6 +437,11 @@ function onSocketConnect(socket) {
     );
   });
 
+  socket.on("CLEAR_NOTIFICATION", async (data) => {
+    const { notificationId } = data;
+    await Notification.findByIdAndUpdate(notificationId, { isCleared: true })
+  });
+
```
Zaydme: For security reasons, I would also add `userId` to the query to make sure the user owns that notification.
But I think since no one has your notifications ids we can sacrifice this for performance.
What do you think @mannykary

client/src/components/Notifications/NotificationDropDown.js
```diff
@@ -258,9 +314,20 @@ export const NotificationDropDown = ({
...
+  const {
+    markNotificationsAsRead,
+    clearNotification: clearRemoteNotification,
+  } = useContext(WebSocketContext);
   const { t } = useTranslation();
+  const dispatch = useDispatch();
+
+  const clearNotification = (notificationId) => {
+    dispatch({
+      type: "LOCAL_NOTIFICATIONS_MARK_AS_CLEARED",
```
Zaydme: Please import this from the wsActions

client/src/actions/wsActions.js
```diff
@@ -26,6 +26,8 @@ export const GET_NOTIFICATIONS_SUCCESS = "GET_NOTIFICATIONS_SUCCESS";
 export const NEW_NOTIFICATION = "NEW_NOTIFICATION";
 export const LOCAL_NOTIFICATIONS_MARK_AS_READ =
   "LOCAL_NOTIFICATIONS_MARK_AS_READ";
+export const LOCAL_NOTIFICATIONS_MARK_AS_CLEARED =
+  "LOCAL_NOTIFICATIONS_MARK_AS_CLEARED";
```
Zaydme: This should be singular, since it's only clearing one notification.

client/src/components/Notifications/NotificationDropDown.js
```diff
@@ -230,9 +270,24 @@ const menu = (notifications, organisationId, t) => {
         </Link>
       </Menu.Item>
       <div className="notifications-container">
+        {notifications.length > 0 && 
+          <ClearAllNotifications>
+            <ClearAllButton
+              onClick={() => {
+                notifications.forEach((notification) => {
+                  clearNotification(notification._id);
```
Zaydme: Yeah this definitely needs to be packed into one single event, so all of this can be a single query.
People might have a lot of notifications, and this is not ideal.

Zaydme (commented):
I have left some comments.
Overall you did a great job @rz-0605

Zaydme:
Some other small UI issues
Screenshot_20201128-142833_Chrome
For the clear all button, I think it would look better next to the settings icon.
Screenshot_20201128-142843_Chrome

client/src/components/Notifications/NotificationDropDown.js
rz-0605: I noticed that other dispatches don't import and directly use the string (such as `updateReadAt`), so I'd though to keep them consistent. Would you also like me to change the other ones, or just `clearNotification`?

Zaydme: No I mean just import this string `"LOCAL_NOTIFICATIONS_MARK_AS_CLEARED" `as `LOCAL_NOTIFICATIONS_MARK_AS_CLEARED` since it is exported from wsActions, so if the string changed, we don't have to change it anywhere. but that's a very minor thing, you can ignore for now.

backend/lib/plugins/socket.js
```diff
@@ -438,8 +438,11 @@ function onSocketConnect(socket) {
   });
 
   socket.on("CLEAR_NOTIFICATION", async (data) => {
+    const { userId } = socket;
     const { notificationId } = data;
-    await Notification.findByIdAndUpdate(notificationId, { isCleared: true })
+    const notification = await Notification.findById(notificationId);
+    if(notification.receiver === userId)
```
Zaydme: Does this work? because `notification.receiver` is `ObjectId`, and `userId` is `String`, maybe you will need to add `notification.receiver.toString()`, but if you tested it and it works then that's fine.

rz-0605: it works in testing, I'm pretty sure mongo automatically converts to string when comparing. I have it that way in my own projects and it never seems to fail. But I'll add `toString` to the next commit just to be sure.

rz-0605:
@mannykary indexing is one of the things I haven't used too much because I've only used small databases. Can you check out the latest commit to see if I did it right?

client/src/context/WebsocketContext.js
```diff
@@ -291,6 +291,14 @@ export default class SocketManager extends React.Component {
     this.socket.emit("MARK_NOTIFICATIONS_AS_READ");
   };
 
+  clearNotification = (notificationId) => {
+    this.socket.emit("CLEAR_NOTIFICATION", { notificationId })
```
mannykary: Can you run Prettier on this file? Looks like there is a missing semicolon at the end of this line

backend/lib/plugins/notifier.js
```diff
@@ -56,6 +56,7 @@ class Notifier {
         photo: triggeredBy.photo,
         type: triggeredBy.type,
       },
+      isCleared: false
```
mannykary: Same here, we have a Prettier rule for adding trailing commas, so if you run Prettier on this file, you'll see it being added.

backend/lib/models/Notification.js
```diff
@@ -89,6 +93,9 @@ notificationSchema.index({
     action: "like",
   },
 });
+notificationSchema.index({
+  receiver: 1
```
mannykary: Same here for Prettier adding trailing comma

backend/lib/models/Notification.js
```diff
@@ -52,6 +52,10 @@ const notificationSchema = new Schema(
         type: String,
       },
     },
+    isCleared: {
+      type: Boolean,
+      required: true
```
mannykary: Same here for Prettier adding trailing comma

backend/lib/plugins/socket.js
```diff
@@ -437,6 +437,19 @@ function onSocketConnect(socket) {
     );
   });
 
+  socket.on("CLEAR_NOTIFICATION", async (data) => {
+    const { userId } = socket;
+    const { notificationId } = data;
+    const notification = await Notification.findById(notificationId);
```
mannykary: This can be combined in one query:

```
await Notification.findOneAndUpdate(
  {
    _id: this.mongo.base.Types.ObjectId(notificationId),
    receiver: this.mongo.base.Types.ObjectId(userId),
  },
  {
    isCleared: true,
  },
);
```

mannykary (changes requested):
Few more minor comments, but overall looking good so far!

mannykary:
> @mannykary indexing is one of the things I haven't used too much because I've only used small databases. Can you check out the latest commit to see if I did it right?

Looks like you did it right, I see the index created in the database.

mannykary (approved):
Nice, LGTM!

Zaydme:
I still have this small issue mobile, where the height of the inner div is longer than the outer div.
Screenshot_20201129-133516_Chrome

Zaydme (approved):
LGTM! Thanks @rz-0605 :raised_hands:

rabbi08 (changes requested):
@rz-0605 , would like to recommend that,

Currently the the hand tooltip is displayed on the whole notification bar, the clickable comments are 'Clear All' and 'Settings' icon, 

so please define the clickable are, by changing the icon when placed on.

Pease see below GIF.
ezgif com-video-to-gif-4

rabbi08 (approved):
Verified the fix on feature branch, it is working fine. Thanks @rz-0605 

Refer below GIF
ezgif com-video-to-gif-5

Issue #1981, opened 2020-11-29
Add Lambda to process avatar delete requests


CamilaArias:
_This is an extension of #1967 to allow the user to delete their avatar. The initial fix will only clear their photo in our db._

When a user deletes their avatar, a request should be created and sent to an SQS queue for a Lambda process to handle asynchronously, effectively removing the pictures from S3.

mannykary:
Example code for Lambda (I haven't tested this code, but I have experimented with the API to see what the response looks like). This will obviously need to be refactored slightly to follow best practices/design patterns.

```
const AWS = require("aws-sdk");
AWS.config.update({ accessKeyId: 'abc', secretAccessKey: 'def' });
const s3 = new AWS.S3();

async function listByPrefix(Bucket, Prefix) {
  const params = {
    Bucket,
    Prefix,
  };
  return s3.listObjects(params).promise();
}

async function deleteObjects(Bucket, keys) {
  const params = {
    Bucket,
    Delete: {
      Objects: keys.map((Key) => { Key }),
    },
  };
  return s3.deleteObjects(params).promise();
}

async function deleteUserAvatars(id) {
  return _deleteAvatars("users", id);
}

async function deleteOrgAvatars(id) {
  return _deleteAvatars("organisations", id);
}

async function _deleteAvatars(userType, id) {
  const prefix = `content/avatars/${userType}/${id}-`;
  const response = await listByPrefix(s3Bucket, prefix);
  const files = response.Contents;
  const keys = files.map((file) => file.Key);
  return deleteObjects(s3Bucket, keys);
}
```

ddveloper:
I will copy it to our backend CDN folder, thanks @mannykary

mannykary:
> I will copy it to our backend CDN folder, thanks @mannykary 

To be clear, this would go in a Lambda. It would be blocking otherwise.

ddveloper:
@mannykary , this is what I just committed: [(commit in 1995)](https://github.com/FightPandemics/FightPandemics/pull/1995/commits/82fde9d5c15f253169458216c17458051738ffee)

mannykary:
Sorry, I think maybe there's a breakdown in communication? We will not be implementing deleting the images from S3 in the web backend because it's going to be blocking the thread and if the user has a lot of images it will be very slow. That's why I suggested just deleting the photo value from the user for now and we can implement actually deleting the image files later.

ddveloper:
[...]

PR #1995, opened 2020-11-30
Issue #1967 - resolution

### LINKED ISSUE #1967: User should be allowed to remove profile picture and replace it with name initials
User should be allowed to remove profile picture and replace it with name initials.

Actual: Once user uploads a picture he can only replace it another picture.

Expected: He should be allowed to replace with name initials

### Added details to solve:
Figma Design for modal: https://www.figma.com/file/EMKcg0eTnOlr7bRZGFdnZu/Web-app-product-spec?node-id=303%3A66

If user doesn't have an avatar, clicking on the camera icon should still open the modal for choosing an image.

If user has an avatar, clicking on the camera icon should open the Edit Avatar modal.
`Upload New` should open the modal for choosing an image which will change the avatar.
`Remove` should clear the avatar and show the user's initials again.

When removing, the `photo` value on the user's record should be cleared in our db which will require a `delete` to be added to user and org profiles.

Tororoi:
Closes #1967 

Profile Picture can be removed. I left the styling pretty much the same, just focused on getting the functionality complete.

When a user clicks the camera button on their avatar:
>if no avatar, goes straight to file select
>if avatar exists, pulls up modal that shows current avatar with buttons to remove avatar or upload new
 **Those buttons will need localization**

When a user clicks remove avatar:
>Avatar is removed, showing initials instead.

When a user clicks upload new:
> File select opens and upon selecting a file, the modal reappears with crop option
>Modal now has the buttons, upload new and submit

When a user clicks the x in the corner of the modal:
> cancels edit avatar modal and resets photoURL


backend/lib/components/CDN/index.js
```diff
@@ -12,6 +12,14 @@ async function uploadUserAvatar(id, file) {
   return _uploadAvatar("users", id, file);
 }
 
+async function deleteOrgAvatars(id) {
```
mannykary: We don't want to call S3 to delete the images directly because if a user has a lot of images, it will be slow and will block the thread, so it will result in a bad user experience. Instead, to make it asynchronous, we should put a message onto an SQS queue with the user ID or org ID for which we want to delete avatar images, and then the SQS messages would be processed by a Lambda to delete the images. So this code would have to move to a Lambda.

I discussed this with @CamilaArias in a Slack DM the other day and was under the impression that it was sufficient to just delete the `photo` attribute from the user and org for now, and implement the delete avatar Lambda later.

mannykary: An alternative solution is rather than keep all the avatar images that were uploaded by a particular user, when the user updates their avatar photo, we should just delete the currently used avatar photo in S3 and then upload the new avatar photo. This way we would only have one avatar image stored per user in S3. I don't see a good reason for keeping all their old avatar images in S3 - the reason why we did that in the first place was we needed to bust the CDN cache by appending a timestamp to the image file name (the image file name is simply `${userId}-${unixTimestamp}`). Otherwise, the avatar image would not appear to change on the frontend because the CDN is still serving the old image file. 

This should be easy to implement by taking the image file name currently stored in the `photo` field, and then delete that file from S3.

mannykary: After discussing with @ddveloper in #1981, we decided that we should revert this code for deleting from S3 since this is outside the scope of #1967, and we will implement it later.

ddveloper: Thanks @mannykary 👍 super clear

ddveloper:
Hi @Tororoi , thanks for this nice dialog, @CamilaArias please help review to see anything to change, 🎨 
(maybe change `Add New` to `Change`? make `Remove Avatar` button red?)

sheeba1:
Just a thought, wouldn't it be nice to ask for confirmation before removing the  avatar /profile picture? thanks!

rabbi08 (dismissed):
@Tororoi Thanks for the fix. 
This is working as expected,
1. Upload new avatar -: PASS
2. Remove existing avatar -> PASS
this revert the user profile logo as 'users's initials' : PASS
3. User can cancel the avatar by clicking away from the pop up.


Great If we can have following add-ons:
- Remove Avatar confirmation, with message:
"Are you sure you want to remove your avatar? it will be reverted to your initials." Ok/Cancel
Ok: Remove
Cancel: cancel the removal process

client/src/components/Picture/UploadPic.js
```diff
@@ -16,7 +16,7 @@ const CustomSubmitButton = styled(BaseButton)`
   color: ${colors.white};
 `;
 
-const CustomCancelButton = styled(BaseButton)``;
+const CustomUploadButton = styled(BaseButton)``;
```
Zaydme: We can just use the BaseButton if no additional styling is needed

client/src/components/Picture/UploadPic.js
```diff
@@ -140,7 +140,41 @@ const UploadPic = ({ cameraIconSize, gtmPrefix, user }) => {
...
+    let removeResponse;
+    let endPoint = "/api/users/current/avatar";
+    if (user.ownerId) {
+      endPoint = `/api/organisations/${user._id}/avatar`;
+    }
+    try {
+      removeResponse = await axios({
+        method: "delete",
+        url: endPoint,
+        // data: formData,
```
Zaydme: Remove comment

client/src/components/Picture/UploadPic.js
```diff
@@ -154,49 +188,77 @@ const UploadPic = ({ cameraIconSize, gtmPrefix, user }) => {
 
   const cropModal = () => {
     return (
-      <Modal
+      <AvatarModal
+        title={
+          <p>
+            {/* needs localization */}
+            {t("Edit Avatar")}
```
Zaydme: Needs localization

client/src/components/Picture/UploadPic.js
```diff
@@ -154,49 +188,77 @@ const UploadPic = ({ cameraIconSize, gtmPrefix, user }) => {
...
+          </p>
+        }
+        onCancel={closeModal}
+        // onClick={(e) => console.log(e.target)}
```
Zaydme: Unnecessary comment

client/src/components/Picture/UploadPic.js
```diff
@@ -154,49 +188,77 @@ const UploadPic = ({ cameraIconSize, gtmPrefix, user }) => {
...
-          ),
+          user && user.photo && !photoURL ? (
+            <CustomUploadButton key="remove" onClick={removePhoto}>
+              {/* needs localization */}
+              {t("Remove Avatar")}
+            </CustomUploadButton>
+          ) : null,
+          <CustomUploadButton key="change" onClick={() => imgUpload.current.click()}>
+            {/* needs localization */}
+            {t("Upload New")}
```
Zaydme: Localization

client/src/components/Picture/UploadPic.js
```diff
@@ -154,49 +188,77 @@ const UploadPic = ({ cameraIconSize, gtmPrefix, user }) => {
...
-            </CustomSubmitButton>
-          ) : (
-            <CustomSubmitButton key="retry" onClick={retry}>
-              {t("avatar.tryAgainBtn")}
-            </CustomSubmitButton>
-          ),
+          user && user.photo && !photoURL ? (
+            <CustomUploadButton key="remove" onClick={removePhoto}>
+              {/* needs localization */}
+              {t("Remove Avatar")}
```
Zaydme: Localization

Zaydme (changes requested):
LGTM, just needs localization.

Zaydme (approved):
Thanks a lot @Tororoi and @ddveloper!

rabbi08 (approved):
Tested on Desktop and mobile, working fine. Thanks

Please see below GIF
ezgif com-video-to-gif 11

PR #2024, opened 2020-12-02
Minor UI Accessibility improvements

CamilaArias:
Changing some colours to pass accessibility.
Buttons from about us page on desktop fit translations (Not changing mobile because we can't make the button longer and we should really change the words on the buttons to be shorter anyway.)

Note: I'm changing DARKISH_GRAY because I added it two days ago for improving the contrast of some text for accessibility purposes, but it's not good enough for our footer background so I tweaked it a tiny bit to use it there as well and I'm aware of the other places where this colour is used because I'm the only one who's used it so far, visually it looks exactly the same. Hope that's ok :)


Zaydme (approved):
Tested locally, Looks good to me!

ddveloper (approved):

PR #2032, opened 2020-12-03
About us changes

Zaydme:
1. remove "Who is this for?" - the text is overlapping with the image, and the image is pretty straightforward, so no need of the text
2. spacing between headings not consistent - like our supporters, our community partners etc are not following the same spacing at the top and bottom
3. community partner logo too big - LA helping hands, not same as the support partners logo. we might add more in future, so better to have them the same size
4. community partner logo redirect to their org profile - is this possible to do before launch? if so, then we can make those changes, if not, post launch priority
5. footer - Copyright text is sticking to the top of the footer and the about us line is sticking to the bottom of the footer, should be center aligned horizontally and verticallly


CamilaArias (dismissed):

ddveloper (dismissed):
LGTM, though I see "Who is this for" for mobile view, I don't think that matters

CamilaArias (approved):

rabbi08 (approved):

PR #2037, opened 2020-12-04
add images for event Product Hunt

ddveloper:
for event Product Hunt images


ddveloper:
for event, please revert if not the right way

PR #2054, opened 2020-12-07
added zendesk widget

CamilaArias:
Adding Zendesk widget on all pages, we can edit from Zendesk.


ddveloper:
I see the button changed to circle for `createPost` but not for the `support` one when switching mode from PC to mobile, I don't know if user would ever have this user case (maybe not), just a comment

rabbi08 (changes requested):
@CamilaArias 
1. the application/page url is exposed along with the message to Zendesk, from where the widget is opened and submitted.
For e.g. when user click the 'Support' while being in the profile page, the user's profile is shared along with the ticket. This is against user's privacy as they don't know if their profile url being shared without their information.

Screen Shot 2020-12-15 at 7 40 58 PM

2. Also, the email subject is user driven, not as per the selected email option by the user. For e.g if user select 'Delete my account' and send email, the subject is not 'Delete email', user's have to type(no significance of these options for the user. The emails are not getting categorized on the basis of the selected issue in the widget. 
We should populate the subject with the selected option.

image

rabbi08:
On Mobile (iOS):
1. Can we have only 'Questionmark' in the circle rather than white circle around it , then it will look in sync with 'Create post' icon. See below screenshot.

IMG_B56CEB75B7D2-1

2. On some page this icon is not looking well placed as in below screenshot .
Can we place this icon on the side just like 'accessibility' icon?

IMG_BBFA11D10D86-1

CamilaArias:
@ddveloper when you change to mobile from the browser you have to refresh

CamilaArias:
@rabbi08 we can't change the URL that is being shared, you can't tie it to their profile because they could've accessed it from anyone's profile. Also we ask for their name to submit the ticket, so they know they're sending their information.

We also can't change the icon.

CamilaArias:
What each form does should be changed from Zendesk.

CamilaArias:
We can put the support icon behind the filters menu

rabbi08:
> @rabbi08 we can't change the URL that is being shared, you can't tie it to their profile because they could've accessed it from anyone's profile. Also we ask for their name to submit the ticket, so they know they're sending their information.
> 
> We also can't change the icon.

I am saying that Zendesk is doing the tie of the URL with the ticket, as shown in the screenshot. Agree that user can click 'Support' from anywhere, that's the issue, we/as Zendesk are appending the page URL from which the support is clicked and submitted. It can be their own profile URL, anyone else, about us page, Feed page etc. Issue, is the end user don't know this.
Its happening at the backend and we can see that in Zendesk 

Also, if user clicks on support email, do user information is shared with email but through widget it is.

Zaydme (approved):
LGTM! thank you @CamilaArias!

rabbi08 (approved):
- Verified on mobile, this looks fine as it is hidden behind the filters: PASS
- User's details are not being from profile page as we no longer provide 'support' icon on user profile page: PASS

Thanks @CamilaArias

PR #2107, opened 2020-12-23
feat: update profile menu

### LINKED ISSUE #2085: After adding new ORG-> click 'view my profile' -> unable to edit the profile -> because the selected profile is still individual user profile.
1. login using valid email/pwd
2. Click on profile -> add new ORG
3. Enter the required details on 'Create ORG profile' page -> click 'Create profile'
4. after the profile created on the next page, click 'View my profile'

Issue: You will observe that the profile page of the newly created ORG is displayed, as expected.
BUT, user cannot EDIT it because the selected profile is the 'Individual' user profile.

So, either we have to change the message of 'View "my" profile' with 'View "Organisation" profile' so that user should know, that they are in their ORG profile so that's why no EDIT rights, unless they switch account to ORG account.

OR we can switch the selected profile to the newly created ORG profile, so that user should be able to edit the profile after creation .

Refer below GIF.
ezgif com-video-to-gif

OzkanAbdullahoglu:
* After new organisation created when the user clicks to go to my profile,
* We switch to the newly created organisation.
Resolves #2085 


OzkanAbdullahoglu:
@rabbi08 Can you re-test in this branch? I was kinda made this faster than normal and did not very well test.

rabbi08:
@OzkanAbdullahoglu 
I have tested the fix on the feature branch, and it is working fine: PASS
Please refer below gif. Thanks

ezgif com-video-to-gif-4

BMariscal (dismissed):

Zaydme:
Would it be possible if we just update the organizationId in localStorage and then redirect, this way we will not need all the isNewOrganization logic. But if that was causing any issues, this is good as well.

OzkanAbdullahoglu:
> Would it be possible if we just update the organizationId in localStorage and then redirect, this way we will not need all the isNewOrganization logic. But if that was causing any issues, this is good as well.

Usually, I am not a fan of localStorage, It is slower. Because of that, I used redux state. Is it worth to try?

Zaydme:
I mean since we use localStorage already for persistency, you will have to update the localStorage anyway in case the user re-opened the website later, they must be on the same profile.
And localStorage isn't really slow, this is client side at the end ¯\_(ツ)_/¯

OzkanAbdullahoglu:
> I mean since we use localStorage already for persistency, you will have to update the localStorage anyway in case the user re-opened the website later, they must be on the same profile.
> And localStorage isn't really slow, this is client side at the end ¯_(ツ)_/¯

I believe we should not use localStorage for persistence either. (This should be one of refactoring case for the near future). I will check the consistency.

OzkanAbdullahoglu:
@Zaydme you can make your review with this new logic.

client/src/pages/ProfileCompleted.js
```diff
@@ -16,13 +17,22 @@ import {
...
   const { t } = useTranslation();
   const location = useLocation();
   useEffect(() => {
     refetchUser();
     // will only run once (on mount) to update navbar org list
   }, [refetchUser]);
+
+  const switchAccountDrawerMenu = () => {
+    if (location?.state?.orgId) {
+      const index = location.state.orgId;
```
Zaydme, line 30: I am only not sure about this part.
Does the router state have the newly created orgId?

client/src/reducers/session/reducer.js
```diff
@@ -13,6 +13,7 @@ const initialState = {
   forgotPasswordRequested: false,
   user: null,
   organisationId: null,
+  isNewOrganisation: false,
```
Zaydme: Do we still need this?

client/src/templates/layouts/NavigationLayout.js
```diff
@@ -72,6 +72,7 @@ const NavigationLayout = (props) => {
     webSocket,
     loggedInOnly,
     organisationId,
+    isNewOrganisation,
```
Zaydme: Same as above

client/src/reducers/session/reducer.js
OzkanAbdullahoglu: I forgot this one and the other. of course not.

client/src/pages/ProfileCompleted.js
OzkanAbdullahoglu: Yeap, below sample from `CreateOrganisationProfile` 

code

Zaydme (approved):
LGTM!

PR #2112, opened 2020-12-29
Feature/2068 change social to icons, do Oauth and API call directly

### LINKED ISSUE #2068: Profile setting page, entering social link should be more user friendly
As we discussed in the Slack channel, need to change the social links to be icons for a better UX result
image

ddveloper:
do Oauth flow and API call directly for social accounts
to solve #2068 

### DIFF
--- a/backend/lib/components/OAuth2/Facebook.js
+++ b/backend/lib/components/OAuth2/Facebook.js
@@ -1,56 +1,30 @@
 const axios = require("axios");
-const qs = require("querystring");
-const { generateUUID } = require("../../utils");
-const {
-  config: {
-    oauth2: { facebook },
-  },
-} = require("../../../config");
+const { Social } = require("./Social");
 
 // reference: https://developers.facebook.com/docs/facebook-login/manually-build-a-login-flow/
-
-const dialogUrlBase = "https://www.facebook.com/v9.0/dialog/oauth";
-const authUrl = "https://graph.facebook.com/v9.0/oauth/access_token";
+const authUrlBase = "https://www.facebook.com/v9.0/dialog/oauth";
+const tokenUrl = "https://graph.facebook.com/v9.0/oauth/access_token";
 const userDataUrl = "https://graph.facebook.com/me";
-// TODO: add 'user_link' after enabling in Facebook API dashboard
-const scopeData = ["id", "email", "first_name", "last_name"].join(",");
 
-const buildOauthUrl = (redirectUrl) => {
-  const stringifiedParams = qs.stringify({
-    client_id: facebook.clientId,
-    redirect_uri: redirectUrl,
-    response_type: "code",
-    scope: ["email", "user_friends", "user_link"].join(","),
-    state: generateUUID({ range: 32 }),
-  });
-  return `${dialogUrlBase}?${stringifiedParams}`;
-};
+class Facebook extends Social {
+  constructor(clientId, secretKey) {
+    super("facebook", clientId, secretKey, authUrlBase, tokenUrl);
+  }
 
-const authenticate = async (codeStr, stateStr, redirectUrl) => {
-  const { data } = await axios({
-    method: "get",
-    params: {
-      client_id: facebook.clientId,
-      client_secret: facebook.secretKey,
-      code: codeStr,
-      redirect_uri: redirectUrl,
-      state: stateStr,
-    },
-    url: authUrl,
-  });
-  return data.access_token;
-};
+  buildOauthUrl(redirectUrl) {
+    const scope = ["email", "user_friends", "user_link"].join(",");
+    return super.buildOauthUrl(scope, redirectUrl);
+  }
 
-const getUserLink = async (accesstoken) => {
-  const { data } = await axios({
-    method: "get",
-    params: {
-      access_token: accesstoken,
-      fields: scopeData,
-    },
-    url: userDataUrl,
-  });
-  return data.user_link;
-};
+  static async getUserLink(accessToken) {
+    // TODO: add 'user_link' after enabling in Facebook API dashboard
+    const scope = ["id", "email", "first_name", "last_name"].join(",");
+    const { data } = await axios.get(userDataUrl, {
+      fields: scope,
+      headers: { Authorization: `Bearer ${accessToken}` },
+    });
+    return data; // data.user_link;
+  }
+}
 
-module.exports = { authenticate, buildOauthUrl, getUserLink };
+module.exports = { Facebook };
--- a/backend/lib/components/OAuth2/Facebook.js
+++ b/backend/lib/components/OAuth2/Facebook.js
@@ -1,16 +1,23 @@
 const { Social } = require("./Social");
 
 // reference: https://developers.facebook.com/docs/facebook-login/manually-build-a-login-flow/
+const authUrlBase = "https://www.facebook.com/v9.0/dialog/oauth";
+const tokenUrl = "https://graph.facebook.com/v9.0/oauth/access_token";
+const userDataUrl = "https://graph.facebook.com/me";
+// TODO: add 'user_link' after enabling in Facebook API dashboard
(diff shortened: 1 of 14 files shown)


ddveloper:
#1722 has authentication part, may overlapped with the Oauth work in this PR

mannykary:
@ddveloper An early comment; this is a perfect example of leveraging OOP design patterns to make the code DRY. I would create a base class from which child classes can inherit. See https://medium.com/beginners-guide-to-mobile-web-development/super-and-extends-in-javascript-es6-understanding-the-tough-parts-6120372d3420. The base class would really be an abstract class to force subclasses to implement methods defined on it. See https://stackoverflow.com/questions/29480569/does-ecmascript-6-have-a-convention-for-abstract-classes

By following this pattern, it would be very easy to add additional OAuth providers by creating a new subclass.

I do like that you created a factory method to pass back the appropriate module, but I think it would be more robust if we also implemented the abstract class pattern.

ddveloper:
> @ddveloper An early comment; this is a perfect example of leveraging OOP design patterns to make the code DRY. I would create a base class from which child classes can inherit. See https://medium.com/beginners-guide-to-mobile-web-development/super-and-extends-in-javascript-es6-understanding-the-tough-parts-6120372d3420. The base class would really be an abstract class to force subclasses to implement methods defined on it. See https://stackoverflow.com/questions/29480569/does-ecmascript-6-have-a-convention-for-abstract-classes
> 
> By following this pattern, it would be very easy to add additional OAuth providers by creating a new subclass.
> 
> I do like that you created a factory method to pass back the appropriate module, but I think it would be more robust if we also implemented the abstract class pattern.

Thanks @mannykary I was trying to find the inherent in JS and you give me the reference 😄 like what pizzly did, we may als configure these clients only by a json metadata, I will create a parent class to abstract the logic

ddveloper:
update: UI need design team's inputs, for APIs see below:
```
Facebook, scopes need audition, right now can get user id, cannot get user_link
Instagram, app enabling need audition, may need user data deletion first
Twitter, to debug, auth URL is not working
Linkedin, my dev app has no verified scope, may need FP's app
Github, works fine
```

backend/lib/components/OAuth2/Facebook.js
```diff
@@ -0,0 +1,26 @@
+const { Social } = require("./Social");
+
+// reference: https://developers.facebook.com/docs/facebook-login/manually-build-a-login-flow/
+const authUrlBase = "https://www.facebook.com/v9.0/dialog/oauth";
```
mannykary, line 4: I think it would be better to set these as static properties on the class. The call to the super constructor would be much simpler that way since we wouldn't have to pass these in and the superclass can just access the properties directly

backend/lib/components/OAuth2/Facebook.js
```diff
@@ -0,0 +1,26 @@
...
+const tokenUrl = "https://graph.facebook.com/v9.0/oauth/access_token";
+const userDataUrl = "https://graph.facebook.com/me";
+
+class Facebook extends Social {
+  constructor(clientId, secretKey) {
+    super("facebook", clientId, secretKey, authUrlBase, tokenUrl, userDataUrl);
+  }
+
+  buildOauthUrl(redirectUrl) {
+    const scope = ["email", "user_friends", "user_link"].join(",");
```
mannykary: Similar vein to the static properties above, since it's not expected that this would change for the same OAuth provider (i.e. the scope passed in will never change during runtime). So this could actually be simplified or even removed since the super method signature is exactly the same as the child class method signature.

backend/lib/components/OAuth2/Facebook.js
```diff
@@ -0,0 +1,26 @@
...
+    return super.buildOauthUrl(scope, redirectUrl);
+  }
+
+  async getUserLink(accessToken) {
+    // TODO: add 'user_link' after enabling in Facebook API dashboard
+    const scope = ["id", "email", "first_name", "last_name"].join(",");
```
mannykary: Similar to my previous comment - just define a static property for scope. We should probably define two different variables for these scope types, or a single variable which is an object, with object properties defining the scope types.

Also, I think this might not be necessary if the data being returned is already in the format we want, like in the case of `Facebook`, `Twitter`, and `LinkedIn`; for `Github`, it appears that we would need to return `data.html_url` instead of `data`. What I would do is change the method name in the child classes to match that of the superclass (`getUserData`), and only override it if the format of the data that we want does not match. Something like this for `Github`, which would override the superclass method:

```
  async getUserData(accessToken) {
    const data = await super.getUserData(accessToken);
    return data.html_url;
  }
```

For `Facebook`, `Twitter` and `LinkedIn`, it is not necessary to override, so calling `getUserData` on instances of these child classes will just call the superclass method.

backend/lib/components/Oauth2.js
```diff
@@ -0,0 +1,35 @@
...
+const { SocialMock } = require("./OAuth2/SocialMock");
+const {
+  config: {
+    oauth2: { facebook, github, instagram, linkedin, twitter },
+  },
+} = require("../../config");
+
+// abstract factory to create oauth2 client
+const createOauth2Client = (socialType) => {
+  if (socialType === "facebook" && facebook.clientId && facebook.secretKey) {
```
mannykary: This could be more DRY if we defined an object mapping each `socialType` to the class:

```
const socialClasses = {
  facebook: {
    SocialClass: Facebook,
    config: facebook,
  },
  ...
};

...

const createOauth2Client = (socialType) => {
  if (!socialClasses.hasOwnProperty(socialType) {
    // Or maybe just return SocialMock since you defined that as a fallback.
    throw new Error('Unsupported socialType'); 
  }
  const { SocialClass, config } = socialClasses[socialType];
  if (config.clientId && config.secretKey) {
    return new SocialClass(config.clientId, config.secretKey);
  }
  return SocialMock;
}
```

mannykary (changes requested):
@ddveloper Looks much better! I see a few things that can be improved and made even more DRY.

backend/lib/components/Oauth2.js
ddveloper: Thank you so much @mannykary for the comments! I will improve them 🚵 based on your advices

backend/lib/components/OAuth2/Facebook.js
ddveloper: @mannykary when I define static properties in the children classes, my parent class member function cannot access them, (e.g. static authUrlBase in Facebook and buildOauthUrl() from Social), so I use `this.authUrlBase` currently so Social can access it (while this way doesn't guarantee children classes define these variables, any ideas how could I make it better?

mannykary: Ah that's a good point. Unfortunately ES6 doesn't support interfaces, so we can't really enforce subclasses not implementing these variables. I suppose you could just pass the variables into the super constructor from the child class and then do a check in the super constructor to make sure all arguments passed in are defined, and then set them as instance variables on the superclass. Not that pretty, but it would work.

```
class BaseClass {
  constructor(a, b, c) {
    // Check all args are defined
    if (!a || !b || !c) throw new TypeError('Must define all arguments');
    this.a = a;
    this.b = b;
    this.c = c;
  }
}

class ChildClass extends BaseClass {
  constructor(a) {
    super(a, 'def', 'hij');
  }
}

> const c = new ChildClass() // pass no arguments - should throw error
Uncaught TypeError: Must define all arguments
    at new BaseClass (repl:4:31)
    at new ChildClass (repl:3:5)
> const c1 = new ChildClass('abc')
undefined
> c1
ChildClass { a: 'abc', b: 'def', c: 'hij' }

```

mannykary:
@ddveloper AFAICT your latest commit looks good, but still need to review the whole diff again, so I'll do that when I have a bit more bandwidth. Feel free to ping me if you need me to review it sooner. Thanks!

ddveloper:
get stuck at scopes requests from Facebook and Linkedin, will unassign myself for now

PR #2117, opened 2021-01-04
#2088 - Changes 'GET /api/posts/postId' response on non-existing postID

### LINKED ISSUE #2088: [API] - endpoint GET /api/posts/{postId}
staging
endpoint GEt /api/posts/{postId}
When we send a different value than MongoDB ObjectID we have Internal Server error 500.
Could we have 404 (not found) or 400 (Bad request) ?
Scree
![Screen Shot 2020-12-15 at 13 37 34

emanuelquerty:

This PR resolves #2088 

### DIFF
--- a/backend/lib/endpoints/posts.js
+++ b/backend/lib/endpoints/posts.js
@@ -370,8 +370,10 @@ async function routes(app) {
           "author.location.zip": false,
         }),
       );
-      if (post === null || postErr) {
-        postErr && req.log.error(postErr, "Can't find a post with given id");
+      if (postErr){
+        req.log.error(postErr, "Can't find a post with given id");
+        throw app.httpErrors.badRequest();
+      }else if (post === null) {
         throw app.httpErrors.notFound();
       }
 
--- a/backend/lib/endpoints/posts.js
+++ b/backend/lib/endpoints/posts.js
@@ -370,9 +370,13 @@ async function routes(app) {
           "author.location.zip": false,
         }),
       );
+      
       if (postErr){
-        req.log.error(postErr, "Can't find a post with given id");
-        throw app.httpErrors.badRequest();
+        if (postErr instanceof mongoose.Error.CastError){
+            req.log.error(postErr, "Can't find a post with given id");
+            throw app.httpErrors.badRequest();
+        }
+        throw app.httpErrors.internalServerError();
       } else if (post === null) {
         throw app.httpErrors.notFound();
       }
--- a/backend/lib/endpoints/posts.js
+++ b/backend/lib/endpoints/posts.js
@@ -373,7 +373,7 @@ async function routes(app) {
       if (postErr){
         req.log.error(postErr, "Can't find a post with given id");
         throw app.httpErrors.badRequest();
-      }else if (post === null) {
+      } else if (post === null) {
         throw app.httpErrors.notFound();
       }
 
--- a/backend/lib/endpoints/posts.js
+++ b/backend/lib/endpoints/posts.js
@@ -370,10 +370,8 @@ async function routes(app) {
           "author.location.zip": false,
         }),
       );
-      if (postErr) {
-        req.log.error(postErr, "Failed retrieving post");
-        throw app.httpErrors.internalServerError();
-      } else if (post === null) {
+      if (post === null || postErr) {
+        postErr && req.log.error(postErr, "Can't find a post with given id");
         throw app.httpErrors.notFound();
       }
 


rabbi08 (dismissed):
for non-existing postIDs 404 -> Not found error is displayed: PASS
For existing postIDs the post is retrieved
Thanks

Refer below screenshot.

Screen Shot 2021-01-04 at 1 27 54 AM

backend/lib/endpoints/posts.js
```diff
@@ -370,10 +370,8 @@ async function routes(app) {
           "author.location.zip": false,
         }),
       );
-      if (postErr) {
-        req.log.error(postErr, "Failed retrieving post");
-        throw app.httpErrors.internalServerError();
-      } else if (post === null) {
+      if (post === null || postErr) {
```
ddveloper: I think it's better to treat these 2 cases differently, please check the `user` [endpoint code as an example](https://github.com/FightPandemics/FightPandemics/blob/9eca8b54ad34bc433b51a4591123417fb5b23b19/backend/lib/endpoints/users.js#L217),

ddveloper (changes requested):

backend/lib/endpoints/posts.js
emanuelquerty: Screen Shot 2021-01-04 at 1 02 23 AM

Hi @ddveloper! Initially, as you can see above, the code before this PR was similar to the user endpoint code that you attached as an example. Resolving the given issue with this PR in two separate cases as it initially was would result in two cases that have exactly the same response. That's why I decided to merge them into one case since I didn't touch the case when "post === null". But if you still think it's best to have them as two separate cases that have the same response, I can go ahead and do that. Please, let me know

Zaydme: If the ObjectId is invalid, exp "`GET /api/posts/1`"  this will result in `postErr` and not `post === null` which is why we get the 500 error.
But now we won't know when the db failed, since it will always return 404. A better thing to do will be validating the ObjectId and returning a 400 bad request error.

Zaydme (changes requested):
Please check my reply on the comment above.

backend/lib/endpoints/posts.js
emanuelquerty: Yeah now I get what you mean. It's just that 400 bad request didn't seem appropriate for the case where the ObjectId is invalid/doesn't exist and 404 seemed more appropriate. But you're right that having two cases that send the same response makes it hard to understand when the db failed.  I have changed the code as per your suggestion. Let me know if there's anything else to change. Thank you!

emanuelquerty:
Hi @Zaydme! I think I have implemented both yours and @ddveloper proposed changes. I have them as two separate changes  as per @ddveloper's feedback and I have the case invalid ObjectId return 400 bad request rather than 404 as per your feedback. I'm not sure if you mean something else. Please, take a look at the last commit and let me know

Zaydme:
What I was suggesting is keeping the internal server error, and the 404.
And adding a pre-validation for the ObjectId, maybe using fluent-schmea if possible.
Another good solution is keeping the internal server error, but checking for the error code and if it was the one for invalid ObjectId return a 400 instead of 500. While always keeping the 404 untouched.

emanuelquerty:
Ohh ... didn't know you mean that. I'm just not too sure about keeping the internal server error as that goes against the issue's resolution. As per #2088, this PR is supposed to change that to either 400 or 404. Since 404 raised the problem pointed by @ddveloper, that's the reason for my understanding that you meant 400. Maybe the second option, keeping the internal server error and checking for ObjectId, in which case, 400 request is returned would still align with the proposed resolution as per #2088. However, I'm not sure about what error code you mean by "checking for the error code". If you can explain that, I'd appreciate it

Zaydme:
What I mean is, inside of `if (postErr) {...}`
Add a
```js
if (err.message instanceof mongoose.Error.CastError)
   // Return 400, invalid ObjectId
  // Else return the normal 500 error
```

Zaydme:
Or 
```js
if (err instanceof mongoose.CastError) {
  // Handle this error
}
```
Not sure which one works, but this is what I meant by checking the error code.

emanuelquerty:
Thank you for the illustration. The first option doesn't work but the second does. I have pushed the changes. Please, let me know if it all looks good.

galeza:
Hi @emanuelquerty 
I tested on your branch.
For invalid ObjectID I got 400:
Screen Shot 2021-01-04 at 18 17 54

For valid ObjectID but not found I got 404:
Screen Shot 2021-01-04 at 18 17 29

Zaydme (approved):
LGTM!
Thank you @emanuelquerty

ddveloper (approved):
LGTM (maybe we'll change this to the schema validation latter)

galeza (approved):
LGTM

rabbi08 (approved):
Verified 
400, 404, and post ID found. Thanks.

PR #2144, opened 2021-01-14
Feature/2120 browser lang gtm

### LINKED ISSUE #2120: Localisation: Add 2 new Data Layer Variables with browser language info
Add 2 new Data Layer Variables `Browser Default Language` and `Language` to send to GTM.

`Browser Default Language` == The browser's default language when a user **first** lands on our platform.
(We already detect the browser's default language in i18n.js, so we can just pass the language detected here) 

`Language` == The language selected on our platform (when user first lands on our platform, this variable should be the browser's default language. if we do not have the default language as an option, defaults to en-US). When the user changes the language (via our language select menu), this variable should be updated with the chosen language.

Please use the same language code **suffix** we used for the language GTM codes (eg. English US is EN_US). Please note, do not use the "NAV_LS_" prefix as this is referring to the Language Select button in the nav bar of the website!

OzkanAbdullahoglu:
Resolves #2120 


doheekim93:
@FedericaZn will have to test this once it's on staging.

To double check, @OzkanAbdullahoglu, we have 2 variables - one for `browser default language` and one for `selected language`? And the `selected language` variable will update whenever the user changes language manually?

OzkanAbdullahoglu:
> @FedericaZn will have to test this once it's on staging.
> 
> To double check, @OzkanAbdullahoglu, we have 2 variables - one for `browser default language` and one for `selected language`? And the `selected language` variable will update whenever the user changes language manually?

Only the last part hasn’t been completed yet, we have two variables, but the selected language is not updating. I 'll handle that asap.

OzkanAbdullahoglu:
right now language data layer should be overwritten when the user changes the language from the menu. I couldn't perform a proper test without accessing gtm.

client/src/locales/languages.js
```diff
@@ -24,7 +24,7 @@ export const localization = {
   ru: "ru",
   "te-IN": "te-IN",
   th: "th",
-  tr: "tr",
+  "tr-TR": "tr-TR",
```
Zaydme: Wouldn't this need to be changed in line 51 as well. probably in gtm constants and the other languages file where the JSONs are imported.

Zaydme (commented):
LGTM!
Other than one thing, I left a comment.

client/src/locales/languages.js
OzkanAbdullahoglu: yes, I think It should be changed for all other languages too. But I couldn't test them all.  I changed this because of local language of the browser. 
Just "tr" would not work for below line 5,

code

I haven't checked line 51 and I don't know where it is used. I'll check that.

OzkanAbdullahoglu:
After I access gtm, I think dataLayers working as expected. (thnx @FedericaZn , @doheekim93)
dlayer1
And this is when we change the language 
chrome_2021-01-19_01-43-28

client/src/locales/languages.js
doheekim93: Are we changing Turkish to tr-tr? If so, we will definitely need to change in GTM constants and also the json file (on Lokalise). Please let me know if we are changing it!

OzkanAbdullahoglu: It is such a mistake. I don't understand. Now it's working or maybe there is an edge case and I couldn't find it again. Anyway, I discarded this change.

PR #2149, opened 2021-01-14
made zendesk widget dynamic and added react-zendesk npm package

### LINKED ISSUE #2126: Support button localization issue
- Environment : staging STG, it isn't out yet in production

- Steps to reproduce (STR): 
1. Simply just open the homepage
2. Check the support button on the bottom right

- Expected results: In my opinion the button label should be the proper translation of "Support", according to the picked language of the page

- Actual results: Even if the language settings on the used PC is English, I keep getting the button label translated to Hungarian (my native language), tried out in Incognito, checked out not just in Chrome but in Firefox browser as well. Got the same result. On the other hand when I run the tests by Cypress it will show Support doesn't matter which language we picked in the language section.

Screenshot of the Hungarian translation while the page language is English:
Capture local bug hungarian instead of english

Screenshot of the Support button during Cypress test while the page language is French
Support button in Cypress test

kamarenmccord:
Removed Zendesk widget from index.html and added it though the <srong>react-zendesk npm package</strong>. Now this component is reactive and you can dynamically set the language. I used the i18n Lib to get the set language and passed it to the zendesk api in a useEffect. The menu items do not change as they are pre-set in the Zendesk account I'm assuming. However the button and menu header does.

#2126


galeza (changes requested):
Hi,
I have a question about mobile. I tested on Android. When I was not logged in Icon Support was visible but after log in it disappears.

galeza:
Hi @rabbi08 
Do you know how this Support icon should behave on the mobile phone? When I'm logged in it disappears. I had to change language several times to see it again. Currently icon is visible on staging on the Help Board when user is logged in.

TimeaKovacs86:
Hi! I checked it again in desktop. Either way, if I am sign in or not... I get the Support button, but it's labeled in Hungarian. I am not sure why is that at all. None of my settings are set in Hungarian on this PC, the browser language is English, I changed the language on the localhost page from the default American English to French, the Support button (Támogatás) is still in Hungarian, And the header and the first sub header of the pop up zendesk window also Hungarian (Kapcsolatfelvétel-Get in touch or something like that and Válassza ki a problémát-Choose from the problems). Then the options under are in English : Report a bug, I want to volunteer with FightPandemics and so on... It might be some kind of settings on my PC, and maybe it's how it should be. Should the Support icon behave like that?

rabbi08 (changes requested):
> Hi @rabbi08
> Do you know how this Support icon should behave on the mobile phone? When I'm logged in it disappears. I had to change language several times to see it again. Currently icon is visible on staging on the Help Board when user is logged in.

@galeza 
IN STG env: The support icon is behaving almost the same on mobile and desktop for me, please see the video from my ph.
ezgif com-video-to-gif

@kamarenmccord In this feature branch: 
1. translation of the 'Support' button is working: PASS
**2. BUT after user logged in the support button is not displayed: FAIL** 

Please see below GIF:
ezgif com-video-to-gif-2


Also, I wanna add, that support icon is not displayed on 'Profile' page.. logged in or logged out, own profile or other user profile, because when we log the ticket through the widget it share the page details and on profile page, it shares the user profile details along with the ticket. So the support option is not available on 'Profile' page.

Please let me know if the query is clear. Thanks

CamilaArias:
@rabbi08 I see from the gif the button disappears from the profile and sign in screen. That is expected because we specifically removed it from there because of security issues, if we fix it that would be reverting the decision we made before.

rabbi08:
> @rabbi08 I see from the gif the button disappears from the profile and sign in screen. That is expected because we specifically removed it from there because of security issues.

Yes @CamilaArias expected and working as expected, but after login its not displayed on all the other pages too. 
Please see my comments and screen recording. Also, @galeza and @TimeaKovacs86 report the same issue.

Thanks

kamarenmccord:
By moving it up to a higher level before the routes I was able to by-pass whatever Css or rule that was turning it off. I realized that When you entered ANY page that the widget turned off logged in or not and it stayed off until the user hit refresh. So i moved it up before the Router and it reloads each time now. @CamilaArias  @rabbi08  @TimeaKovacs86  @galeza

rabbi08 (dismissed):
Verified the fix in feature branch as working fine. Thanks @kamarenmccord 

Please see below GIF showing Zendesk widget for both logged and non-logged-in users.
ezgif com-video-to-gif-5

Please see below GIF showing the translation of the Zendesk widget 
ezgif com-video-to-gif-6

galeza (dismissed):
LGTM

TimeaKovacs86 (dismissed):
Great job! Thank you!

package-lock.json
```diff
@@ -46,37 +30,18 @@
...
-    "regenerator-runtime": {
-      "version": "0.13.7",
-      "resolved": "https://registry.npmjs.org/regenerator-runtime/-/regenerator-runtime-0.13.7.tgz",
-      "integrity": "sha512-a54FxoJDIr27pgf7IgeQGxmqUNYrcV338lf/6gH456HZ/PhX+5BcwHXG9ajESmwe6WRO0tAzRUrRmNONWgkrew=="
-    },
-    "shallowequal": {
-      "version": "1.1.0",
-      "resolved": "https://registry.npmjs.org/shallowequal/-/shallowequal-1.1.0.tgz",
-      "integrity": "sha512-y0m1JoUZSlPAjXVtPPW70aZWfIL/dSP7AFkRnniLCrK/8MDKog3TySTBmckD+RObVxH0v4Tox67+F14PdED2oQ=="
+    "react-zendesk": {
```
ddveloper: Thanks @kamarenmccord , maybe no need to change this json file?

kamarenmccord: I added back the other 40 some lines of code in this and inserted the Zendesk at the end. Is this okay or should i just remove all the changes?

ddveloper: yes Kamaren, I tested on my local setup, adding library under the `src\package.json` and `src\package-lock.json` is working, so I don't see we need to modify this outter layer package-lock.json, let me know if this makes sense 😃

kamarenmccord: absolutely, I must have ran the npm installer in the wrong directory. Thanks for that clarification.

rabbi08 (dismissed):
Thanks @kamarenmccord @ddveloper 
This looks fine the button is working fine and translation is working for the button.

PR #2157, opened 2021-01-18
find post based on user location, sort by time

### LINKED ISSUE #2141: Filter: Search results, based on location(country), are not sorted as per time.
1. Login into STG/PROD "fightpandemics.com"
2. Navigate to Helpboard and apply location filter = 'United States' 

Issue: You will observe that 6 months old posts are displayed on the top for the selected location.
But when you keep scrolling, you will see recent post for the searched location.
So search results are as per location but they are not sorted as per time when they are posted. 

Expected Result: When user apply location filter -> the latest posts for the selected location should be displayed as per the age i.e. latest on top and so on.

See below screenshots/GIFs
Screen Shot 2021-01-13 at 7 47 18 PM

Screen Shot 2021-01-13 at 7 52 06 PM

ezgif com-video-to-gif-4

Please note, This issue is related to #2053 

**Some Exceptions need to be verified while fixing this issue:**
1. When we search on the base of country, this issue always exist, but when user expand the search location i.e. city/state/country, results are better. Please see this in GIF.

If I search 'United States' 'United Kingdom' 'Spain' 'Barcelona, Spain' the results are NOT accurate.

**Note: So, the locations are not matched properly when broader location is searched for.**

ddveloper:
for searching posts near user's location, limit the distance in a fixed range and sort by time
for issue 2141

### DIFF
--- a/backend/lib/endpoints/posts.js
+++ b/backend/lib/endpoints/posts.js
@@ -64,35 +64,46 @@ async function routes(app) {
       ];
 
       // prefer location from query filters, then user if authenticated
-      let searchNearBy = false;
       let location;
       if (queryFilters.location) {
         location = queryFilters.location;
       } else if (actor && !ignoreUserLocation) {
         location = actor.location;
-        searchNearBy = true;
       }
 
+      // determine if search around (smller range than city),
+      let searchAround = false;
       if (location) {
+        let locationFilter = [
+          { visibility: "worldwide" },
+          {
+            visibility: "country",
+            "author.location.country": location.country,
+          },
+          {
+            visibility: "state",
+            "author.location.country": location.country,
+            "author.location.state": location.state,
+          },
+          {
+            visibility: "city",
+            "author.location.country": location.country,
+            "author.location.state": location.state,
+            "author.location.city": location.city,
+          },
+        ];
+
+        // search particular on city/state/country
+        if (location.country && !location.state)
+          locationFilter = locationFilter.slice(1);
+        else if (location.state && !location.city)
+          locationFilter = locationFilter.slice(2);
+        else if (location.city && !location.neighborhood)
+          locationFilter = locationFilter.slice(3);
+        else searchAround = true;
+
         filters.push({
-          $or: [
-            { visibility: "worldwide" },
-            {
-              visibility: "country",
-              "author.location.country": location.country,
-            },
-            {
-              visibility: "state",
-              "author.location.country": location.country,
-              "author.location.state": location.state,
-            },
-            {
-              visibility: "city",
-              "author.location.country": location.country,
-              "author.location.state": location.state,
-              "author.location.city": location.city,
-            },
-          ],
+          $or: locationFilter,
         });
       }
       /* eslint-enable sort-keys */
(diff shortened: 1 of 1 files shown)


ddveloper:
@rabbi08 please help to check if this PR (another method) can fix the issue

CamilaArias:
Thanks @ddveloper! It fixes the sorting by time for signed in and out users which is very important.
One issue I see is that I get a post if I choose 2, 5 or 30km in range, but nothing if I click city or state. I get two posts in 30km but one in "Country" and both posts are in the same country. It's probably related to missing some data in the posts but it feels contradictory and confusing.

rabbi08:
Thanks @ddveloper 

Following are few observations:

1. While filtering on the base of the country, only the Post having address as 'US' but not all the posts from US. Please see below screen:

https://user-images.githubusercontent.com/67168400/105382510-236ac800-5bde-11eb-8535-d93de0164b3f.mov

2. While applying the option 'Show results near me': 
the search results display not relevant result in between while sorting as per time. For e.g. sorting location is 'New York' but while displaying the results it display the result of 'Princeton NJ' in between the post associated with 'New York' location, because on the basis of time sorting. 
This scenario is like little tricky. i.e. if I want to see post near me sorted as per time, then as time it gets priority over location, so it is displayed over same location but older post.

Query/Issue: so what if the latest post is from a location which is not near me, then that will take priority. 

https://user-images.githubusercontent.com/67168400/105383151-cde2eb00-5bde-11eb-8c44-8ce49fee31e9.mov

For filters, the search is not displayed as expected but, 'show post near me' seems like working, need to look more into this scenario, though. 

Still testing, will keep updates here. Thanks

ddveloper:
Thanks @rabbi08 
for issue 1, I will try to fix it
for issue 2, the reason is I changed to search a range of 100km range from center of the search, since 'Princeton NJ' is within 100km range from New York, so the result is within the search result

> Query/Issue: so what if the latest post is from a location which is not near me, then that will take priority.  

in this case, if you enabled 'near me' it will not show, the search for 'near me' is collecting all result within 100km then sorting them based on latest post

ddveloper:
About issue 1, I found the reason is not all posts has `visibility` set to 'country' or 'worldwide', in this case, if you search for posts in a country but the post sets its visibility to be 'city', it won't show

rabbi08:
> Thanks @rabbi08
> for issue 1, I will try to fix it
> for issue 2, the reason is I changed to search a range of 100km range from center of the search, since 'Princeton NJ' is within 100km range from New York, so the result is within the search result
> 
> > Query/Issue: so what if the latest post is from a location which is not near me, then that will take priority.
> 
> in this case, if you enabled 'near me' it will not show, the search for 'near me' is collecting all result within 100km then sorting them based on latest post

- I am using 'show result' option for this query, I would like add below observations:
   1. 100km is not like near me, like distance of 2hrs, can we keep like 50km/miles.

> About issue 1, I found the reason is not all posts has `visibility` set to 'country' or 'worldwide', in this case, if you search for posts in a country but the post sets its visibility to be 'city', it won't show

@ddveloper as I did consider this scenario and tested it, but It doesn't look like this is the issue. Please see the below video for the same. It looks like while comparing the country names, it is only comparing the post where only country names are given. Thanks

https://user-images.githubusercontent.com/67168400/105411296-380c8780-5c01-11eb-84a4-e1a55677f2ed.mov

ddveloper:
Thanks @rabbi08 , I reduced the search range default to be 50km, also the city/state/country search is fixed, please have a try 😃

ddveloper:
will be covered by issue #2171 , close this one

PR #2179, opened 2021-01-26
Lokalise: Translations update (adding Croatian)

jyclam:
This pull request was initiated by Lokalise (user Do Hee Kim) at 2021-01-26 20:07:30


ddveloper:
merged into #2162

PR #2187, opened 2021-01-28
enable veriff badge, tested in prod

ddveloper:
enable Veriff Badge feature, it's working in the production


rabbi08 (approved):
thanks @ddveloper 
Verified the feature of 'Veriff' in production with help of @doheekim93, hence we can enable this feature in production. Thanks

PR #2196, opened 2021-01-31
Feature/2192 organistation preformance issue

### LINKED ISSUE #2192: endpoint /api/organisations failed to paginate
when you try these 3 endpoints, you'll notice the pagination for the organisations one doesn't work
need to fix it to address the performance issue
```
https://staging.fightpandemics.work/api/posts
https://staging.fightpandemics.work/api/users
https://staging.fightpandemics.work/api/organisations
```

kamarenmccord:
<h2>Switched the original endpoint</h2>
<p>How does this make performance better?
<ul>
<li>Original endpoint returned all organizations which caused performance issues and long load times</li>
<li>The endpoint /search is now the default endpoint and returns up to 10 results unless limit is passed in the query-string</li>
<li>the endpoint now returns quicker</li>
<br/>
we can change the hard limit if desired (hard limit right now is 10)
<br/><br/>
Since the endpoint "search" is now "/" we should make sure all organization related content returns properly.
<br/><br/>
_Please be concise and link any related issue(s):_<br/>
#2192


rabbi08:
@kamarenmccord 
Is this PR ready for Review?

Thanks

ddveloper (approved):
LGTM : )

rabbi08:
This looks fine, but same performance issue have also been observed for 
"/api/users/" too:

https://app.datadoghq.eu/synthetics/details/qin-qpt-tdy?resultId=6440132386955817431
image
 
@kamarenmccord can we fix above issue too with this one?
 Thanks

kamarenmccord:
@rabbi08 The code between the users and the organizations look almost identical. Including how the result limitations are set. I am not really sure why is it taking so long to respond. could this be due to a larger volume of users vs organizations?

rabbi08 (approved):
Thanks @kamarenmccord for the explanation and fix.

As I discussed with @ddveloper, he said that pagination is implemented for users, so we might don't have identical issues for users. Currently, I think we can push this fix and observe further. 

Thanks

PR #2210, opened 2021-02-10
Feature/2201 profile redesign menu tabs

clubkemp:
..../components/Tabs/BasicTabs.js is how I started to design the tabs. Build it in it's own separate component then apply the styling and import it onto the profile page. The problem here, is that the tabs are made to show/hid the components that are nested inside of it. The first method would have been to ditch the nesting functionality and use a callback (lines 368-265 in the profile component) to update a piece of state we can use to render the proper components.

Right now we are using the other way, which is to just import the tabs components directly into profile and then nest everything we need inside of the different tabpanes. No state control needed, the component itself handles everything.


clubkemp:
components\Profile\ProfileTabs.js is ready to be reviewed. This is the component that I am using to build the mobile version of the tabs to toggle between Activity, Posts, Thanks, Badge. This component will be used as the tabs for the cards in the profile component. You feed `<ProfileTabs /> `props to be how you need it below is a list of the props it is expecting and what it defaults to if nothing is specified. bolded props are required

- **callback** -> this is the callback function that will change your state so you can filter what component is being rendered
- **tabs** -> an array of the items you want to appear as tabs
- position -> where you want the tabs to render top, left, right, bottom. Defaults to top
- def -> which tab index position you want to be active by default
-  disabled -> an array of equal length to your tabs listing which ones are disabled (true) and which ones are active (false)

```
<ProfileTabs
            callback={handleView}
            tabs={["Activity", "Posts", "Thanks", "Badge"]}
            position={"top"}
            def={"1"}
            disabled={[true, false, true, true]}
          />
```
`const [view, setView] = useState("posts");`

I setup my callback to take in my index from the ProfileTabs component and alter my view state
```
const handleView = (key) => {
    switch (key) {
      case "0":
        setView("activity");
        break;
      case "1":
        setView("posts");
        break;
      case "2":
        setView("thanks");
        break;
      case "3":
        setView("badge");
        break;
    }
  };
```

aditilamkhede:
@clubkemp is this ready for review or still in progress?

clubkemp:
Closing PR due to the fact we are moving this code onto a branch off of the branch feature/profile-redesign

PR #2216, opened 2021-02-12
Resuable ProfileTabs component for use in profile redesign

### LINKED ISSUE #2201: Create "Menu tabs"
Build a Menu navigation tab with 1 tab that says "Posts"

clubkemp:
components\Profile\ProfileTabs.js is ready to be reviewed. This is the component that I am using to build the mobile version of the tabs to toggle between Activity, Posts, Thanks, Badge. This component will be used as the tabs for the cards in the profile component. You feed `<ProfileTabs /> `props to be how you need it below is a list of the props it is expecting and what it defaults to if nothing is specified. bolded props are required

- **callback** -> this is the callback function that will change your state so you can filter what component is being rendered
- **tabs** -> an array of the items you want to appear as tabs
- position -> where you want the tabs to render top, left, right, bottom. Defaults to top
- def -> which tab index position you want to be active by default
-  disabled -> an array of equal length to your tabs listing which ones are disabled (true) and which ones are active (false)

```
<ProfileTabs
            callback={handleView}
            tabs={["Activity", "Posts", "Thanks", "Badge"]}
            position={"top"}
            def={"1"}
            disabled={[true, false, true, true]}
          />
```
`const [view, setView] = useState("posts");`

I setup my callback to take in my index from the ProfileTabs component and alter my view state
```
const handleView = (key) => {
    switch (key) {
      case "0":
        setView("activity");
        break;
      case "1":
        setView("posts");
        break;
      case "2":
        setView("thanks");
        break;
      case "3":
        setView("badge");
        break;
    }
  };
```

### DIFF
--- a/client/src/components/Profile/ProfileTabs.js
+++ b/client/src/components/Profile/ProfileTabs.js
@@ -1,0 +1,42 @@
+import React from "react";
+import BasicTabs from "components/Tabs/BasicTabs";
+import styled from "styled-components";
+
+// Styled component named StyledButton
+const StyledTabs = styled.div`
+  .ant-tabs-ink-bar {
+    border-bottom: 2px solid #282828;
+    width: 200px;
+  }
+  .ant-tabs-tab-btn {
+    color: #939393;
+  }
+  .ant-tabs-tab-active > .ant-tabs-tab-btn {
+    font-family: Poppins;
+    font-size: 14px;
+    font-style: normal;
+    font-weight: 700;
+    line-height: 21px;
+    letter-spacing: 0px;
+    text-align: center;
+    color: #282828;
+  }
+  .ant-tabs-tab-disabled > .ant-tabs-tab-btn {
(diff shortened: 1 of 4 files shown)


aditilamkhede (changes requested):
@clubkemp in handle view and line 52 Instead of hardcoding values there, can you define a array at the top and fetch array values from locales(I may be using wrong term here). By locales I mean, the language json files. There is already a profile section there, just add it to that. @clubkemp just make sure to use figma styles, just a reminder, you might have already done that.

clubkemp:
I've gone ahead and added a section in en_US.json file to hold the navigation text for these
```
"views": {
      "activity": "Activity",
      "posts": "Posts",
      "organizations": "Organizations",
      "badges": "Badges",
      "thanks": "Thanks"
    }
```
 I then used the` t()` function to pull those values in and build an array to send to the components. This makes the callback function much more simple and flexible.

PR #2217, opened 2021-02-13
Feature/profile tabs

clubkemp:
components\Profile\ProfileTabs.js is ready to be reviewed. This is the component that I am using to build the mobile version of the tabs to toggle between Activity, Posts, Organizations, Thanks, Badge. This component will be used as the tabs for the cards in the profile component. You feed <ProfileTabs /> props to be how you need it below is a list of the props it is expecting and what it defaults to if nothing is specified. bolded props are required

callback -> this is the callback function that will change your state so you can filter what component is being rendered
tabs -> an array of the items you want to appear as tabs
position -> where you want the tabs to render top, left, right, bottom. Defaults to top
def -> which tab index position you want to be active by default
disabled -> an array of equal length to your tabs listing which ones are disabled (true) and which ones are active (false)

```
<ProfileTabs
            callback={handleView}
            tabs={viewsArray}
            position={"top"}
            def={"1"}
            disabled={[true, false, true, true, true]}
          />
```
`const [view, setView] = useState(t("profile.views.posts"));`

The callback is pretty simple, it simply matches the index of the key to the index of the array you fed into your tabs
```
const handleView = (key) => {
    setView(viewsArray[key]);
  };
```

### DIFF
--- a/client/src/components/Profile/ProfileTabs.js
+++ b/client/src/components/Profile/ProfileTabs.js
@@ -1,0 +1,42 @@
+import React from "react";
+import BasicTabs from "components/Tabs/BasicTabs";
+import styled from "styled-components";
+
+// Styled component named StyledButton
+const StyledTabs = styled.div`
+  .ant-tabs-ink-bar {
+    border-bottom: 2px solid #282828;
+    width: 200px;
+  }
+  .ant-tabs-tab-btn {
+    color: #939393;
+  }
+  .ant-tabs-tab-active > .ant-tabs-tab-btn {
+    font-family: Poppins;
+    font-size: 14px;
+    font-style: normal;
+    font-weight: 700;
+    line-height: 21px;
+    letter-spacing: 0px;
+    text-align: center;
+    color: #282828;
+  }
+  .ant-tabs-tab-disabled > .ant-tabs-tab-btn {
+    color: #e1e1e1;
+  }
+  .ant-tabs-nav-list {
+    margin: auto;
+    width: 90%;
+    display: flex;
+    justify-content: space-between;
+  }
+`;
+
+export default function ProfileTabs(props) {
+  // Use it like any other component.
+  return (
+    <StyledTabs>
+      <BasicTabs {...props} />
+    </StyledTabs>
+  );
+}
--- a/client/src/components/Profile/ProfileTabs.js
+++ b/client/src/components/Profile/ProfileTabs.js
@@ -26,9 +26,10 @@ const StyledTabs = styled.div`
   }
   .ant-tabs-nav-list {
     margin: auto;
-    width: 90%;
+    width: 800px;
     display: flex;
-    justify-content: space-between;
+    justify-content: space-around;
+    overflow: visible;
   }
 `;
 
--- a/client/src/components/Tabs/BasicTabs.js
+++ b/client/src/components/Tabs/BasicTabs.js
@@ -8,15 +8,8 @@ const BasicTabs = ({ tabData }) => {
   return (
     <Tabs tabPosition={position} defaultActiveKey={defaultView}>
       {tabs.map((e, index) => {
-        if (e.display) {
-          return (
-            <TabPane tab={e.tabName} key={index}>
-              {e.tabView}
-            </TabPane>
-          );
-        }
         return (
-          <TabPane tab={e.tabName} key={index} disabled>
+          <TabPane tab={e.tabName} key={index} disabled={!e.display}>
             {e.tabView}
           </TabPane>
         );
--- a/client/src/components/Tabs/BasicTabs.js
+++ b/client/src/components/Tabs/BasicTabs.js
@@ -3,22 +3,23 @@ import { Tabs } from "antd";
 
 const { TabPane } = Tabs;
 
-const BasicTabs = ({ callback, tabs, position, def, disabled }) => {
+const BasicTabs = ({ tabData }) => {
+  const { defaultView, position, tabs } = tabData;
   return (
-    <Tabs
-      tabPosition={position ? position : "top"}
-      defaultActiveKey={def ? def : "0"}
-      onChange={callback}
-    >
+    <Tabs tabPosition={position} defaultActiveKey={defaultView}>
       {tabs.map((e, index) => {
-        if (disabled) {
-          if (disabled[index] === true) {
-            return <TabPane tab={e} disabled key={index}></TabPane>;
-          } else {
-            return <TabPane tab={e} key={index}></TabPane>;
-          }
+        if (e.display) {
+          return (
+            <TabPane tab={e.tabName} key={index}>
+              {e.tabView}
+            </TabPane>
+          );
         }
-        return <TabPane tab={e} key={index}></TabPane>;
+        return (
+          <TabPane tab={e.tabName} key={index} disabled>
+            {e.tabView}
+          </TabPane>
+        );
       })}
     </Tabs>
   );
--- a/client/src/components/Tabs/BasicTabs.js
+++ b/client/src/components/Tabs/BasicTabs.js
@@ -4,10 +4,6 @@ import { Tabs } from "antd";
 const { TabPane } = Tabs;
 
 const BasicTabs = ({ callback, tabs, position, def, disabled }) => {
-  useEffect(() => {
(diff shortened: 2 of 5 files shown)


aditilamkhede (approved):
@clubkemp can you put views under individual section (inside json file).

linadinh2602:
Can you include screenshot of how this looks?

client/src/components/Tabs/BasicTabs.js
```diff
@@ -0,0 +1,31 @@
+import React, { useEffect, useState } from "react";
+import { Tabs } from "antd";
+
+const { TabPane } = Tabs;
+
+const BasicTabs = ({ callback, tabs, position, def, disabled }) => {
+  useEffect(() => {
+    console.log(tabs);
```
linadinh2602: remove console.log

aditilamkhede:
> Can you include screenshot of how this looks?

Yes @clubkemp if you could include a video or screenshot that will be great.

client/src/components/Profile/ProfileTabs.js
```diff
@@ -0,0 +1,43 @@
...
+  .ant-tabs-nav-list {
+    margin: auto;
+    width: 800px;
+    display: flex;
+    justify-content: space-around;
+    overflow: visible;
+  }
+`;
+
+export default function ProfileTabs(props) {
```
linadinh2602, line 36: Why do we need a separate component for ProfileTab and one for BasicTab?

client/src/components/Tabs/BasicTabs.js
```diff
@@ -0,0 +1,31 @@
...
+  return (
+    <Tabs
+      tabPosition={position ? position : "top"}
+      defaultActiveKey={def ? def : "0"}
+      onChange={callback}
+    >
+      {tabs.map((e, index) => {
+        if (disabled) {
+          if (disabled[index] === true) {
+            return <TabPane tab={e} disabled key={index}></TabPane>;
```
linadinh2602: The view that will be displayed each tab is the children of the <TabPane> components. The current props does not allow the parent component of this basic tab to pass in the children of the TabPane, so it's not possible to render anything other than an empty view with the current design right now. For example in this demo code:
const Demo = () => (
  <Tabs defaultActiveKey="1" onChange={callback}>
    <TabPane tab="Tab 1" key="1">
      Content of Tab Pane 1
    </TabPane>
    <TabPane tab="Tab 2" key="2">
      Content of Tab Pane 2
    </TabPane>
    <TabPane tab="Tab 3" key="3">
      Content of Tab Pane 3
    </TabPane>
  </Tabs>
);
The "Content of Tab Pane 1", Content of Tab Pane 2" and Content of Tab Pane 3" are the children of the TabPane, and we need a way to pass such children into these pane.

linadinh2602 (changes requested):

client/src/components/Tabs/BasicTabs.js
clubkemp: done!

client/src/components/Profile/ProfileTabs.js
clubkemp: BasicTabs creates a default ant-d styled tab, that could be used anywhere in the application (outside of the profile component) as long as you pass in the needed props. The ProfileTab components adds on the styling you need to make it look how we want it in the profile section. This seemed consistent with the pattern I saw elsewhere in FP.

So if someone wanted tabs for say the feed section, but they didn't want it black and they wanted it to be blue with some fancy color for the highlighted, they wouldn't have to make a whole new tabs component, just a similar styled version of BasicTabs

client/src/components/Tabs/BasicTabs.js
clubkemp: That is the traditional way these tabs are used to display/hide content is through the child/parent hierarchy of the tabPanes. To make it more reusable in other places, and to avoid a bunch of div nesting we are using a more stateful approach. Let me elaborate

You would setup a piece of state in your component, let's call it view

`const [view, setView] = useState(pane1);`

Every time you click a different Tab the callback you pass into BasicTab changes your state so click TabPane1 the view is pane1, click TabPane2 and view=pane2 etc.

this allows you to organize your components like this
```
<BasicTabs />
{pane1 ? Content of Tab Pane 1 : null }
{pane2 ? Content of Tab Pane 2 : null }
{pane3 ? Content of Tab Pane 3 : null }

```

Otherwise anytime anyone wants to create tabs, they will have to just make the tabs from scratch inside of their components which isn't super hard, just a pretty W.E.T way of doing it.

Does that make sense? Maybe we need to meet to walk through it?

clubkemp:
Untitled_ Feb 14, 2021 12_34 PM

client/src/components/Tabs/BasicTabs.js
linadinh2602: I'm a bit concern since the way you implement it may break the expectation of antd-Tab component. The antd-Tab component expects the view of each Tab to be the children of TabPane. It does all the animation and views switching and view placement logic there. If you move the tab view to be the sibling of the parent <Tab> component instead, then the consumer of your tab component has to re-implement view switching logic, animation, and placement themselves (for example, using the react hook like in your example for view switching), and that is a lot of core logic to be re-implemented. Can you upload a video of this tab switching between different views with some placeholder text in it? I think if you try that you'll see the animation that comes with the Tab component will be missing, and you have to manually place the view in the correct location via flexbox.
To address your concern regarding the reusability of BasicTab, it can be achieved without compromising the built-in functionality of the Tab component bypassing the children's view itself as props to the Basic Tab component. For example, if you look at `AvatarModal` component in the code base, You'll see the footer prop is a React component that is being passed through. We can achieve reusability using this pattern, so instead of just passing to BasicTab the tab name, it can pass an object which includes {'tabName': ..., 'tabView': [a React component], disabled: [boolean]} This can also help remove the disabled array and wrap it in the same object. I believe doing this way will allow us to reuse more functionality from the tab component without compromising the reusability of the basic tab component that you build.

clubkemp: The functionality is the same regardless (See below with the two different structures compared). It basically comes down to if you want to write your components inplace with all the rest of your components, or if you want to write them above in a view object which is passed in as a prop. If the latter makes more sense to you, it'll probably make more sense to everyone else. I'll go ahead a refactor

Nesting the components
FightPandemics

Controlling with state
FP - Stateful tabs

linadinh2602:
I think you may have forgotten to committed the change in BasicTab component :) I see your change in the Profile page, but not the changes in BasicTab. From looking at your GIF, I believe you have made the appropriate change there too, but just forget to commit and push the file up

client/src/components/Profile/ProfileTabs.js
```diff
@@ -0,0 +1,43 @@
+import React from "react";
+import BasicTabs from "components/Tabs/BasicTabs";
+import styled from "styled-components";
+
+// Styled component named StyledButton
+const StyledTabs = styled.div`
```
linadinh2602, line 6: I think you can do styled(BasicTabs), then you don't have to wrap this around a div

client/src/components/Tabs/BasicTabs.js
```diff
@@ -0,0 +1,28 @@
...
+const { TabPane } = Tabs;
+
+const BasicTabs = ({ tabData }) => {
+  const { defaultView, position, tabs } = tabData;
+  return (
+    <Tabs tabPosition={position} defaultActiveKey={defaultView}>
+      {tabs.map((e, index) => {
+        if (e.display) {
+          return (
+            <TabPane tab={e.tabName} key={index}>
```
linadinh2602: you can remove the if block if you do disabled={!e.display}

linadinh2602 (approved):
just minor comments, please feel free to merge after you address the issues.

client/src/components/Profile/ProfileTabs.js
clubkemp: That makes sense, but I am not able to get it to work without the div wrapper. Marking resolved for now and if it poses as a breaking feature later on I'll address then

client/src/components/Tabs/BasicTabs.js
clubkemp: done!

Issue #2200, opened 2021-02-06
Build the reusable card component


aditilamkhede:
A reusable card with the ability to contain 2 or 3 tabs within it

PR #2219, opened 2021-02-14
#2200 - Add Redesign PostTabCard Component

linadinh2602:
image
- PostTabCard take in an array of cardContent number of maxPost, look like this: 
          <PostTabCard cardContents={[
            {
              title: 'Request',
              posts: Object.values(posts.posts),
            },
            {
              title: 'Offer',
              posts: Object.values(posts.posts),
            }
          ]} />
- if maxPost is not given, it'll default to 3
- Currently, the menu dropdown doesn't have any callback implemented yet based on the discussion because it'll be changing in the future.
- This PR doesn't integrate any component yet
_Please be concise and link any related issue(s):_ #2200 


clubkemp (commented):
What exactly are you modifying in the Post.js file? It looks like a lot of formatting changes, did you actually make changes to that component that effect PostTabCard? If you did could you let me know some line numbers so I can review, otherwise should discard the changes??

linadinh2602:
@clubkemp  I didn't change the Post.js, I didn't realized it was updated due to the linter on my VSCode. You can safely ignore the change in that file.

client/src/components/Feed/Post.js
```diff
@@ -467,17 +463,17 @@ const Post = ({
...
         </div>
       }
       thumb={
         post?.author?.photo ? (
           post.author.photo
         ) : (
-          <TextAvatar>{AvatarName}</TextAvatar>
-        )
+            <TextAvatar>{AvatarName}</TextAvatar>
+          )
```
linadinh2602: Will revert changes in Post.js. This is updated by accident.

clubkemp (approved):

aditilamkhede (changes requested):
@linadinh2602 it looks good, just few things.
1. As we want this component to be reusable, I would suggest to move all styled components to separate js. and export them. That way any body can reuse them. 
2. wherever px are used, if you can use rem as unit, please do so.
3. Also put your formatdate function in util so it becomes reusable if needed.
4. Add a comment over your functions, what it does.

client/src/components/Feed/PostTabCard.js
```diff
@@ -0,0 +1,182 @@
...
+  )
+}
+
+const PostTabCard = ({ cardContents, maxPosts = 3 }) => {
+  const tabs = cardContents.map((item) => ({ title: item.title }))
+  return (
+    <Container>
+      <Tabs tabs={tabs}
+        initialPage={1}
+        onChange={(tab, index) => { console.log('onChange', index, tab); }}
```
aditilamkhede: remove these logs too

linadinh2602:
@aditilamkhede Sure I can expose everything in a separate file, but I'm worried that making everything public (reusable) by default will make things more confusing. quoted this article: https://itnext.io/thinking-in-styled-components-e230ea37c52c
"In short, don’t default to making things reusable just for the heck of it, because there are tradeoffs. It’s not necessarily better to make things as reusable as possible."

aditilamkhede:
> @aditilamkhede Sure I can expose everything in a separate file, but I'm worried that making everything public (reusable) by default will make things more confusing. quoted this article: https://itnext.io/thinking-in-styled-components-e230ea37c52c
> "In short, don’t default to making things reusable just for the heck of it, because there are tradeoffs. It’s not necessarily better to make things as reusable as possible."

right @linadinh2602 but in our case our main aim was to create reusable components, like your post card, we have to create Request and Offer cards too. those styles which you created for post card, we are going to create many cards with same style. I hope you get my point here. Also for your formatdate function, you may also need it fir request and offer cards.

aditilamkhede:
> > @aditilamkhede Sure I can expose everything in a separate file, but I'm worried that making everything public (reusable) by default will make things more confusing. quoted this article: https://itnext.io/thinking-in-styled-components-e230ea37c52c
> > "In short, don’t default to making things reusable just for the heck of it, because there are tradeoffs. It’s not necessarily better to make things as reusable as possible."
> 
> right @linadinh2602 but in our case our main aim was to create reusable components, like your post card, we have to create Request and Offer cards too. those styles which you created for post card, we are going to create many cards with same style. I hope you get my point here. Also for your formatdate function, you may also need it fir request and offer cards.

Also I may have wrote by mistake all styles from postcard.js but the ones which will be reusable just expose those.

aditilamkhede (approved):

PR #2223, opened 2021-02-17
Feature/desktop view tabs

clubkemp:
First, I know we talked about making the desktop view look like the menu from the help board, but I wanted to get the functionality working and mocked up to the actual designs of Figma before we decided to go that route. I do like this way which matches the Figma designs because it provides consistency for this page across views. Thoughts??

- Adding a utils function to check the window size so that we can change the position of the tabs from top to left
- Using the position property to modify the css of the ProfileTabs styled component so it doesn't break changing to position left/right

Also, I didn't modify the UploadPic component, it's all singe spaces and I can't figure out how to unstag them. I've tried to checkout the specific file from feature/profile-design and staging and I can't get the thing off this PR. Suggestions on how are welcome.
FightPandemics (1)


aditilamkhede:
> First, I know we talked about making the desktop view look like the menu from the help board, but I wanted to get the functionality working and mocked up to the actual designs of Figma before we decided to go that route. I do like this way which matches the Figma designs because it provides consistency for this page across views. Thoughts??
> 
> * Adding a utils function to check the window size so that we can change the position of the tabs from top to left
> * Using the position property to modify the css of the ProfileTabs styled component so it doesn't break changing to position left/right
> 
> Also, I didn't modify the UploadPic component, it's all singe spaces and I can't figure out how to unstag them. I've tried to checkout the specific file from feature/profile-design and staging and I can't get the thing off this PR. Suggestions on how are welcome.
> FightPandemics (1)

Hi @clubkemp yes we should match Figma. Cami confirmed this, it should match figma not Help Board. So great that you followed figma. I will check the PR today and provide my input.

client/src/pages/Profile.js
```diff
@@ -348,9 +350,10 @@ const Profile = ({
     return <ErrorAlert message={error} type="error" />;
   }
   if (loading) return <Loader />;
+
   const tabViews = {
     defaultView: "1",
-    position: "top",
+    position: window.width < 800 ? "top" : "left",
```
aditilamkhede: @clubkemp instead of hardcoding the values here and creating a utility function. use this 
  window.screen.width <= parseInt(mq.phone.wide.maxWidth);
 first thing is no hardcoding, second for all types of screens theres is screen size defined in themes.js, that way if changes are required it will be at just one place.
Also, to note that, any screensize 767 or less is mobile screen, 800 might fail some scenarios.

aditilamkhede (changes requested):
@clubkemp  Also in profile.js, line 352 or similar where you have defined const tabViews = {
in tabs property, tabs should be able to adjust based on conditional statement. For example, for own view we will see requests and posts menu but for others view we should see only Posts tab. When I tried it it shows a clear gap i.e. if menu is hidden there is a big gap between 2 menu tabs.
This is showing all tabs
Screen Shot 2021-02-17 at 2 59 43 PM
Hiding tab
Screen Shot 2021-02-17 at 2 59 22 PM

client/src/pages/Profile.js
clubkemp: Makes sense. So where does mq come from? Is there already a dev dependency that we can pull that from or are you suggesting we npm install something like react-responsive?

aditilamkhede: It's in theme.js --> import { theme, mq } from "constants/theme";

aditilamkhede:
> @clubkemp Also in profile.js, line 352 or similar where you have defined const tabViews = {
> in tabs property, tabs should be able to adjust based on conditional statement. For example, for own view we will see requests and posts menu but for others view we should see only Posts tab. When I tried it it shows a clear gap i.e. if menu is hidden there is a big gap between 2 menu tabs.
> This is showing all tabs
> Screen Shot 2021-02-17 at 2 59 43 PM
> Hiding tab
> Screen Shot 2021-02-17 at 2 59 22 PM

@clubkemp I hope you also noticed this as well.

client/src/pages/Profile.js
aditilamkhede: @clubkemp also try to use rem wherever possible instead of px.

aditilamkhede (approved):

PR #2229, opened 2021-02-22
Feature/2171 sort options dropdown

clubkemp:
Creating a dropdown for the sort feature for once we have our backend sorted out. The dropdown takes in a callback handleSortDropdown which we can use to fire whatever sorting features we come up with

I had to slightly modify the BaseSelector component to no longer have a parent container function in there. I backtracked it to the only other dropdown I could find and made sure it was still working. Posted about it in the general engineering channel but didn't hear anything

FightPandemics (2)

### DIFF
--- a/client/src/components/Selector/Selector.js
+++ b/client/src/components/Selector/Selector.js
@@ -13,24 +13,32 @@ const BaseSelector = ({
   minWidth,
   minHeight,
   popUpContainer,
-}) => (
-  <StyledSelector
-    suffixIcon={suffixIcon}
-    defaultValue={defaultValue}
-    filterOptions={filterOptions}
-    onChange={onChange}
-    getPopupContainer={
-      popUpContainer ? () => popUpContainer : () => document.body
-    }
-    minWidth={minWidth}
-    minHeight={minHeight}
-  >
-    {options.map((item) => (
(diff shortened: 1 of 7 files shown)


ddveloper:
Thanks @clubkemp , is there a mobile view dropdown for the same component? I don't see that in the mobile view

clubkemp (commented):
@ddveloper, the notion board didn't really include mobile view information so I just left it out, but you are right, should just be included.
 
https://www.notion.so/fightpandemics/Filters-Sort-By-Feed-e14712b0687749628c09947363e74e88

I went ahead and am using some ternary logic to display the dropdown component in two different places based on mobile view or not
FightPandemics (5)

client/src/components/Selector/Selector.js
```diff
@@ -10,23 +10,35 @@ const BaseSelector = ({
...
-        {item[key]}
-      </Option>
-    ))}
-  </StyledSelector>
-);
+  minWidth,
+  minHeight,
+  popUpContainer,
+}) => {
+  // const checkPopUp = () =>{
```
linadinh2602: is this comment out?

linadinh2602 (changes requested):
There is a comment out I think you forgot to delete. Feel free to merge after that

PR #2240, opened 2021-02-25
Feature/profile redesign menues

clubkemp:
### New option for navigating the section on the profile page using sectionView state instead of tabs

Created wrappers for the mobile (top nav) and desktop (side nav) version. Both are being displayed using a ternary operator that checks if the screen is mobile or not `window.width <= parseInt(mq.phone.wide.maxWidth)`

the menu onClick function takes in the value of the click and updates the sectionView state based off what you clicked on, this state will be used to hide/show your components

The mobile nav can probably stay where it is, but we will want to position the desktop wrapper once we are building the desktop view of things, ie we probably need a sidebar for it

FightPandemics (4)

### DIFF
--- a/client/src/components/Profile/ProfileComponents.js
+++ b/client/src/components/Profile/ProfileComponents.js
@@ -265,4 +265,40 @@ export const DescriptionDesktop = styled.div`
   color: ${colors.darkerGray};
 `;
 
-export const MobileMenuWrapper = styled(Menu)``;
+export const MobileMenuWrapper = styled(Menu)`
+  &.ant-menu {
+    width: 100%;
+    background-color: transparent;
+    border-bottom: 1px solid #e0e0e0;
+    height: 4rem;
+    display: flex;
+    justify-content: space-between;
+    overflow: auto;
+    &::-webkit-scrollbar {
+      display: none;
+    }
+    -ms-overflow-style: none;
+    scrollbar-width: none;
+  }
+  li.ant-menu-item {
+    margin: 0.8rem 0;
+    height: 3rem;
+    padding-bottom: 0rem;
+    overflow: visible;
+
+    color: ${theme.colors.darkerGray};
+    font-size: ${theme.typography.size.large};
+    line-height: 21px;
+    &:hover {
+      color: ${theme.colors.darkerGray};
+    }
+    text-align: center;
+  }
+
+  &.ant-menu .ant-menu-item-selected {
+    background-color: transparent;
+    border-bottom: 0.2rem solid ${theme.colors.black};
+    font-weight: bold;
+    overflow: visible;
+  }
+`;
--- a/client/src/components/Profile/ProfileComponents.js
+++ b/client/src/components/Profile/ProfileComponents.js
@@ -266,11 +266,14 @@ export const DescriptionDesktop = styled.div`
 `;
 
 export const MobileMenuWrapper = styled(Menu)`
+  &.ant-menu-vertical {
+    border-right: none;
+  }
   &.ant-menu {
     width: 100%;
     background-color: transparent;
     border-bottom: 1px solid #e0e0e0;
-    height: 4rem;
+    height: 4.5rem;
     display: flex;
     justify-content: space-between;
     overflow: auto;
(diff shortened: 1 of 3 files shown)


client/src/components/Profile/ProfileComponents.js
```diff
@@ -8,6 +9,8 @@ import { mq, theme } from "../../constants/theme";
 
 const { colors } = theme;
 
+const { SubMenu } = Menu;
```
aditilamkhede: are we using this? If not remove it.

aditilamkhede (approved):
@clubkemp  LGTM

client/src/components/Profile/ProfileComponents.js
```diff
@@ -261,3 +264,72 @@ export const DescriptionDesktop = styled.div`
...
+  }
+  li.ant-menu-item {
+    margin: 0.8rem 0;
+    height: 3rem;
+    padding-bottom: 0rem;
+    overflow: visible;
+
+    color: ${theme.colors.darkerGray};
+    font-size: ${theme.typography.size.large};
+    line-height: 21px;
```
linadinh2602: should you change all px and pt to rem?

client/src/pages/Profile.js
```diff
@@ -544,7 +485,98 @@ const Profile = ({
...
+            selectedKeys={sectionView}
+            onClick={handleMenuToggle}
+          >
+            {navMenu.map((item, index) => (
+              <Menu.Item key={item.name} disabled={item.disabled}>
+                {item.name}
+              </Menu.Item>
+            ))}
+          </MobileMenuWrapper>
+        ) : null}
```
linadinh2602, line 499: I'm a bit confused here. Can you instead of :null render desktop view?

client/src/pages/Profile.js
```diff
@@ -352,117 +399,11 @@ const Profile = ({
...
-              )}
-            </FeedWrapper>
-          </div>
-        ),
-      });
-    }
-    return TabViews;
+  const handleMenuToggle = (e) => {
+    setSectionView(e.key);
+    console.log("click ", e);
```
linadinh2602: remove console.log

client/src/pages/Profile.js
```diff
@@ -168,6 +174,47 @@ const Profile = ({
   const getActorQuery = () => {
     return organisationId ? `&actorId=${organisationId}` : "";
   };
+  const buildNavMenu = () => {
```
linadinh2602, line 177: With the navigation functionality built in Profile may hard to reuse in the future because every time we want to have a component that supports similar navigation then we will have to rewrite the logic for navigating the correct view and position the view in the right location. I think we should create a new component that's based on Menu, but add this navigation functionality so that it's more reusable (just a suggestion, that's fine if you disagree).

linadinh2602 (changes requested):
I left some comments just for some small changes.

client/src/pages/Profile.js
clubkemp: menu already is a component all by itself, I didn't make a new menu component All I've done is add a wrapper for menu that styles it according to our profile design. This is how the menu was designed on the feed page as well, so the handleMenuToggle IS actually passed down to Menu as an onClick prop and executed there, then we pass the event back to profile.

Does that clear up your concerns? Or should we talk it over some more?

client/src/pages/Profile.js
clubkemp: We can in this instance because they are going in the same place on this page. It might not be structured that way once we arrange the content for desktop view. I guess I should ask you and @aditilamkhede  if you would like me to arrange the page so there is a sidebar and you guys can just drop your content in the feed section on the right?

I can roll that up in this PR or we could do a seperate feature/desktop design PR. What makes sense to the team?

client/src/components/Profile/ProfileComponents.js
clubkemp: removed

client/src/pages/Profile.js
aditilamkhede: @clubkemp what I would like at least here is that, if we could move this function to utils and feed it an array or object of input. So it becomes reusable. Because we will be needing that at various places.

clubkemp: Makes sense. Ok I'll get a new branch going to make this more modular.

SinghRomika:
@clubkemp The build link for testing is giving a 404 error. Can you please check and give me the link to review. Thanks!
 https://featureprofile-redesign-menues.fightpandemics.xyz.

aditilamkhede:
> @clubkemp The build link for testing is giving a 404 error. Can you please check and give me the link to review. Thanks!
> https://featureprofile-redesign-menues.fightpandemics.xyz.

@SinghRomika, I was merging that with profile branch and it deletes the branch after merging. thus the error.

SinghRomika:
@aditilamkhede  thanks for sharing the link will review it!

PR #2249, opened 2021-02-28
#2248 Sorting posts by likes, updatedAt, views and shares

### LINKED ISSUE #2248: Sorting posts by likes, updatedAt, views and shares
In endpoints/posts.js, get the list of posts based on the order of ascending("'asc") or descending ("dsc") which:
- `GET /posts?sort=latest&order=asc` return posts sorted by latest posts by its order
- `GET /posts?sort=likes...` return posts sorted by most likes by its order
- `GET /posts?sort=views...` return posts sorted by view by its order
- `GET /posts?sort=shares...` return posts sorted by share by its order

linadinh2602:
In endpoints/posts.js, get the list of posts based on the order of ascending("'asc") or descending ("dsc") which:
- GET /posts?sort=latest&order=asc return posts sorted by latest posts by its order
- GET /posts?sort=likes... return posts sorted by most likes by its order
- GET /posts?sort=views... return posts sorted by view by its order
- GET /posts?sort=shares... return posts sorted by share by its order
_Please be concise and link any related issue(s): #2248

### DIFF
--- a/backend/lib/endpoints/posts.js
+++ b/backend/lib/endpoints/posts.js
@@ -48,6 +48,8 @@ async function routes(app) {
           authorId,
           ignoreUserLocation,
           filter,
+          sort,
+          order,
           keywords,
           limit,
           objective,
@@ -175,7 +177,11 @@ async function routes(app) {
             { $match: { $and: filters, $text: { $search: keywords } } },
             { $sort: { score: { $meta: "textScore" } } },
           ]
-          : [{ $match: { $and: filters } }, { $sort: { _id: -1 } }];
+          : sort
+            ? [
+              { $sort: { [sort]: order === "asc" ? 1 : -1 } }
+            ]
+            : [{ $match: { $and: filters } }, { $sort: { _id: -1 } }];
       /* eslint-enable sort-keys */
 
       /* eslint-disable sort-keys */
@@ -250,6 +256,8 @@ async function routes(app) {
             visibility: true,
             createdAt: true,
             updatedAt: true,
+            views: true,
+            shares: true,
           },
         },
       ];
--- a/backend/lib/endpoints/posts.js
+++ b/backend/lib/endpoints/posts.js
@@ -176,13 +176,11 @@ async function routes(app) {
           ? [
             { $match: { $and: filters, $text: { $search: keywords } } },
             { $sort: { score: { $meta: "textScore" } } },
-          ]
-          : sort
-            ? [
-              { $sort: { [sort]: order === "asc" ? 1 : -1 } }
-            ]
-            : [{ $match: { $and: filters } }, { $sort: { _id: -1 } }];
+          ] : [{ $match: { $and: filters } }, { $sort: { _id: -1 } }];
       /* eslint-enable sort-keys */
+      if (sort) {
+        sortAndFilterSteps.push({ $sort: { [sort]: order === "asc" ? 1 : -1 } })
+      }
 
       /* eslint-disable sort-keys */
       const paginationSteps =
--- a/backend/lib/endpoints/schema/posts.js
+++ b/backend/lib/endpoints/schema/posts.js
@@ -4,6 +4,8 @@ const { strictSchema, strictQueryStringSchema } = require("./utils");
 const {
   EXPIRATION_OPTIONS,
   POST_OBJECTIVES,
+  SORT_OPTIONS,
+  ORDER_OPTIONS,
   POST_TYPES,
   VISIBILITY_OPTIONS,
 } = require("../../models/Post");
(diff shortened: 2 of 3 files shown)


backend/lib/endpoints/schema/posts.js
```diff
@@ -13,6 +13,8 @@ const getPostsSchema = {
     .prop("actorId", S.string())
     .prop("authorId", S.string())
     .prop("filter", S.string()) // URI encoded JSON; TODO: figure out way to custom validation
+    .prop("sort", S.string()) // Sorting posts by likes, updatedAt, views and shares
```
ddveloper: it's better to add strict validation for the inputs, since it goes directly to the DB querying, could refer to the [Enum options like this one](https://github.com/FightPandemics/FightPandemics/blob/50edf4a5b84f16dbbffb864e85772715275023b3/backend/lib/endpoints/schema/posts.js#L9)

backend/lib/endpoints/posts.js
```diff
@@ -175,7 +177,11 @@ async function routes(app) {
             { $match: { $and: filters, $text: { $search: keywords } } },
             { $sort: { score: { $meta: "textScore" } } },
           ]
-          : [{ $match: { $and: filters } }, { $sort: { _id: -1 } }];
+          : sort
+            ? [
+              { $sort: { [sort]: order === "asc" ? 1 : -1 } }
```
ddveloper: it seems when the `sort` is passed in, the `$match: { $and: filters }` is removed, not sure if that's from the requirement or still need that ❔

clubkemp (commented):
This all looks good to me. I noticed you are commiting into feature/add-view-and-share-api. Is that the branch were this should go? Or should we pull into feature/2171-sort-feeds (where the dropdown got pulled into). Then we can integrate and test, PR any additional changes and merge into staging.

What do you think?

backend/lib/endpoints/posts.js
```diff
@@ -176,13 +176,11 @@ async function routes(app) {
...
             { $sort: { score: { $meta: "textScore" } } },
-          ]
-          : sort
-            ? [
-              { $sort: { [sort]: order === "asc" ? 1 : -1 } }
-            ]
-            : [{ $match: { $and: filters } }, { $sort: { _id: -1 } }];
+          ] : [{ $match: { $and: filters } }, { $sort: { _id: -1 } }];
       /* eslint-enable sort-keys */
+      if (sort) {
```
ddveloper, line 181: Hi @linadinh2602 , I just realized this is complex than I thought
> when the location is passed in, do we sort based on distance or the sort argument (likes, views, etc.)?
> when the keyword is passed in, do we sort based on keyword relevant or sort argument (likes, views, etc.)?

The current change appends a new `sort` clause after the sortAndFilterSteps, which could result in 2 `sort` steps for mongoDB to process I guess, like showing below, in which case the 1st sort will be overwritten by the 2nd one I assume, we may need to understand the requirement and combine them

image

clubkemp: Would it make more sense in each sort case to check if sort exists and if it does sort by sort param, otherwise sort by id?

`[{ $match: { $and: filters } }, sort ?{ $sort: { [sort]: order === "asc" ? 1 : -1 } } : { $sort: { _id: -1 } }];`

ddveloper: we have following sort mechanisms right now in the endpoint
```
by distance if location is present
by relevance if keyword is present
by 'likes/views/shares/..." from the drop-down menu
```
I think the PRD needs to define these behavior and based on the definition we can implement

ddveloper: e.g. if we combine location with the drop-down like below, since no `$maxDistance` is indicated I assume the `$geoNear` will return all records, if `sort === 'likes'` is present at the same time, will the result be any difference from the result sorted by 'likes' itself?
```      const sortAndFilterSteps = location
        ? [
            {
              $geoNear: {
                distanceField: "distance",
                key: "author.location.coordinates",
                near: {
                  $geometry: {
                    coordinates: location.coordinates,
                    type: "Point",
                  },
                },
                query: { $and: filters },
              },
            },
            sort
              ? { $sort: { [sort]: order === "asc" ? 1 : -1 } }
              : { $sort: { distance: 1, _id: -1 } },
          ]
```

backend/lib/models/Post.js
```diff
@@ -4,6 +4,8 @@ const { Schema, model, ObjectId } = require("mongoose");
 const EXPIRATION_OPTIONS = ["day", "week", "month", "forever"];
 const VISIBILITY_OPTIONS = ["city", "country", "state", "worldwide"];
 const POST_OBJECTIVES = ["request", "offer"];
+const SORT_OPTIONS = ["likes", "updatedAt", "views", "shares"];
```
ddveloper, line 7: just realized these 2 options are for the searching feature only, it seems they're not attributes of a `Post`, so maybe it's better to define them in the `lib/endpoints/schema/posts.js` instead ?

clubkemp: I will resolve this in the next branch where we fix the sorting exclusivity issue on the frontend

backend/lib/endpoints/posts.js
clubkemp: We will be resolving this on the front-end by making sorting mutually exclusive

clubkemp (approved):

PR #2243, opened 2021-02-25
Feature/profile redesign

aditilamkhede:
### What does this pull request aim to achieve? 💯
1. Navigation Menu for desktop and mobile

_Please be concise and link any related issue(s):_ 

- [ ] I have checked that no one else is working on similar changes.
- [ ] I have read the latest [**README**](README.md) and understand the development workflow.
- [ ] The name of this branch is: **`feature/<branch_name>`** for a regular feature, **`hotfix/<branch_name>`** for a hotfix, and **`service/<branch_name>`** for a microservice, where `<branch_name>` briefly describes the issue(s) resolved and is prefixed with the issue number(s) (e.g. `1127-1130-update-git-branching-model`). The branch is created in a cloned version of this repo, **not a forked repo**.
- [ ] This branch is rebased/merged with the **latest staging** (or the **latest production** if it is a hotfix).
- [ ] There are no merge conflicts.
- [ ] There are no console warnings and errors.
- [ ] On the right hand side of this PR, I have linked the appropriate issue(s) under "Issues" and added a project progress under "Projects".
- [ ] The title describes the issue(s) being addressed using format: **`<issue_numbers> - <brief_issue_description>`** (e.g. `#1127/#1130 - Update Git Branching Model`)

PR #2263, opened 2021-03-08
#2262 - Integrate PostTabCard mobile view in Profile Page

### LINKED ISSUE #2262: Integrate PostTabCard mobile view in Profile Page
Integrate <PostTabCard> mobile view in Profile Page

linadinh2602:
- Integrated tab card with Profile page (mobile view)
- Refactor integration mode from fetching data each time a user clicks on a tab to only fetching when data is not loaded
- Updating reducer and redux store to 2 views per [userId] "Requests", and "Offers" for the Profile page only
- "Requests" and "Offers" each have 3 modes "active", "inactive", and "all"
- If a user is self, they will be able to see "Requests" and "Offers" with "active" and "inactive" mode
- If a user isn't self, they will see "Posts" view with "Requests" and "Offers" all" mode
_Please be concise and link any related issue(s): #2262

### DIFF
--- a/backend/lib/endpoints/posts.js
+++ b/backend/lib/endpoints/posts.js
@@ -68,20 +68,13 @@ async function routes(app) {
       ];
       // "IA" stands for InActive or Archived
       //In future we will also have "D" for Drafts
-      if (postMode === 'IA') {
+      if (postMode === "IA") {
         filters = [
           {
-            $and: [{ expireAt: { $ne: null } }, { expireAt: { $lt: new Date() } }],
-            status: { $ne: "removed" },
-          },
-        ];
-      }
-
-      // "A" stands for Active or Active
-      if (postMode === 'A') {
-        filters = [
-          {
-            $or: [{ expireAt: { $eq: null } }, { expireAt: { $gt: new Date() } }],
+            $and: [
+              { expireAt: { $ne: null } },
+              { expireAt: { $lt: new Date() } },
+            ],
             status: { $ne: "removed" },
           },
         ];
@@ -136,14 +129,29 @@ async function routes(app) {
 
       //workMode filter
       if (workMode) {
-        const workModes = workMode.map(mode => mode.toLowerCase())
-        if (workModes.includes('both')) {
-          filters.push({ $or: [{ $and: [{ workMode: null }, { types: { $nin: ['Remote Work'] } }] }, { workMode: { $in: workModes } }] });
-        }
-        else if (workModes.includes('remote')) {
-          filters.push({ $or: [{ $and: [{ workMode: null }, { types: { $in: ['Remote Work'] } }] }, { workMode: { $in: workModes } }] });
-        }
-        else {
+        const workModes = workMode.map((mode) => mode.toLowerCase());
+        if (workModes.includes("both")) {
+          filters.push({
+            $or: [
+              {
+                $and: [
+                  { workMode: null },
+                  { types: { $nin: ["Remote Work"] } },
+                ],
+              },
+              { workMode: { $in: workModes } },
+            ],
+          });
+        } else if (workModes.includes("remote")) {
+          filters.push({
+            $or: [
+              {
+                $and: [{ workMode: null }, { types: { $in: ["Remote Work"] } }],
+              },
+              { workMode: { $in: workModes } },
+            ],
+          });
+        } else {
           filters.push({ workMode: { $in: workModes } });
         }
       }
@@ -190,45 +198,45 @@ async function routes(app) {
       /* eslint-disable sort-keys */
       const sortAndFilterSteps = location
         ? [
-          {
-            $geoNear: {
-              distanceField: "distance",
-              key: "author.location.coordinates",
-              near: {
-                $geometry: {
-                  coordinates: location.coordinates,
-                  type: "Point",
+            {
+              $geoNear: {
+                distanceField: "distance",
+                key: "author.location.coordinates",
+                near: {
+                  $geometry: {
+                    coordinates: location.coordinates,
+                    type: "Point",
+                  },
                 },
+                query: { $and: filters },
               },
-              query: { $and: filters },
             },
-          },
-          { $sort: { distance: 1, _id: -1 } },
-        ]
+            { $sort: { distance: 1, _id: -1 } },
+          ]
         : keywords
-          ? [
+        ? [
             { $match: { $and: filters, $text: { $search: keywords } } },
             { $sort: { score: { $meta: "textScore" } } },
           ]
-          : [{ $match: { $and: filters } }, { $sort: { _id: -1 } }];
+        : [{ $match: { $and: filters } }, { $sort: { _id: -1 } }];
       /* eslint-enable sort-keys */
 
       /* eslint-disable sort-keys */
       const paginationSteps =
         limit === -1
           ? [
-            {
-              $skip: skip || 0,
-            },
-          ]
+              {
+                $skip: skip || 0,
+              },
+            ]
           : [
-            {
-              $skip: skip || 0,
-            },
-            {
-              $limit: limit || POST_PAGE_SIZE,
-            },
-          ];
+              {
+                $skip: skip || 0,
+              },
+              {
+                $limit: limit || POST_PAGE_SIZE,
+              },
+            ];
       /* eslint-enable sort-keys */
 
       /* eslint-disable sort-keys */
@@ -303,13 +311,13 @@ async function routes(app) {
       const totalResultsAggregationPipeline = await Post.aggregate(
         keywords && !location
           ? [
-            { $match: { $and: filters, $text: { $search: keywords } } },
-            { $group: { _id: null, count: { $sum: 1 } } },
-          ]
+              { $match: { $and: filters, $text: { $search: keywords } } },
+              { $group: { _id: null, count: { $sum: 1 } } },
+            ]
           : [
-            { $match: { $and: filters } },
-            { $group: { _id: null, count: { $sum: 1 } } },
-          ],
+              { $match: { $and: filters } },
+              { $group: { _id: null, count: { $sum: 1 } } },
+            ]
       );
       /* eslint-enable sort-keys */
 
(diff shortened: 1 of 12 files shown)


aditilamkhede:
@Lina Dinh I just had a look at PR #2262, I haven't looked at the code yet. But these are my initial observation. 
It applies to both desktop and mobile: If I create a new post, it is not visible until I refresh the page. 
After you change from mobile to desktop, only 1 post is visible in desktop view.
Pagination also not working for desktop view after you change from mobile to desktop. As when we scroll down it should load more data as we scroll (if it has).
I was getting error 400. I don't have steps to reproduce. I will try more.
Note: I have tried all of these in #2243 None of these exist. I believe you implemented caching to avoid unnecessary backend calls.

backend/lib/endpoints/posts.js
```diff
@@ -68,10 +68,20 @@ async function routes(app) {
...
+        filters = [
+          {
+            $and: [{ expireAt: { $ne: null } }, { expireAt: { $lt: new Date() } }],
+            status: { $ne: "removed" },
+          },
+        ];
+      }
+
+      // "A" stands for Active or Active
+      if (postMode === 'A') {
```
aditilamkhede: We don't need this here. If postMode is inactive then only it will add filters for inactive/archived mode. The default data which comes is only for active posts only.

clubkemp (approved):
I noticed the same things Aditi did, but no new bugs from me. Nice work, there is a lot pushed here. I do have a general question. It looks like a lot of the fetching got moved to the profile view. Is this getting moved in from another place. I think before it was in Aditi's component, so is it no longer in there and getting passed the fetched posts as props? I just want to make sure we are being as DRY as possible here. Maybe the post fetch should actually live in a controller somewhere and be usesed both for feed and profile even. I dunno, maybe thats a future feature?

hiran11:
@linadinh2602  New post created is not getting displayed in "Requests" and "Offers" with "active" mode.

client/src/reducers/posts/reducer.js
```diff
@@ -24,6 +26,114 @@ const postsReducer = (state = innitialState, action) => {
         isLoading: false,
       };
     }
+    case POSTS_ACTIONS.FETCH_REQUESTS_ACTIVE_SUCCESS: {
+      const { payload } = action;
```
ddveloper: I am not familiar with Redux, but seems here the logic can be simplified ❔

ddveloper (commented):
Thank you @linadinh2602 sorry I am not familiar with how Redux works, so if I get it wrong please correct me, I don't know how the backend query is carried out in different layout, just want to add some general comment, since there are different layout to represent the `posts` data, it seems let different components reuse single backend querying would make sense.

client/src/pages/Profile.js
```diff
@@ -376,21 +440,45 @@ const Profile = ({
...
   useEffect(() => {
     refetchPosts(); // will trigger loadPosts(if needed) (by toggling toggleRefetch)
-  }, [internalTab]); // eslint-disable-line react-hooks/exhaustive-deps
+  }, [internalTab, sectionView]); // eslint-disable-line react-hooks/exhaustive-deps
 
   const refetchPosts = (isLoading, loadMore) => {
     dispatch(postsActions.resetPageAction({ isLoading, loadMore }));
     if (page === 0) {
-      setToggleRefetch(!toggleRefetch);
+      const view = getView();
```
aditilamkhede, line 451: all the conditions have same line of code inside it, i.e. setToggleRefetch(!toggleRefetch);. 
I would suggest, 
1. Either remove the conditional logic, if all use cases can be full-filled.
2. Or combine it into one condition

client/src/reducers/posts/reducer.js
linadinh2602: Originally I think it makes sense for feed & profile page to be reusing the same redux action. However, the profile changes into loading a sub-group of posts (active / archived / request / response / all) so reusing the existing state in the store makes it actually much more complicated if I want to cached the data. Thus, I remap the posts to this structure to allow caching to happen. Otherwise, the user will need to wait for a fetch everytime they interact with the component which I think is inefficient. Each query takes about a second, so it's not insignificant.

linadinh2602: Hmm though maybe there's a way to group all the logic together like @aditilamkhede suggests too. Let me think about it.

linadinh2602:
> I noticed the same things Aditi did, but no new bugs from me. Nice work, there is a lot pushed here. I do have a general question. It looks like a lot of the fetching got moved to the profile view. Is this getting moved in from another place. I think before it was in Aditi's component, so is it no longer in there and getting passed the fetched posts as props? I just want to make sure we are being as DRY as possible here. Maybe the post fetch should actually live in a controller somewhere and be usesed both for feed and profile even. I dunno, maybe thats a future feature?

I believe the profile page should be the parent component that handle fetching & logic, where our desktop & mobile view should be the consumer of that logic (presentational component). Feed is a different parent page and should have different logic. I think trying to reuse logic in Feed may not always be the right thing to do since Feed's purpose and Profile's purpose is different now. If I still use the same logic as Feed, I'll no longer be able to cache the data for example. In feed it doesn't matter since user doesn't have a way to click back and forth a lot, but with Tab view I think it matters a lot more since people may click back and forth.

client/src/pages/Profile.js
```diff
@@ -621,6 +688,10 @@ const Profile = ({
   const emptyFeed = () =>
     filteredPost.length < 1 && !isLoading;
 
+  if (!actorId || !userId) {
+    return <></>;
+  }
+
   return (
     <>
       <ProfileBackgroup />
```
aditilamkhede: Can you also not combine these together? They look same just condition is different instead can you not use sectionview = ''Offer" || sectionview = ''Posts" || sectionview = ''Request"

aditilamkhede (changes requested):

client/src/pages/Profile.js
linadinh2602: Sorry, I'm not sure I understand your request. Can you elaborate? I don't see how checking for "actorId" and "userId" relates to sectionView

backend/lib/endpoints/posts.js
linadinh2602: Will fix, thanks!

aditilamkhede:
@linadinh2602 here is my observation: 
1. I am **not signed in** -> When I come to help board and click on a users name, It should display that users profile. But it is displaying a blank page. If I log in, it works as expected.
2. If I delete a post in desktop view it deletes a post from backend but is not updating UI. In mobile view it works as expected.

clubkemp:
Looking good Lina! Quick question, when I land on the mobile view I see the three latest compact cards. Is it part of the design that I should be taken to the post when I click on one of those cards? Or do I have to go into the see all? Which as I am typing this, I'm realizing that I am not able to click on see all either, so maybe your click handlers got lost somewhere?

aditilamkhede:
> Looking good Lina! Quick question, when I land on the mobile view I see the three latest compact cards. Is it part of the design that I should be taken to the post when I click on one of those cards? Or do I have to go into the see all? Which as I am typing this, I'm realizing that I am not able to click on see all either, so maybe your click handlers got lost somewhere?

@clubkemp I don't think it was mentioned that when user clicks on the post in mobile view, it should open that post. For See All link, I still need to link it.

client/src/pages/Profile.js
```diff
@@ -587,26 +696,92 @@ const Profile = ({
...
+                    key={item.name}
+                    disabled={item.disabled}
+                    id={GTM.user.profilePrefix + GTM.profile[item.gtm]}
+                  >
+                    {item.name}
+                  </Menu.Item>
+                ))}
+              </MobileMenuWrapper>
+              <div style={{ width: "100%" }}>
+                {sectionView === "Requests" && (
```
aditilamkhede, line 737: @linadinh2602 as such LGTM, here I have a suggestion . Sectionview is offer or request or post you are showing Postcard, you can just dynamically pass internal tab type i.e. active/Archived or  requests/Offers. So instead of repeating it 3 times just write it once. Just a suggestion .

client/src/pages/Profile.js
```diff
@@ -585,9 +600,11 @@ const Profile = ({
 
   const emptyFeed = () => filteredPost.length < 1 && !isLoading;
 
-  if (!userId) {
+  if (authLoading == null || authLoading === true) {
+    console.log("NOT LOADING");
```
aditilamkhede: please remove console.log

aditilamkhede (commented):
LGTM.

aditilamkhede (approved):

client/src/pages/Profile.js
linadinh2602: I think that is a good suggestion. I'll merge it now, and work on this in a separate PR to not blocked the code getting in staging.

hiran11:
@linadinh2602 : I could see "see  all" doesnt display all post . here is the screen shot and steps i have attached

Environment : 
Mobile ,Android OS  9.0 , Chrome browser 

Precondition: user must be logged with credentials
1. Open the url 
2. Tap on "view my profile" 
3. Tap on "offers" tab
4. Click on Active posts

 This is the post display in mobile 
image

This is the post display in windows 

image

When i click see all to check 4 th post , it doesnt work . Correct me if am using wrong build url . 

https://featureprofile-redesign.fightpandemics.xyz/profile/602d4fd57351b21700f0f858

aditilamkhede:
> https://featureprofile-redesign.fightpandemics.xyz/profile/602d4fd57351b21700f0f858

We haven't linked see-all yet. Could you please elaborate what do you mean by see-all here.

hiran11:
@linadinh2602  i could view 4 post in the desktop  but in mobile i could view only 3 post . The 4th post is not displayed in the mobile view

aditilamkhede:
> @linadinh2602 i could view 4 post in the desktop but in mobile i could view only 3 post . The 4th post is not displayed in the mobile view

@hiran11 mobile view only display 3 posts. Desktop view will display all the posts for that tab.

client/src/pages/Profile.js
aditilamkhede: sounds good.

hiran11:
@aditilamkhede ok so the 4 th post will be displayed when "see all" is connected . Got it

PR #2264, opened 2021-03-08
Feature/profile redesign bugfixes

clubkemp:
Couple fixes
- Putting the help location back onto the posts, got removed out of components/feed/post.js
- hiding drafts as an option on the menu 
- fixing misses on GTM tags
- fixing the isSelf undefined bug

I compared our feature/profile-redesign branch to staging and the only thing that I could find that was off was the section where the remote/ in-person was missing from the post card, and a typo that was in the translation file. Everything else was either a new file, or something that got moved into a util file for example


aditilamkhede (approved):
@clubkemp Thanks for quick fixes. I have only one question before I merge, cypress tests came from merging staging right?

clubkemp:
> @clubkemp Thanks for quick fixes. I have only one question before I merge, cypress tests came from merging staging right?

yep!

JayashreeSamantasinghar (approved):
Thanks

PR #2292, opened 2021-03-23
Feature/2171sortFeed

clubkemp:
Adding a sort dropdown to the helpboard that will sort the feeds based on user input. At this time the functionality is mutually exclusive, meaning that a user can only be sorting by one thing at a time. A couple ways to do this is with the actual dropdown itself. Another way is if a user selects a location filter, toggles show near me, or uses the search query. Any activation of one sorting mechanism clears out the others and applies that sort

V2 of this sort functionality will need to use secondary sorts to achieve a more user friendly experience


client/src/pages/PostPage.js
```diff
@@ -279,7 +279,7 @@ const PostPage = ({ user, updateComments, isAuthenticated }) => {
         <Body>{t("post.expire")}</Body>
         <StyledLink
           id={GTM.organisation.completedPrefix + GTM.profile.continueToFeed}
-          to="/feed"
+          to={"/feed"}
```
aditilamkhede: it doesn't have to be in curly brackets

client/src/utils/windowSize.js
```diff
@@ -0,0 +1,26 @@
+import { useState, useEffect } from "react";
```
aditilamkhede, line 1: @clubkemp didn't we add this also in profile redesign?

aditilamkhede (dismissed):
@clubkemp Thanks!! LGTM, just minor comments

client/src/utils/windowSize.js
clubkemp: yes we did, I needed it for this feature as well. They should be named and are the same exact thing so shouldn't be a merge conflict right?

client/src/pages/PostPage.js
clubkemp: Removed them

